PR #11, opened 2018-03-22
Clean names multiindex

### LINKED ISSUE #10: clean_names breaks for MultiIndex Columns
MultiIndex columns returns `expected str not tuple` type error.

If you instead used,
`df.rename(columns=lambda x: x.lower().replace('_', ''))`
this would work for standard and MultiIndex DataFrames.

Can do PR if required.

JoshuaC3:
Change `cleaan_names` so that it works for MultiIndexes and also add tests for this.

Fixes #10

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -31,26 +31,22 @@ def clean_names(df):
     :param df: The pandas DataFrame object.
     :returns: A pandas DataFrame.
     """
-    columns = [(c.lower()
-                .replace(' ', '_')
-                .replace('/', '_')
-                .replace(':', '_')
-                .replace("'", '')
-                .replace('’', '')
-                .replace(',', '_')
-                .replace('?', '_')
-                .replace('-', '_')
-                .replace('(', '_')
-                .replace(')', '_')
-                .replace('.', '_')
-                ) for c in df.columns]
-
-    newcolumns = []
-    for col in columns:
-        # NOTE: Replace repeating underscores with single ones
-        newcol = re.sub('[_]{2,}', '_', col)
-        newcolumns.append(newcol)
-    df.columns = newcolumns
+    df = df.rename(
+    	columns=lambda x: x.lower()
+	                       .replace(' ', '_')
+                           .replace('/', '_')
+                           .replace(':', '_')
+                           .replace("'", '')
+                           .replace('’', '')
+                           .replace(',', '_')
+                           .replace('?', '_')
+                           .replace('-', '_')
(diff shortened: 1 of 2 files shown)


janitor/functions.py
```diff
@@ -31,26 +31,22 @@ def clean_names(df):
...
+                           .replace('’', '')
+                           .replace(',', '_')
+                           .replace('?', '_')
+                           .replace('-', '_')
+                           .replace('(', '_')
+                           .replace(')', '_')
+                           .replace('.', '_')
+    )
+
+    df = df.rename(columns=lambda x: re.sub('_+', '_', x))
```
ericmjl: Great stuff!

tests/test_functions.py
```diff
@@ -26,6 +26,17 @@ def null_df():
...
+def multiindex_dataframe():
+    data = {
+        ('a', 'b'): [1, 2, 3],
+        ('Bell__Chart', 'Normal  Distribution'): [1, 2, 3],
+        ('decorated-elephant', "r.i.p-rhino :'("): [1, 2, 3],
+    }
+    df = pd.DataFrame(data)
+    return df
+
+
```
ericmjl: Wonderful, thanks for following the programming pattern!

tests/test_functions.py
```diff
@@ -106,3 +117,45 @@ def test_convert_excel_date():
...
+    levels = [
+        ['a', 'bell_chart', 'decorated_elephant'],
+        ['b', 'normal_distribution', 'r_i_p_rhino_']
+    ]
+
+    labels = [[0, 1, 2], [0, 1, 2]]
+
+    expected_columns = pd.MultiIndex(levels=levels, labels=labels)
+    assert set(df.columns) == set(expected_columns)
```
ericmjl: Wow, you really went above-and-beyond here! Thanks for doing this!

ericmjl:
Thanks for your contribution, @JoshuaC3! I am going to merge.

tests/test_functions.py
JoshuaC3: Thanks - My only thought on this was, should we be handling `levels`, `labels` and maybe `expected_columns` at a global level so that I am not just repeating the setup of variables? I am not experienced with writing test so I am not sure what is good or standard practice.

JoshuaC3:
Thank you for the positive feedback. I learnt a lot contributing here.

There are some other parts I that I think could be improved. I will raise some issues for them.

PR #13, opened 2018-03-28
Clean names remove outer underscores

### LINKED ISSUE #12: Remove Trailing (and Leading) underscores from clean_names
Pandas handles this relatively well but would be a good as a kwarg for `jn.clean_names()`.

The default `False` or `None` could leave leading, `True` or `'both'` to remove all and trailing and then pass `'leading'` and `'trailing`' or similar to remove each.

JoshuaC3:
Adds kwarg to the function `clean_names` to strip leading and trailing underscores.

I did change some the syntax previously decided in the original issue. This was because I used pythons built-in `strip`, `lstrip` and `rstrip` string functions to remove the underscores. Therefore, it seemed more intuitive and "pythonic" to use `'left'` and `'right'` (and `'both'`).

I also added a shorthand `'l'` and `'r'` and `True`. This type of behaviour is common place in the Pandas API e.g. [here](https://github.com/pandas-dev/pandas/blob/v0.22.0/pandas/core/generic.py#L5319) and [here](https://github.com/pandas-dev/pandas/blob/a00154dcfe5057cb3fd86653172e74b6893e337d/pandas/io/parsers.py#L172), so this should be familiar to users.

We can also add `'end'` and `'start'` easily if these are still desired. Or, if you feel the above is not correct, I can just add exclusively `'both'`, `'end'` and `'start'` for a more terse/concise API.

Cheers

Closes #12

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -32,7 +32,7 @@ def _strip_underscores(df, strip_underscores=None):
 
     if strip_underscores in ['left', 'l']:
         df = df.rename(columns=lambda x: x.lstrip('_'))
-    elif strip_underscores in ['riaght', 'r']:
+    elif strip_underscores in ['right', 'r']:
         df = df.rename(columns=lambda x: x.rstrip('_'))
     elif strip_underscores == 'both' or strip_underscores is True:
         df = df.rename(columns=lambda x: x.strip('_'))
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -8,7 +8,35 @@ from .errors import JanitorError
 import re
 
 
-def clean_names(df):
+def _strip_underscores(df, strip_underscores=None):
+    """
+    Strip underscores from the beginning, end or both of the
+    of the DataFrames column names.
+
+    .. code-block:: python
+
+        df = _strip_underscores(df, strip_underscores='left')
+
+    :param df: The pandas DataFrame object.
+    :param strip_underscores: A str of either 'left', 'right' or 'both'.
+    :returns: A pandas DataFrame.
+    """
+    underscore_options = [None, 'left', 'right', 'both', 'l', 'r', True]
+    if strip_underscores not in underscore_options:
+        raise JanitorError(
+            """strip_underscores must be one of: %s""" % underscore_options
+            )
+
+    if strip_underscores in ['left', 'l']:
+        df = df.rename(columns=lambda x: x.lstrip('_'))
+    elif strip_underscores in ['right', 'r']:
+        df = df.rename(columns=lambda x: x.rstrip('_'))
+    elif strip_underscores == 'both' or strip_underscores is True:
+        df = df.rename(columns=lambda x: x.strip('_'))
+    return df
+
+
+def clean_names(df, strip_underscores=None):
     """
     Clean column names.
 
@@ -29,6 +57,7 @@ def clean_names(df):
         df = jn.DataFrame(df).clean_names()
 
     :param df: The pandas DataFrame object.
+    :param strip_underscores: A str of either 'left', 'right' or 'both'.
     :returns: A pandas DataFrame.
     """
     df = df.rename(
@@ -44,9 +73,11 @@ def clean_names(df):
                            .replace('(', '_')
                            .replace(')', '_')
                            .replace('.', '_')
-    )
+
+                           )
 
     df = df.rename(columns=lambda x: re.sub('_+', '_', x))
+    df = _strip_underscores(df, strip_underscores)
     return df
 
 
@@ -189,8 +220,8 @@ def get_features_targets(df, target_columns, feature_columns=None):
     else:
         if isinstance(target_columns, str):
             xcols = [c for c in df.columns if target_columns != c]
-        elif (isinstance(target_columns, list)
-                or isinstance(target_columns, tuple)):
+        elif (isinstance(target_columns, list) or
+              isinstance(target_columns, tuple)):
             xcols = [c for c in df.columns if c not in target_columns]
         X = df[xcols]
     return X, Y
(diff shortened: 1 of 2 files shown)


janitor/functions.py
```diff
@@ -8,7 +8,35 @@
...
+    """
+    Strip underscores from the beginning, end or both of the
+    of the DataFrames column names.
+
+    .. code-block:: python
+
+        df = _strip_underscores(df, strip_underscores='left')
+
+    :param df: The pandas DataFrame object.
+    :param strip_underscores: A str of either 'left', 'right' or 'both'.
```
ericmjl: Please document the options in line 24, and what they're intended to do.

janitor/functions.py
```diff
@@ -289,8 +320,8 @@ def convert_excel_date(df, column):
     :param str column: A column name.
     :returns: A pandas DataFrame with corrected dates.
     """
-    df[column] = (pd.TimedeltaIndex(df[column], unit='d')
-                  + dt.datetime(1899, 12, 30))
+    df[column] = (pd.TimedeltaIndex(df[column], unit='d') +
+                  dt.datetime(1899, 12, 30))
```
ericmjl: Logical and mathematical operators are put on a new line for clarity (particularly if there are multiple math operations going on). Please revert this.

janitor/functions.py
```diff
@@ -189,8 +220,8 @@ def get_features_targets(df, target_columns, feature_columns=None):
     else:
         if isinstance(target_columns, str):
             xcols = [c for c in df.columns if target_columns != c]
-        elif (isinstance(target_columns, list)
-                or isinstance(target_columns, tuple)):
+        elif (isinstance(target_columns, list) or
+              isinstance(target_columns, tuple)):
```
ericmjl: Logical and mathematical operators are put on a new line for clarity (particularly if there are multiple math operations going on). Please revert this.

janitor/functions.py
```diff
@@ -29,6 +57,7 @@ def clean_names(df):
         df = jn.DataFrame(df).clean_names()
 
     :param df: The pandas DataFrame object.
+    :param strip_underscores: A str of either 'left', 'right' or 'both'.
```
ericmjl: Likewise here, please document the options for clarity purposes, and what they're intended to do.

ericmjl:
@JoshuaC3 thanks for the PR! This looks great. Just for my own memory recordkeeping, here's my thoughts below:

> I did change some the syntax previously decided in the original issue. This was because I used pythons built-in strip, lstrip and rstrip string functions to remove the underscores. Therefore, it seemed more intuitive and "pythonic" to use 'left' and 'right' (and 'both').

Good reasoning. Let's follow the Pandas API then. 

> I also added a shorthand 'l' and 'r' and True. This type of behaviour is common place in the Pandas API e.g. here and here, so this should be familiar to users.

As mentioned in the code review, I'd like to have those options clearly documented, otherwise it'll be a nightmare for my future self to decipher!

> We can also add 'end' and 'start' easily if these are still desired. Or, if you feel the above is not correct, I can just add exclusively 'both', 'end' and 'start' for a more terse/concise API.

Not necessary, reasoning for shorthand and copying the "left/right" paradigm is good enough justification.

janitor/functions.py
JoshuaC3: Sure. Sorry, I only changed it as I was getting these two cautions from flake8.
```
janitor/functions.py:223:15: W503 line break before binary operator
janitor/functions.py:323:19: W503 line break before binary operator
```
Reverted to operator on newline.

janitor/functions.py
JoshuaC3: Added better documentation of `strip_underscores`.

JoshuaC3:
I think that is all of the changes now.

Do I create another PR or does this one update now that I have pushed the changes to this branch?

ericmjl:
@JoshuaC3: The latter is true! :smile:

ericmjl:
@JoshuaC3: one thing I couldn't easily decipher from the test - does the strip underscore test explicitly deal with the left and right underscores? Please let me know.

JoshuaC3:
It did not previously but I am pushing the left and right tests now. I have not done tests for `'l'`, `'r'` or `True` but can do if needed.

ericmjl:
> I have not done tests for 'l', 'r' or True but can do if you think it is needed.

@JoshuaC3 I think it's better to have more tests than fewer. Let's get those in as well while you have the momentum, otherwise my future self laziness will get in the way of making it happen!

JoshuaC3:
All done :) learning lots from this too so thanks.

tests/test_functions.py
```diff
@@ -159,3 +159,46 @@ def test_multiindex_clean_names_pipe(multiindex_dataframe):
...
+    ]
+
+    labels = [[1, 0, 2], [1, 0, 2]]
+
+    expected_columns = pd.MultiIndex(levels=levels, labels=labels)
+    assert set(df.columns) == set(expected_columns)
+
+
+def test_clean_names_strip_underscores_right(multiindex_dataframe):
+    df = clean_names(multiindex_dataframe, strip_underscores='right')
```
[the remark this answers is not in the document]
JoshuaC3: We only need that for 'l' and 'left'. The original multiindex_dataframe has a right/trailing underscore on `r_i_p_rhino_`. There were no leading/left underscores so I added some with `df = multiindex_dataframe.rename(columns=lambda x: '_' + x)` . Is this ok?

BTW, It was actually the trailing underscore in `r_i_p_rhino_` that made me think that this strip_underscores was needed!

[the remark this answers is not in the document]
ericmjl: > We only need that for 'l' and 'left'. The original multiindex_dataframe has a right/trailing underscore on r_i_p_rhino_. There were no leading/left underscores so I added some with df = multiindex_dataframe.rename(columns=lambda x: '_' + x) . Is this ok?

Ok got it! Thanks for pointing it out, I should have read the code a bit more closely.

ericmjl:
@JoshuaC3 everything looks great! Thanks for contributing :smile:.

JoshuaC3:
Great :smile: I got there eventually. Thanks for your guidance.

ericmjl:
It's my pleasure, @JoshuaC3! 

Btw, I got my start contributing to open source software through the guidance of the matplotlib team, who guided me basically the same way. If and when you get a chance, pay it forward to someone else too! :smile:

PR #22, opened 2018-07-18
Rework pandas flavour

ericmjl:
Enables new way of using pandas dataframes: just import janitor, and all of janitor's functions are automatically attached and registered onto a pandas DataFrame object! No more wrapping :smile:


ericmjl:
With special thanks to @Zsailer!

PR #31, opened 2018-10-24
Add limit_column_characters function with tests

szuckerman:
As per [issue 15](https://github.com/ericmjl/pyjanitor/issues/15) I've added a limit_column_characters function.

There's three tests; they should be sufficient.

I don't think we could have gone with the suggestion to use the df_column_uniquify function in [this stackoverflow answer](https://stackoverflow.com/a/44957247) since it will only work if the duplicate columns are next to each other. If they skip and repeat later, the count starts again.

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -36,7 +36,9 @@ def _strip_underscores(df, strip_underscores=None):
     """
     underscore_options = [None, "left", "right", "both", "l", "r", True]
     if strip_underscores not in underscore_options:
-        raise JanitorError(f"strip_underscores must be one of: {underscore_options}")
+        raise JanitorError(
+            f"strip_underscores must be one of: {underscore_options}"
+        )
 
     if strip_underscores in ["left", "l"]:
         df = df.rename(columns=lambda x: x.lstrip("_"))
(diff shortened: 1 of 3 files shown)


janitor/functions.py
```diff
@@ -788,3 +788,55 @@ def add_column(df, colname: str, value):
...
+      first distinct column will be 10 characters and the remaining will be 12
+      characters (assuming a column separator of one character)
+    :param col_separator: The separator to use for counting distinct column values.
+      I think an underscore looks nicest, however a period is a common option as well.
+      Supply an empty string (i.e. '') to remove the separator.
+
+    """
+
+    assert isinstance(column_length, int), "`column_length` must be an integer!"
+    assert isinstance(col_separator, str), "`col_separator` must be a string!"
```
ericmjl: Very nice! I'm always a fan of defensive programming checks :smile:

szuckerman:
Sure! I'll run it through `black` and update the PR. Might not get to it today though; `black` is on my list for new tools to try out 😃 .

szuckerman:
Looks like `black` has an 88 character limit; gotta change it to 79 for Travis.

ericmjl:
Right: `black -l 79 .` should do the trick.

szuckerman:
I'm really not sure why the Travis build keeps failing... Are the previous commits making the problem?

ericmjl:
@szuckerman this looks like an error on my side. Your PR looks good to me at this point. I just have to fix something underlying, I guess.

ericmjl:
As things turn out, it was an issue with conda. Let me fix this on the PR.

ericmjl:
Woohoo, everything looks good to me! Merging in. Thanks @szuckerman for your contribution!

PR #33, opened 2018-10-26
Fixed some logic and added examples

szuckerman:
I added some logic to `limit_column_characters` so that the loops won't run if all the column headers are unique.

I also realized there was a 'bug' in `add_column` that would allow someone to overwrite the column if the same name was used. I added an assertion to check for that.

I also added a `fill_values` parameter to `add_column` so users can add 'repeating' values if the tuple or list isn't as long as the DataFrame.

I also added examples for both of the methods.


examples/add_column.md
```diff
@@ -0,0 +1,92 @@
+# df.add_column()
```
ericmjl: @szuckerman I'm super happy that you've added examples! 

Can I ask, how open are you to formatting this not as a Markdown file, but instead as a ReStructured Text file under `docs/`? The RST files are built as part of the hosted documentation on ReadTheDocs, and it'll get more visibility. 

That said, if it's too much of a burden, I'm happy to delegate or handle the work for a future PR, and accept this one in the interim.

janitor/functions.py
```diff
@@ -784,14 +801,34 @@ def add_column(df, colname: str, value):
         for the column name to be compatible with the Feather binary
         format (this is a useful thing to have).
     :param value: Either a single value, or a list/tuple of values.
+    : fill_remaining: If value is a tuple or list that is smaller than
```
ericmjl: Extraneous space. 
```suggestion
    :fill_remaining: If value is a tuple or list that is smaller than
```

szuckerman: Actually it should probably read `:param fill_remaining:`.

examples/add_column.md
szuckerman: Sure; I'm fine with the RST formatting. I'll update accordingly. 

Are you ok with the structure of the examples? I'm thinking I should update to adding the parameter values in the header similar to the Pandas documentation (instead of having empty parentheses).

janitor/functions.py
szuckerman: I'll accept it for now to push the change, but will update in a later commit.

szuckerman:
I added another function `row_to_names` with documentation and tests.  I also converted the md files to rst and put them in the /docs folder.

Looking back at this PR, I'm a bit confused since there's still 'changes' being requested from 4 days ago, but I can't quite find anywhere to accept or make them.

ericmjl:
@szuckerman I think I should approve changes first, and then re-review. That might be causing the confusion.

ericmjl:
@szuckerman looks like the tests are failing on pycodestyle. Would you be kind enough to fix them please? Details [here](https://travis-ci.org/ericmjl/pyjanitor/builds/449946401?utm_source=github_status&utm_medium=notification).

ericmjl:
@szuckerman just wanted to catch this ahead-of-time, but if possible, let's limit the rest of this PR to fixing the changes here. PRs are generally better if kept short and atomic, which makes it easier to review.

szuckerman:
@ericmjl sorry about gumming-up this pull request with the `row_to_names` func... Just getting started with the github build system and didn't realize that it would get included with the original pull request when pushing to my repo.

I'm fixing all the changes so this will get pushed through.

ericmjl:
@szuckerman no worries! It's a learning journey, and I did the exact thing you did here. Thankfully it was just a second feature add, so no problem with that. :smile:.

ericmjl:
@szuckerman btw, it looks like everything is good to go, and it's good to merge. Let me make sure you're acknowledged in the contributors list too.

szuckerman:
Interesting.... I see myself on the contributor's list.

ericmjl:
Oh, wait, silly me. That's right, you are on there.

It's for this reason I am known to be algebra-blind!

Everything looks good. I am going to approve changes and merge.

PR #45, opened 2018-11-06
Add round_to_fraction with tests and example

szuckerman:
I added a `round_to_fraction` which rounds floats in a column to the nearest fraction.

I.e. you can round to the nearest half, third or 1/7th.

I saw it as a feature request for R's janitor: https://github.com/sfirke/janitor/issues/235

### DIFF
--- a/docs/round_to_fraction.rst
+++ b/docs/round_to_fraction.rst
@@ -1,0 +1,124 @@
+df.round_to_fraction()
+======================
+
+Description
+-----------
+
+This method modifies a column of floats to a nearest fraction, given a
+certain denominator.
+
+Parameters
+----------
+
+df
+~~
+
+A pandas dataframe.
+
+colname
+~~~~~~~
+
+The column which to apply the fraction transformation.
+
+denominator
+~~~~~~~~~~~
+
+The denominator of the fraction to round to. For example, ``2`` would
+values to the nearest half (i.e. ``1/2``).
+
+Setup
+-----
+
+.. code:: python
+
+    import pandas as pd
+    import janitor
+     
+    data_dict = {
+        "a": [1.23452345, 2.456234, 3.2346125] * 3,
+        "Bell__Chart": [1/3, 2/7, 3/2] * 3,
+        "decorated-elephant": [1/234, 2/13, 3/167] * 3,
+        "animals": ["rabbit", "leopard", "lion"] * 3,
+        "cities": ["Cambridge", "Shanghai", "Basel"] * 3,
+    }
+
+Example1: Rounding the first column to the nearest half
+-------------------------------------------------------
+
+.. code:: python
+
+    example_dataframe = pd.DataFrame(data_dict)
+     
+    example_dataframe.round_to_fraction('a', 2)
+
+Output
+~~~~~~
+
+::
+
+         a  Bell__Chart  decorated-elephant  animals     cities
+    0  1.0     0.333333            0.004274   rabbit  Cambridge
+    1  2.5     0.285714            0.153846  leopard   Shanghai
+    2  3.0     1.500000            0.017964     lion      Basel
+    3  1.0     0.333333            0.004274   rabbit  Cambridge
+    4  2.5     0.285714            0.153846  leopard   Shanghai
+    5  3.0     1.500000            0.017964     lion      Basel
+    6  1.0     0.333333            0.004274   rabbit  Cambridge
+    7  2.5     0.285714            0.153846  leopard   Shanghai
+    8  3.0     1.500000            0.017964     lion      Basel
+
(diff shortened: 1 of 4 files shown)


examples/round_to_fraction.md
```diff
@@ -0,0 +1,92 @@
...
+data_dict = {
+    "a": [1.23452345, 2.456234, 3.2346125] * 3,
+    "Bell__Chart": [1/3, 2/7, 3/2] * 3,
+    "decorated-elephant": [1/234, 2/13, 3/167] * 3,
+    "animals": ["rabbit", "leopard", "lion"] * 3,
+    "cities": ["Cambridge", "Shanghai", "Basel"] * 3,
+}
+```
+
+## Example1: Rounding the first column to the nearest half
```
ericmjl:
```suggestion
## Example 1: Rounding the first column to the nearest half
```

examples/round_to_fraction.md
```diff
@@ -0,0 +1,92 @@
...
+    1  2.5     0.285714            0.153846  leopard   Shanghai
+    2  3.0     1.500000            0.017964     lion      Basel
+    3  1.0     0.333333            0.004274   rabbit  Cambridge
+    4  2.5     0.285714            0.153846  leopard   Shanghai
+    5  3.0     1.500000            0.017964     lion      Basel
+    6  1.0     0.333333            0.004274   rabbit  Cambridge
+    7  2.5     0.285714            0.153846  leopard   Shanghai
+    8  3.0     1.500000            0.017964     lion      Basel
+
+## Example2: Rounding the first column to nearest third
```
ericmjl:
```suggestion
## Example 2: Rounding the first column to nearest third
```

examples/round_to_fraction.md
```diff
@@ -0,0 +1,92 @@
...
+    1  2.333333     0.285714            0.153846  leopard   Shanghai
+    2  3.333333     1.500000            0.017964     lion      Basel
+    3  1.333333     0.333333            0.004274   rabbit  Cambridge
+    4  2.333333     0.285714            0.153846  leopard   Shanghai
+    5  3.333333     1.500000            0.017964     lion      Basel
+    6  1.333333     0.333333            0.004274   rabbit  Cambridge
+    7  2.333333     0.285714            0.153846  leopard   Shanghai
+    8  3.333333     1.500000            0.017964     lion      Basel
+
+## Example3: Rounding the first column to the nearest third and rounding each value to the 10,000th place
```
ericmjl:
```suggestion
## Example 3: Rounding the first column to the nearest third and rounding each value to the 10,000th place 
```

tests/test_functions.py
```diff
@@ -37,6 +37,19 @@ def dataframe():
     return df
 
 
+@pytest.fixture
+def dataframe_floats():
```
ericmjl: I'm wondering if we can remove some duplication here.

If you change the `def dataframe()` fixture such that it returns all of the float values below, do tests pass? If so, I think there will be less code to maintain by making the `dataframe()` fixture return the contents of the `dataframe_floats()` fixture instead.

docs/round_to_fraction.rst
```diff
@@ -0,0 +1,124 @@
+df.round_to_fraction()
```
ericmjl: The examples guides have been great! I'm wondering about how it will render on readthedocs, and the maintainability of two separate files (one Markdown, one RST). Have you fiddled with sphinx before? Readthedocs uses Sphinx to build documentation, and you can build it locally too. In that way, there's fewer duplicate things to maintain.

tests/test_functions.py
szuckerman: I changed the dataframe fixture to include floats. Just had to change a few tests to reflect.

docs/round_to_fraction.rst
szuckerman: Sure, I'll work with that. I'm just more used to Markdown (since you can even use it in these textboxes), but RST is ok too.

billdenney:
@ericmjl and @szuckerman, When implemented, I added a bit more functionality than initially proposed: I allowed vectors for each of the arguments, and I added an automatic number of digitis.  See sfirke/janitor#282.

sfirke:
I love the interplay between these packages: the idea in the R repo, the implementation in Python, then revisions in the R version.

szuckerman:
@billdenney Thanks for checking in! 

I think we naturally have that functionality since all these functions are merely methods on pandas DataFrames. The method will always accept a vector as an input argument and shouldn't accept a vector for the denominator, which would duplicate data and cause a mismatch in rows.

ericmjl:
And @sfirke, to copy things you like here over too! :smile:

PR #46, opened 2018-11-12
remove special chars

shantanuo:
Added a few entries to replace list


ericmjl:
@shantanuo thanks for the additional changes! I just reviewed the tests, it appears that the tests you've added/modified cause others to fail. 

Could you take a look at the test cases that fail? There are two ways to do so.

You can run the tests locally if you have `pytest` installed. Be sure you are in the `pyjanitor/` directory (top-level), and then run `pytest` at the terminal.

Alternatively, you can click on the "details" link to see the output from Travis.

shantanuo:
Please accept the PR sent by the user "szuckerman" because his code is more pythonic. I have chained the "replace" method several times that was unnecessary.
He has also added "preserve_case" parameter to clean_names functions.
Is there any way to revert back my PR?

ericmjl:
@shantanuo that's very gracious of you! Ok, I will accept his PR instead. 

As for your PR, I just have to close it, and that'll be good enough. If you want, you can also delete your fork of pyjanitor, it won't affect my version.

It'll take me a little bit of time to do the release after merger. Hang tight!

Issue #15, opened 2018-04-17
limit column characters


shantanuo:
Is there any way to limit the length of column name? for e.g. this_is_very_long_column_needs_truncated should be truncated to left 10 or 20 characters. This may lead to duplicate column names - those should be appended by numbers like this_is_very_long_1 and this_is_very_long_2

ericmjl:
@shantanuo this functionality doesn't exist at the moment. I'm also not fully understanding your use case. Could you help me answer the following questions? Your answers will hopefully guide this discussion.

1. Is truncation desirable?
1. If truncation is desirable, then duplicate column names is a problem. Would suffixing `['_1', '_2', ...]` then be desirable?
1. If truncation is not desirable, then is duplicate column names still a problem?

shantanuo:
I would just incorporate the function "df_column_uniquify" from ...

https://stackoverflow.com/questions/24685012/pandas-dataframe-renaming-multiple-identically-named-columns

Limiting characters in column is important because when I use to_sql method to export pandas data to MySQL, I often get an error if the column names are too long :)

I do always use... 
df.columns = [list of corrected column names]

But your approach of using a class / module for this is much better. Thanks for sharing.

ericmjl:
@shantanuo I'm happy to accept a contribution from you for this functionality. (I'm available to respond to the issue tracker, but in the short term I have deadlines at work, which means I don't have the bandwidth at the moment to implement new features.)

shantanuo:
http://pandas.pydata.org/pandas-docs/version/0.19.2/whatsnew.html#read-csv-has-improved-support-for-duplicate-column-names

Assuming that the duplicate column names problem has been solved, I will suggest to make these changes...
1) Remove special characters like #$%
2) Remove one or more spaces by single dash 
3) Use lowercase
4) Allow max of 35 characters

I use this regular expression:

df.columns=df.columns.str.replace("[^\w\s]+","").str.replace("\s+","_").str.lower().str[:35]

I am not sure how to implement it in this module though :)

ericmjl:
@shantanuo I definitely have to get to writing documentation on how to make a contribution. In the meantime, if you'd like to contribute something, the best thing to do is to contribute a function to the `janitor.functions` module via a Pull Request. 

[...]

PR #47, opened 2018-11-13
Added remove_special and case_type to clean names

szuckerman:
I added a few parameters to the clean_names method. Namely:

1. I added a `remove_special` parameter that removes all special characters and only retains letters, numbers and underscores.
2. I changed `preserve_case` to `case_type`. This way I was able to add an 'upper' option (I like uppercase column names). This also allows it to be more extensible in the future if you want to add camelCase, Title_Case, etc.

These should close out Issue #15, except that there still isn't an option to choose the character spaces should be changed to. I think they should always be underscores, though, and the current default is fine.

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -52,7 +52,12 @@ def _strip_underscores(df, strip_underscores=None):
 
 
 @pf.register_dataframe_method
-def clean_names(df, strip_underscores=None, preserve_case=False):
+def clean_names(
+    df,
+    strip_underscores: str = None,
+    case_type: str = "lower",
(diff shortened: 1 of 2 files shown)


szuckerman:
It looked like the tests failed since I changed the `animals` column to `animals@#$%^`. For some reason I didn't have the `test_reorder_columns` in my `functions.py` and it was failing on that test (because it was looking for `animals`). Should hopefully be running fine now.

ericmjl:
Thank you, @szuckerman for the contributions! I Everything looks good, I will get this merged in.

PR #53, opened 2018-11-20
Add some documentation to docstrings

szuckerman:
I removed the `rst` files that were cluttering up the `docs/` folder and put them in the doctrings of the respective functions so they'll render on readthedocs.

I kept the markdown versions in `examples/` since I think those look nice on github, though.


ericmjl:
Before I merge, @szuckerman, would you be kind enough to build the docs locally and take a screenshot of what they will look like?

ericmjl:
I just wanted to make sure it looks correct, so that we don't have to do a follow-up commit later on to fix anything.

szuckerman:
pyjanitor_add_column

The above is an example from `add_column` of what the rendered docstrings look like.

szuckerman:
pyjanitor_add_column
Looks like the other file didn't upload correctly. Here's a jpg version.

PR #55, opened 2018-11-22
preserve original_columns

### LINKED ISSUE #54: Preserve original names
While renaming the dataframe, I need to preserve the original names. For e.g.

```
santandar_data = pd.read_csv(r"train.csv", nrows=40000)  
santandar_data.shape  

santandar_data.original_names=santandar_data.columns

ndf=santandar_data

ndf.original_names

Index(['ID', 'var3', 'var15', 'imp_ent_var16_ult1', 'imp_op_var39_comer_ult1',
       'imp_op_var39_comer_ult3', 'imp_op_var40_comer_ult1',
       'imp_op_var40_comer_ult3', 'imp_op_var40_efect_ult1',
       'imp_op_var40_efect_ult3',
       ...
       'saldo_medio_var33_hace2', 'saldo_medio_var33_hace3',
       'saldo_medio_var33_ult1', 'saldo_medio_var33_ult3',
       'saldo_medio_var44_hace2', 'saldo_medio_var44_hace3',
       'saldo_medio_var44_ult1', 'saldo_medio_var44_ult3', 'var38', 'TARGET'],
      dtype='object', length=371)

```
The ndf dataframe object has a property original_names that works correctly. But when I use clean_names function, I do not get this functionality.

```
df=santandar_data.clean_names(case_type="upper", remove_special=True).limit_column_characters(3)
df.original_names
```

> AttributeError: 'DataFrame' object has no attribute 'original_names'

shantanuo:
added 2 lines to clean_names function. At the start of the function, mylist=df.columns and at the end before returning df added df.original_columns = mylist.

Resolves #54.


janitor/functions.py
```diff
@@ -537,16 +539,12 @@ def fill_empty(df, columns, value):
         for col in columns:
             assert (
                 col in df.columns
-            ), "{col} missing from dataframe columns!".format(
-                col=col
-            )
+            ), "{col} missing from dataframe columns!".format(col=col)
```
ericmjl: This, I believe, was the result of using `black` without the 79 character limit.

I am going to push up a `pyproject.toml` file, so that we can standardize on 79 characters when using the `black` tool.

ericmjl: In the meantime, could you ensure that you use: `black -l 79` on the contents?

janitor/functions.py
```diff
@@ -97,6 +97,7 @@ def clean_names(
         Only letters, numbers and underscores are preserved.
     :returns: A pandas DataFrame.
     """
+    mylist = list(df.columns)
```
ericmjl: I'd suggest doing the following:

1. Firstly, make preservation of names optional, by putting it in the `clean_names` function signature.
2. Check to see if `preserve_names` is set too `True`. If yes, then store it.

janitor/functions.py
```diff
@@ -133,6 +134,7 @@ def _remove_special(col):
 
     df = df.rename(columns=lambda x: re.sub("_+", "_", x))
     df = _strip_underscores(df, strip_underscores)
+    df.original_names = mylist
```
ericmjl: With the aforementioned changes, this line would be unnecessary.

janitor/functions.py
```diff
@@ -537,16 +539,12 @@ def fill_empty(df, columns, value):
...
-            )
+            ), "{col} missing from dataframe columns!".format(col=col)
             df[col] = df[col].fillna(value)
     else:
         assert (
             columns in df.columns
-        ), "{col} missing from dataframe columns!".format(
-            col=columns
-        )
+        ), "{col} missing from dataframe columns!".format(col=columns)
```
ericmjl: Same as previous comment above.

janitor/functions.py
shantanuo: This line is necessary at the end because df.rename will replace original_names property.

janitor/functions.py
shantanuo: As suggested, I have added a parameter called original_columns. The default is True because it does not consume any resources. The user can disable it.

janitor/functions.py
```diff
@@ -57,6 +57,7 @@ def clean_names(
     strip_underscores: str = None,
     case_type: str = "lower",
     remove_special: bool = False,
+    original_columns: bool = True,
```
ericmjl: @shantanuo thanks for adding it in. However, when reading the API, with `original_columns`, it does not read as consistently as `preserve_original_columns`. (The more general idea is to provide an 'action term', which is what makes the API declarative.) 

The API really matters here, for that is pyjanitor's biggest contribution to the community. Could you change this to something that reads as an action term, e.g. `preserve_original_columns`, or `record_original_columns` or `backup_original_columns`? Do try to consider using the least ambiguous term across cultures, if possible.

janitor/functions.py
```diff
@@ -96,7 +97,9 @@ def clean_names(
     :param remove_special: (optional) Remove special characters from columns.
         Only letters, numbers and underscores are preserved.
     :returns: A pandas DataFrame.
+    :param original_columns: (optional) Preserve original column names.
     """
+    mylist = list(df.columns)
```
ericmjl: On matters of code style, I think `my_list` is not a sufficiently informative name. It's a matter of good Python coding practice to provide variable names that are descriptive. Could you change this line please?

Here's a suggested change:

```suggestion
    # Store the original column names, so that they can be retrieved later on.
    original_column_names = list(df.columns)
```

janitor/functions.py
```diff
@@ -133,6 +136,8 @@ def _remove_special(col):
 
     df = df.rename(columns=lambda x: re.sub("_+", "_", x))
     df = _strip_underscores(df, strip_underscores)
+    if original_columns:
```
ericmjl: Suggest adding a comment line here, because it is logically connected with the previous change, but is not connected in contiguous lines.

```suggestion
    # Store the original column names, if it is enabled by user.
    if original_columns:
```

janitor/functions.py
```diff
@@ -133,6 +136,8 @@ def _remove_special(col):
...
+        df.original_columns = mylist
```
```suggestion
        df.original_columns = original_column_names
```

janitor/functions.py
```diff
@@ -96,7 +97,9 @@ def clean_names(
     :param remove_special: (optional) Remove special characters from columns.
         Only letters, numbers and underscores are preserved.
     :returns: A pandas DataFrame.
+    :param original_columns: (optional) Preserve original column names.
```
```suggestion
    :param original_columns: (optional) Preserve original column names. This is later retrievable using `df.original_columns`.
```

ericmjl:
@shantanuo, thanks for being so patient about this! I will merge in!

PR #58, opened 2018-11-26
Currency conversion

szuckerman:
This PR resolves issue #_______.

# PR Checklist

1. [ ] PR in from a fork off your branch. Do not PR from <your_username>:master, but rather from <your_username>:<branch_name>.

## Quick Check

    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

- [ ] Ensure that you have added tests.
- [ ] Run all tests (`$ pytest .`) locally on your machine.
    - [ ] Check to ensure that test coverage covers the lines of code that you have added.
    - [ ] Ensure that all tests pass.

## Documentation Changes

- [ ] Build the docs locally.

# PR Description

- Added `convert_currency` method. It will convert a column from one currency to another.
- Added `convert_currency.md` to examples.
- Added the same example from `convert_currency.md` to the docstring of `convert_currency`.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -8,10 +8,13 @@ import re
 from functools import reduce
 from functools import partial
 from warnings import warn
+import json
+from datetime import date, datetime
 
 import pandas as pd
 import numpy as np
 from sklearn.preprocessing import LabelEncoder
+import requests
 
 import pandas_flavor as pf
 
@@ -1019,5 +1022,106 @@ def round_to_fraction(
     )
 
     df[colname] = df[colname].apply(_round_to_fraction_partial)
+
+    return df
+
+
+def _convert_currency(
+    from_currency: str = None,
+    to_currency: str = None,
+    historical_date: date = None,
+):
+    """
+    Currency conversion for Pandas DataFrame column.
+    The API used is: https://exchangeratesapi.io/
+    """
+
+    url = "https://api.exchangeratesapi.io"
+
+    if historical_date:
+        assert isinstance(historical_date, date) | isinstance(
+            historical_date, datetime
+        ), "historical_date must be a date or datetime object!"
+        if isinstance(historical_date, datetime):
+            assert historical_date >= datetime(
+                1999, 1, 4
+            ), "datetime must be later than 1999-01-04!"
+            string_date = str(historical_date)[:10]
+        else:
+            assert historical_date >= date(
+                1999, 1, 4
+            ), "date must be later than 1999-01-04!"
+            string_date = str(historical_date)
+        url = url + "/%s" % string_date
+    else:
+        url = url + "/latest"
+
+    currency_set = {
+        "AUD",
+        "BGN",
+        "BRL",
+        "CAD",
+        "CHF",
+        "CNY",
+        "CZK",
+        "DKK",
+        "EUR",
+        "GBP",
+        "HKD",
+        "HRK",
+        "HUF",
+        "IDR",
+        "ILS",
+        "INR",
+        "ISK",
+        "JPY",
+        "KRW",
+        "MXN",
+        "MYR",
+        "NOK",
+        "NZD",
+        "PHP",
+        "PLN",
+        "RON",
+        "RUB",
+        "SEK",
+        "SGD",
+        "THB",
+        "TRY",
+        "USD",
+        "ZAR",
+    }
+
+    assert (
+        from_currency in currency_set
+    ), f"from_currency must be one of {currency_set}"
+    assert (
+        to_currency in currency_set
+    ), f"to_currency must be one of {currency_set}"
+
+    payload = {"base": from_currency, "symbols": to_currency}
+
+    result = requests.get(url, params=payload)
+
+    assert result.status_code == 200, (
+        "Exchange Rate API failed to receive a 200 response from the server. "
+        "Please try again later."
+    )
+
+    currency_dict = json.loads(result.text)
+    rate = currency_dict["rates"][to_currency]
+    return rate
+
+
+@pf.register_dataframe_method
+def convert_currency(
+    df,
+    colname: str = None,
+    from_currency: str = None,
+    to_currency: str = None,
+    historical_date: date = None,
+):
+    rate = _convert_currency(from_currency, to_currency, historical_date)
+    df[colname] = df[colname] * rate
 
     return df
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -7,6 +7,7 @@ import datetime as dt
 import re
 from functools import reduce
 from functools import partial
+from functools import lru_cache
 from warnings import warn
 import json
 from datetime import date, datetime
@@ -1026,6 +1027,7 @@ def round_to_fraction(
     return df
 
 
+@lru_cache(maxsize=32)
 def _convert_currency(
     from_currency: str = None,
     to_currency: str = None,
@@ -1033,6 +1035,7 @@ def _convert_currency(
 ):
     """
     Currency conversion for Pandas DataFrame column.
+    Helper function for `convert_currency` method.
     The API used is: https://exchangeratesapi.io/
     """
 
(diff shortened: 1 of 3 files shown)


szuckerman:
The merge conflict appears to be with the `check` function that was added at the bottom of `functions.py`.

ericmjl:
@szuckerman thanks for putting in the PR. 

I'm wondering if currency conversion falls under the scope of `pyjanitor`? It seems to be a more finance/accounting domain-specific activity. In terms of user experience, having `convert_currency` can be convenient, but I'm thinking that it doesn't belong in the janitor top-level namespace, but in a sub-domain space.

To illustrate what I'm thinking, instead of:

```python
import janitor
df = (
    pd.DataFrame(...)
    .clean_names(...)
    .more_functions(...)
    .convert_currency(...)
)
```

I instead think it makes sense to have finance-related data cleaning routines in a finance submodule:

```python
import janitor
import janitor.finance

df = (
    pd.DataFrame(...)
    .clean_names(...)
    .more_functions(...)
    .convert_currency(...)
)
```

It's a minor change, but the biggest advantage here is that we can logically separate the namespaces, and when an end-user uses tab-completion, the finance-related ones will not show up unless explicitly requested.

Ultimately, I am hoping to position pyjanitor as a general purpose data cleaning tool; domain-specific namespaces might also be useful, and I'm open to having them in here, but I do want to ensure that things are architected thoughtfully first.

What are your thoughts on this architectural change? Let me know here, we can discuss.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
     df[colname] = df[colname].apply(_round_to_fraction_partial)
 
     return df
+
+
+@lru_cache(maxsize=32)
+def _convert_currency(
```
ericmjl: As mentioned in my overall comment, I'd suggest moving this into `pyjanitor/finance.py`.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+        "Please try again later."
+    )
+
+    currency_dict = json.loads(result.text)
+    rate = currency_dict["rates"][to_currency]
+    return rate
+
+
+@pf.register_dataframe_method
+def convert_currency(
```
ericmjl: Ditto with above function, suggest moving it into `pyjanitor.finance`.

tests/test_functions.py
```diff
@@ -619,3 +619,20 @@ def test_round_to_nearest_half(dataframe):
     assert df.iloc[6, 1] == 1.0
     assert df.iloc[7, 1] == 2.5
     assert df.iloc[8, 1] == 3.0
+
+
+def test_make_currency_api_request(dataframe):
+    """This test assumes that the GBP will always be more than the USD."""
```
ericmjl: From a testing engineering perspective, this is a very big assumption to make, even though it is reasonable in our daily experience.

I wonder what you might think about the following test instead?

This test ensures that the API provider is live. This way, in automated nightly builds, if it goes down, we can get notified. It might be good to have a backup exchange rate API as well. 

```python
r = requests.get('https://api.exchangeratesapi.io')
assert r.status_code == 200
```

This one might be the most crucial for the functionality of `convert_currency`.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+    """
+    Currency conversion for Pandas DataFrame column.
+    Helper function for `convert_currency` method.
+    The API used is: https://exchangeratesapi.io/
+    """
+
+    url = "https://api.exchangeratesapi.io"
+
+    if historical_date:
+        assert isinstance(historical_date, date) | isinstance(
```
ericmjl: I think I changed the type-checking assert statements to use the `check` function instead. It's cleaner, and raises a TypeErrror rather than an AssertionError, which is clearer, and in case someone runs Python while disabling asserts (it's possible), the type-checking errors will be raised.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+            historical_date, datetime
+        ), "historical_date must be a date or datetime object!"
+        if isinstance(historical_date, datetime):
+            assert historical_date >= datetime(
```
ericmjl: I'd encourage you to change this to:
```suggestion
            if historical_date < datetime(1999, 1, 4):
                raise ValueError("datetime must be later than 1999-01-04!")
```

For the same reasons as above.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+                1999, 1, 4
+            ), "datetime must be later than 1999-01-04!"
+            string_date = str(historical_date)[:10]
+        else:
+            assert historical_date >= date(
```
ericmjl: Same requested change as above.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+        "RUB",
+        "SEK",
+        "SGD",
+        "THB",
+        "TRY",
+        "USD",
+        "ZAR",
+    }
+
+    assert (
```
ericmjl: I'd like to suggest providing a `check_currency()` helper function.

```python
def check_currency(currency):
    if currency not in currency_set:
        raise ValueError(f"currency {currency} not in supported currency set, {currency_set}")
```

This would allow refactoring:

```suggestion
    check_currency(from_currency)
```

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+        from_currency in currency_set
+    ), f"from_currency must be one of {currency_set}"
+    assert (
```
```suggestion
    check_currency(to_currency)
```

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+        to_currency in currency_set
+    ), f"to_currency must be one of {currency_set}"
+
+    payload = {"base": from_currency, "symbols": to_currency}
+
+    result = requests.get(url, params=payload)
+
+    assert result.status_code == 200, (
```
ericmjl: Also suggest changing this to some other kind of error, e.g. ConnectionError.

```suggestion
    if result.status_code != 200:
        raise ConnectionError("Exchange Rate API failed to receive a 200 response from the server. "
        "Please try again later."
        )
```

janitor/functions.py
ericmjl: Uses the refactored value check.

tests/test_functions.py
```diff
@@ -619,3 +619,20 @@ def test_round_to_nearest_half(dataframe):
...
+
+    df = dataframe.convert_currency("a", "USD", "GBP")
+
+    assert df.iloc[0, 0] < 1
+
+
+def test_currency_makes_column_decrease(dataframe):
+    """This test assumes that the GBP will always be more than the USD."""
```
ericmjl: This test likewise makes a very large assumption about exchange rates, one that cannot be guaranteed to be true all the time. I'd suggest removing this one, or replacing it with a test that checks the dtype of the resultant column.

janitor/functions.py
```diff
@@ -1021,3 +1025,171 @@ def _round_to_fraction(number, denominator, digits=np.inf):
...
+            4  2.048056     0.285714            0.153846  leopard   Shanghai
+            5  2.697084     1.500000            0.017964     lion      Basel
+            6  1.029370     0.333333            0.004274   rabbit  Cambridge
+            7  2.048056     0.285714            0.153846  leopard   Shanghai
+            8  2.697084     1.500000            0.017964     lion      Basel
+
+        """
+
+    rate = _convert_currency(from_currency, to_currency, historical_date)
+    df[colname] = df[colname] * rate
```
ericmjl: What are your thoughts about adding an optional keyword argument that allows a user to create a new column suffixed with the converted currency?

ericmjl:
As for the conflicts, I think they will go away if you move the newly proposed functions into a `finance.py` module.

tests/test_functions.py
szuckerman: I had the same concerns when writing the tests as well. You're right that the 200 check should be sufficient.

Maybe instead of using another API (since this appeared to be one of the few that was totally free) we can hard-code the rates from a certain day and fall back to that if the API fails. For example, use something like 'exchange rate on first day of most recent quarter' or 'average exchange rate over the past week'. 

Or we can still use the above hard-coded solution and use other APIs as well. We can add a parameter for people to choose which method/API is required and have an order of succession if they fail.

janitor/functions.py
szuckerman: You read my mind.... Literally thought of that after submitting the PR.

janitor/functions.py
szuckerman: I like it, but maybe call it `_ check_currency` to indicate that it's private?

tests/test_functions.py
ericmjl: > Maybe instead of using another API (since this appeared to be one of the few that was totally free) we can hard-code the rates from a certain day and fall back to that if the API fails.

Not a bad idea. What about this other idea? Provide a function that caches the currency exchange rates for just the current day, which can then be read if offline? That would be more work, though, and I'd prefer for it to be a separate PR, to give us more time to think about it, and also prevent this one from mushrooming. :smile:.

In response to my own thought:

> It might be good to have a backup exchange rate API as well.

On second thought, this might not be a good idea, since this the one you've put up is a free service. To keep things simple, it may be better to just use this API for now.

janitor/functions.py
ericmjl: @szuckerman can do too! Though of course, there's no truly private functions in Python, but yes, the prefixed underscore is a good indicator.

ericmjl:
@szuckerman feel free to move ahead with moving this into finance. If you pull the latest master, you'll see a template of how I did it for biology.

janitor/functions.py
szuckerman: Makes sense.. Changing asserts to use `check` instead now.

PR #64, opened 2018-11-28
63 add column str fix

### LINKED ISSUE #63: add_column sanity checking broken for strings
There's a sanity check I put in there where if the supplied value is a sequence, it makes sure that the length is the same as the number of rows in the `DataFrame` as long as `fill_remaining` is false. This is done by checking for the existence of `__len__`. I need to add logic to exclude this checking for `str` objects. Fixing now.

zbarry:
@ericmjl 

closes #63

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -964,7 +964,9 @@ def add_column(df, col_name: str, value, fill_remaining: bool = False):
 
     nrows = df.shape[0]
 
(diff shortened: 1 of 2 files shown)


zbarry:
Going to lunch. Don't merge yet

ericmjl:
LGTM! Merging and closing.

PR #66, opened 2018-11-28
Add min_max_scale

ericmjl:
# PR Checklist

1. [x] PR in from a fork off your branch. Do not PR from <your_username>:master, but rather from <your_username>:<branch_name>.

## Quick Check

    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# PR Description

- Adds in a min_max_scale function that scales either a single column or a dataframe to the bounds of [0, 1]. 

# Relevant Reviewers

- @ericmjl


zbarry:
Maybe make it scale to user-specifiable min/max? Formula for an array is:

`arr_new = (arr_old - arr_old.min()) * (new_max - new_min) / (arr_old.max() - arr_old.min()) + new_min`

(caching the min/max calculations of course)

https://en.wikipedia.org/wiki/Normalization_(image_processing)

ericmjl:
Sounds like a good idea.

zbarry:
Spelling error in commit message. Gotta scrap it.

zbarry:
In docstring:

> If minimum and maximum are provided, then the data are linearly scaled to be between the two values. Otherwise, they are scaled to be between the minimum and maximum of the value.

Don't you mean if not provided, they are scaled from 0 to 1?

ericmjl:
> Spelling error in commit message. Gotta scrap it.

No. 😛

>

> In docstring:
> 
> > If minimum and maximum are provided, then the data are linearly scaled to be between the two values. Otherwise, they are scaled to be between the minimum and maximum of the value.
> 
> Don't you mean if not provided, they are scaled from 0 to 1?

Good catch. Updating docstrings.

zbarry:
sec

zbarry:
https://gist.github.com/zbarry/f0e377b3c4aaea355d53df782c0ce89f

Suggested docstring.

I'd also consider `s/minimum/old_min`, `s/maximum/old_max`, `s/new_minimum/new_min` ...

zbarry:
The isoelectric point example should probably be expounded on a bit as to what you mean exactly

zbarry:
Would it make sense to raise an exception if max > min?

zbarry:
What about the case where someone tries to min/max a `DataFrame` with some (or all) non-numerical columns?

It might actually be useful to gracefully handle detecting which things are numeric.

zbarry:
Non-abbreviated minimum / maximums are still in dstring in the description part

zbarry:
Either way, probably would want to include a test for an exception being raised if the user tries to min/max a particular column that's not numeric

ericmjl:
Dood, you really could do with consolidating your comments into a single box dude. Makes things easier to track. Just edit your old comments. Stream-of-consciousness doesn't facilitate discussions, which is what the issue tracker and PR boards are intended for.

> The isoelectric point example should probably be expounded on a bit as to what you mean exactly.

Done.

> What about the case where someone tries to min/max a DataFrame with some (or all) non-numerical columns? It might actually be useful to gracefully handle detecting which things are numeric.

That should be for a later PR.

> Would it make sense to raise an exception if max > min?

Good idea.

zbarry:
Lol whoops.

ericmjl:
Ok, I'm ready to merge. We can make the function cover a broader use case later.

PR #81, opened 2018-12-06
Make currency column numeric function

szuckerman:
This PR resolves #(put issue number here).

# PR Checklist

## Quick Check

    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [X] Build the docs locally.

# PR Description

I added a new function called `make_currency_column_numeric`. It basically takes a column that should be a float, but was stored as `str` because of dollar signs, etc.

It also includes: 
- Removing 'accounting' style fields (i.e. the numbers with parentheses, like (1.00) ) 
- Casting words to numeric values that may be found in the column
- Removing rows where the value is a word or other string

# Relevant Reviewers

- @ericmjl


janitor/functions.py
```diff
@@ -1599,3 +1605,297 @@ def check(varname: str, value, expected_types: list):
...
+        return x
+
+
+def _replace_original_empty_string_with_none(x):
+    if x != "ORIGINAL_NA":
+        return x
+
+
+@pf.register_dataframe_method
+def make_currency_column_numeric(
```
ericmjl: @szuckerman pardon me for being nitpicky here, but I'm wondering if there's a more expressive verb than "make"? 

Java-land is littered with getters and setters because, and I'm wondering if we can avoid the same fate with makers.

In addition to that, since this is a currency-related function, does it make more sense to put this in the `finance.py` submodule?

janitor/functions.py
```diff
@@ -1599,3 +1605,297 @@ def check(varname: str, value, expected_types: list):
...
+    df,
+    col_name: str,
+    type: str = None,
+    cast_non_numeric: dict = None,
+    fill_all_non_numeric: float = None,
+    remove_non_numeric: bool = False,
+):
+    """
+    :param df: The DataFrame
```
ericmjl: Python docstrings usually should have a one-liner summary of what it is doing.

In pyjanitor, the rest of the functions also describe in words what is being accomplished.

Could you add those two elements please?

szuckerman:
I was considering putting this in the `finance` submodule, but thought it was too common of a use case just to have there.

janitor/functions.py
szuckerman: Sure; I'm not married to any of the names. 

What do you think of:
1. `convert_currency_column_to_num`?
2. `currency_column_to_numeric`?
3. `currency_column_to_num`?

janitor/functions.py
szuckerman: Whoops; added all the examples but not the summary!

janitor/functions.py
szuckerman: I went with `2`.

ericmjl:
> I was considering putting this in the `finance` submodule, but thought it was too common of a use case just to have there.

If so, then should the function name be perhaps changed to a more generic one, rather than a currency column?

szuckerman:
Maybe the name should be more general, but I was having problems coming up with something that wouldn’t be misleading. 

For example, to_numeric has a specific definition in most APIs and something like “ 0.04” would fail because of the white space. 

Similarly, I feel that calling it something like “string_to_numeric” is also misleading since that might imply that the function converts words to numeric values (like “Gold” —> 3, “Silver” —> 2, etc).

ericmjl:
@szuckerman understood! 

I'm definitely getting a good feeling about the utility of the function; I just want it to have a good name (because the name corresponds to the API, and it's hard to change the API once people become used to it).

You know, I have an alternate idea, let's see what you think about this:

There are some pretty good ideas baked into the function. Let's put this in the finance submodule for now. What we can do is refactor out the nice generalizable parts later on. Doing this will give us a domain-specific name that maps nicely to a very specific finance-related task, while buying us time to think of the generalizable name later. Maybe what we need is some non-finance example use cases, really, to clarify what the generalized name will be.

What are your thoughts?

szuckerman:
(Sorry for being MIA for a few days)

I think you're right; we can leave it in the `finance` submodule for now until it's more generalized.

However, my problem now is naming the function. `string_column_to_numeric`? That almost sounds like I have a column of words that I'm going to map to numeric values. 

If we generalize the function I think that having a `type="currency"` argument similar to the `type="accounting"` argument would be useful.

I think that this is a good example of what might be the threshold for including in the main package or the submodules. The main package should ensure that a DataFrame is in a state to run pandas functions without any extra work. If a number column is encoded as `str`, there's extra work involved. 

Whereas for the submodules it's more specific to the use-case. For example, and I'm not saying we should implement this but just a thought, for the `finance` submodule there could be a `moving_average` function for a column. I know there's some gray area, but the overall idea seems to make sense, thoughts?

szuckerman:
The travis build failed on a line with whitespace :(

ericmjl:
> (Sorry for being MIA for a few days)

No worries, @szuckerman! Good software is developed thoughfully! 

> However, my problem now is naming the function. `string_column_to_numeric`? That almost sounds like I have a column of words that I'm going to map to numeric values.

In finance, is there a term that accountants/financial people use? A verb that would match their mental model? I think that’ll give us a good verb (let’s try to keep it to two verbs generally) that can describe succinctly the action being taken.

> If we generalize the function I think that having a `type="currency"` argument similar to the `type="accounting"` argument would be useful.

:thumbsup: Let’s see whether we or other users encounter a general use case for the function. That way it’ll solve someone’s actual need. Amazon-style (listen to our customers, which in our case, includes ourselves).

ericmjl:
@szuckerman damn, I think i've left this PR hanging for too long. Is it ready to merge? Would you be kind enough to update the relevant branch from master? Let's make this the last thing we PR into `master`.

PR #83, opened 2018-12-09
changed transform to transform_column in docstring to match func name

zbarry:
Old docstring name for `transform_column` remained after function name was apparently changed.

@ericmjl


ericmjl:
LGTM, thanks @zbarry!

Issue #79, opened 2018-12-06
reorder_columns does not mutate the original dataframe


zbarry:
I'm thinking about modding it to do so to be consistent with everything else I implemented.

zbarry:
@ericmjl @szuckerman - I updated this post and would like your thoughts.

ericmjl:
@zbarry I'm glad you brought this up, this has also been on my mind for a while. 

I'm in favour of standardizing, but like the `inplace=True/False` issue, it means we may end up having to place a burden on new contributions to ensure that the functions conform to what we decide now. Both standardizing on mutating the original and not mutating the original will result in technical debt. We just have to decide which one we want more. Unlike the `inplace=True` issue, this is one that is related to the performance of `pyjanitor`, rather than the public-facing API. 

Let me write down some of my thoughts at this point.

As we all probably know, standardizing on not mutating the original dataframe probably means that we will likely incur a memory and performance cost for large dataframes. Every function would then have to explicitly copy the original dataframe and only make modifications on the modified one. 

Standardizing on mutating the original dataframe might be technically more long-winded, which makes development just a teeny bit more difficult. For example, I'm not quite sure how to implement `.coalesce()` in a way that mutates the original. Maybe I'm not well-versed in the pandas API.

In terms of what users will notice, speed is definitely a thing. However, I think of the pandas userbase as being pareto distributed, with 20% dealing with @zbarry's scale of data. The other 80% deal with in-memory data. 

@szuckerman, do let us know what you think here. With 3 of us, we won't need a tiebreaker vote. I'm happy to go both ways, though my current inclinations lean towards a lazier form of development.

zbarry:
Happy to pick up the slack re: technical debt ;)

zbarry:
@ericmjl @szuckerman 

Alright, I had some inspiration last night and think I might have come up with a user-friendly, explicit approach to operating in place when working with pyjanitor functions.

## To somewhat sum up the previous discussion:

- It's really confusing that some janitor functions mutate the original dataframe, and others do not
- For contributors, low barrier-to-entry to implementing their desired functionality is beneficial - if they don't think in place for speed / efficiency is necessary, one has to consider the extra burden enforcing in place by default might impose on them
- For high-performance applications, dataframe copying is a big performance problem
- In my own experience, I've accidentally mutated dataframes I did not mean to. "Explicit is better than implicit" is not happening here.

## Proposed solution

The pyjanitor modus operandi would be as follows:

[...]

Issue #73, opened 2018-12-01
Jupyter notebook example of pyjanitor usage


zbarry:
I kind of want to put something together to show usage, talk about philosophy of pyjanitor, etc.

zbarry:
👍

PR #85, opened 2018-12-09
Demo pyjanitor walkthrough notebook

zbarry:
#73 

@ericmjl 

Here is a first draft of the notebook for discussion.

You can see the effects of #79 in the cells where I break down the data cleaning pipeline step-by-step progressively across cells, where some are just `df.thing()` and others are `df = df.thing()` because of the in place inconsistency.


ericmjl:
@zbarry totally in-line with what I was thinking! Nice stuff. Anything you think needs discussing? From my reading of it, LGTM!

zbarry:
👍 gtg then

ericmjl:
What's your Twitter account?

PR #110, opened 2019-01-28
updated add_column tests to use hypothesis

ericmjl:
As per title. Does not resolve any issue.


ericmjl:
One thing nice about using Hypothesis here is it created a test example that was not considered in the original. (No blame here - just the goal of continuous improvement!)

PR #117, opened 2019-01-30
Filter date

### LINKED ISSUE #105: Add filter_date function
This issue is to track creating a `filter_date` function, or whatever else you might want to call it.

The main idea is that even though there are already `filter` functions, date filtering is a bit more complex. For example, when filtering strings you can just write a string, but for dates it usually needs to be a `date` or `datetime` object, and not all of them play together nicely.

More important than the above is that when filtering a range, the lines can get really long, like so:

```python
start_date = date(2017, 1, 1)
end_date = date(2018, 1, 1)
df = df[(df.date_column >= start_date) & (df.date_column <= end_date)]
```

I think something like the following is nicer:

```python
df = df.filter_date('date_column', start='2018-01-01', end='2018-01-01')
```
It could also include arguments for `year`, `years`, etc.

szuckerman:
This PR resolves #105 .

# PR Checklist

## Quick Check

    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

- [X] Ensure that you have added tests.
- [X] Run all tests (`$ pytest .`) locally on your machine.
    - [X] Check to ensure that test coverage covers the lines of code that you have added.
    - [ ] Ensure that all tests pass. (well, not the conda ones, as mentioned)

## Documentation Changes

- [X] Build the docs locally.

# PR Description

- This function is a nice date filter function for DataFrames.
- One issue, though, is the last line `return df.loc[_date_filter_conditions(_filter_list), :]` gives a `SettingwithCopyWarning` and the way to avoid it would be to return a copy of the filter with `.copy()`, but I know we've shied away from doing things like that in the past. I figure it's easy enough to add later if it becomes an issue.

# Relevant Reviewers

- @ericmjl
- @zbarry


janitor/functions.py
```diff
@@ -836,6 +837,234 @@ def filter_on(df, criteria, complement=False):
...
+        25      26 2020-03-11
+    """
+
+    check("column", column, [str])
+
+    def _date_filter_conditions(conditions):
+        """
+        Taken from: https://stackoverflow.com/a/13616382
+        """
+        return reduce(np.logical_and, conditions)
```
ericmjl: SMAHT!

ericmjl: Said with my most butchered Bostonian accent.

ericmjl:
Looks great, thanks @szuckerman!

janitor/functions.py
zbarry: 🤦‍♂️

PR #127, opened 2019-02-10
update py3 dep to require 3.6 or later rather than specifically 3.6

zbarry:
https://github.com/ericmjl/pyjanitor/issues/126

Just went ahead and did this because it was so trivial; not trying to bypass discussion.

@ericmjl @szuckerman


ericmjl:
Trivial indeed. LGTM.

Issue #111, opened 2019-01-28
Documenting how tests are organized


ericmjl:
Heads-up, @zbarry and @szuckerman.

I have refactored the test suite a little bit (actually, quite a lot). The changes hopefully make the test suite easier to develop against and enable newer contributions. I will be writing a section on how to write tests (as part of the CONTRIBUTING.rst) file

The key changes here are:

- Each function gets tested in a `test_<function_name>.py` file. This structure allows us to create multiple tests for each function, without cluttering up one big `test_functions.py` file.
- Testing utilities, such as fixtures and hypothesis strategies, are now part of the main library. This enables them to be imported into the tests instead. Side note: We do not need to have tests for the testing utilities.

Big thanks to both of you for your contributions thus far, I really appreciate the work that has gone in :smile:.

szuckerman:
I really like the refactoring of the test suite (especially using the Hypothesis module; it looks awesome).

However, to run the tests it just took me about 30 seconds while, in the past, I remember it being around 7 seconds or so.

I think all the `import`s at the top might be unnecessarily slowing it down. There's over 30 files, which means `hypothesis` and `pandas` could be loaded that many times.

I think a master file with just one `import pandas as pd` and a `from tests import *` might work.

szuckerman:
@zbarry yes, but I think it depends how the files are run. I pass the `tests/` folder to `pytest` which moves through each file. If `pytest` is running each file separately then the imports may be an issue.

ericmjl:
@szuckerman the slowdown in testing is a natural consequence of using Hypothesis. Previously, we were using only one example. Hypothesis is dynamically generating up to (I think) 50-200 examples, looking for a way to falsify each test. 

I was able to find some functions for which hypothesis helped discover an example that I wasn't originally thinking of, but could break the function and needed to be checked for inside the function. (e.g. a zero-length list being passed into the function)

zbarry:
Gotcha. Is there a slick way to be able to get Hypothesis to only operate on a subset of tests when developing to get a fast iteration time, and at the end, we run the whole suite before pushing?

ericmjl:
I'd be judicious with the use of hypothesis, btw. For example, I don't know how to generate examples well for the biology functions. 

Going forth, it's fine to use an example-based test (our original paradigm) rather than a property-based test (using Hypothesis). 

To selectively run a test, use pytest.mark to label the test. You'll see a few examples in the test suite:

```python
[...]

PR #128, opened 2019-02-11
Test refactoring

szuckerman:
Regarding the test refactoring (#111):

A few changes:

1.  Apparently, best practices is to put `fixtures` for `pytest` in a `conftest.py` file. The tests will automatically catch all the fixtures there; we don't need to explicitly import then.

1. We can access global variables across tests through the following (in `conftest.py`):

```python
TEST_DATA_DIR = 'tests/test_data'
EXAMPLES_DIR = 'examples/'

def pytest_configure():
    pytest.TEST_DATA_DIR = TEST_DATA_DIR
    pytest.EXAMPLES_DIR = EXAMPLES_DIR
```

Now, wherever `pytest` is defined, `pytest.TEST_DATA_DIR`, etc are defined as well. This is helpful for those files that need test files.

3. I moved the `import` for `rdkit` in the `test_chemistry.py` function within a test so I can skip the test when running the whole test suite. I'm still having issues with `conda` and it's just easier for me to do it that way, but understand if you want the `import` at the top of the file for consistency.

- @ericmjl
- @zbarry


ericmjl:
Great stuff, thanks for the help here, @szuckerman!

Issue #118, opened 2019-01-30
Release cycles


szuckerman:
Even though Issue #3 discusses the 1.0 release to PyPI, I'm wondering about how you would want to organize occasional releases.

(Assuming the X.Y.Z numbering syntax)

I'm not sure if we should be bumping up the "Y" number when adding a new function. 

I'm thinking something like the following:
- When there's refactoring or a new function added, update the Z number
- If a new submodule is added, update the Y number
- If a significant amount of changes have happened in a while (not sure how to quantify yet) update the X number

I also think that it might be good to save the `master` branch as the "this is what goes to PyPI" branch and have a `dev` branch that gets the "every so often" updates. [Git has a useful tagging mechanism for when you want to note a specific version in the commit history.](https://git-scm.com/book/en/v2/Git-Basics-Tagging)

ericmjl:
@szuckerman for versioning, this is a great idea! I have been extremely sloppy in this regard.

I think the proposal that you have for versioning is reasonable, so let's go with that. Any further debate on this one would be an exercise in minutiae, I think. 

Your other PR will be the first time I bump up the Z number instead of the Y number.

Tagging is a very good idea as well - IIRC, every time I run bumpversion, it is auto-tagged.

As for master/dev split, it's definitely a good idea. For newcomers, would it be a tad confusing, though? I already had to guide @jcvall through the git workflow once - it would be a barrier to contribution if now newcomers had to also deal with different branches too. 

I can see doing 'master => dev', with 'release => master' as the alternative, and the default branch being still set to 'master', but that would also be confusing for experienced devs XD.

zbarry:
Do you think it would be a big issue? I would guess you'd just put in your `CONTRIBUTING.md` that you should branch from `dev` instead of `master` when you begin to make your changes. Don't know if that's too much more additional mental load / barrier to entry.

zbarry:
yah

zbarry:
https://github.com/ericmjl/pyjanitor/tree/dev

Also realized that given this new branch protocol, tagging each significant merge into `master` with the new PyPI version might be in order.

ericmjl:
Let me go in and protect the master branch.

ericmjl:
Guys, done. New default branch for new PRs is dev. Also protected `master` such that it requires an explicit PR. 

@szuckerman and @zbarry thanks for taking the initiative on this. In retrospect, I think this is better; perhaps my old way of thinking was too simplistic :smile:.

[...]

PR #129, opened 2019-02-12
updated contributing w.r.t. a proposal to introduce a dev-master merge split

zbarry:
First draft addressing proposal in #118. 

@ericmjl @szuckerman

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -101,7 +101,11 @@ Ready to contribute? Here's how to set up `pyjanitor` for local development.
 
 4. Create a branch for local development::
 
-    $ git checkout -b name-of-your-bugfix-or-feature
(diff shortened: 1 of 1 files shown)


ericmjl:
LGTM! @szuckerman, if you're good with the documentation, then feel free to press the big green button!

zbarry:
Once merged, we're gonna want to just create `dev` and push it.

zbarry:
image

Strange error with python 3.5 / python 3.7 references in the same log.

ericmjl:
I think a simple restart of the job should do the trick. Do you guys have the necessary access privileges to do so?

zbarry:
> I think a simple restart of the job should do the trick. Do you guys have the necessary access privileges to do so?

No idea how I'd even check

PR #131, opened 2019-02-12
Add then function

### LINKED ISSUE #130: "then" function
I had an idea for a function that allows for users to run any arbitrary function in the method chain.

For example, I think it's annoying if someone has their own logic and needs to do this:

```python
def custom_logic(df):
    ...
    return df

df = (df.clean_names()
    .add_column('a_column'))

df = custom_logic(df)
```

Instead, people should be able to do the following:

```python
def custom_logic(df):
    ...
    return df

df = (df.clean_names()
    .then(custom_logic)
    .add_column('a_column')
```

szuckerman:
This PR resolves #130 
# PR Checklist

## Quick Check

    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

- [X] Ensure that you have added tests.
- [X] Run all tests (`$ pytest .`) locally on your machine.
    - [X] Check to ensure that test coverage covers the lines of code that you have added.
    - [X] Ensure that all tests pass.

## Documentation Changes

- [X] Build the docs locally.

# PR Description

- This allows people to add arbitrary functions to the method chain.

# Relevant Reviewers

- @ericmjl
- @zbarry

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2408,3 +2408,21 @@ def impute(df, column: str, value=None, statistic=None):
     if value is not None:
         df[column] = df[column].fillna(value)
     return df
+
+
+@pf.register_dataframe_method
+def then(df: pd.DataFrame, func) -> pd.DataFrame:
+    """
+    :Description:
+
+    Add an arbitrary function to run in the pyJanitor method chain.
+
+    :param df: A pandas dataframe.
+    :param func: A function you would like to run in the method chain.\
+    It should take one parameter and return one parameter, each being the\
+    DataFrame object. After that, do whatever you want in the middle.\
+    Go crazy.
+    """
+
+    df = func(df)
+    return df
--- a/examples/then.md
+++ b/examples/then.md
@@ -1,0 +1,97 @@
+# df.then()
+
+## Description
+This method allows for arbitrary functions to be used in the pyJanitor method-chaining fluent API.
+
+## Parameters
+### df
+A pandas dataframe.
+     
+### func
+A function that takes one parameter and returns that same parameter. For consistency it should be `df` to indicate that it's the Pandas DataFrame.  After that, do whatever you want in the middle. Go crazy.
+     
+
+## Setup
+```python
+import pandas as pd
+import janitor
+ 
+data_dict = {
+    "a": [1.23, 2.45, 3.23] * 3,
+    "Bell__Chart": [1, 2, 3] * 3,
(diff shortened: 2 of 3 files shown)


szuckerman:
I know we want to start adding to the `dev` branch, but I couldn't find a `dev` branch yet. Feel free to close this if you'd rather have it going there.

ericmjl:
@szuckerman @zbarry both of you should have full admin privileges on the repo. I'm trusting you guys with a ton here :smile:. You can be our pioneer, @szuckerman!

janitor/functions.py
```diff
@@ -2408,3 +2408,21 @@ def impute(df, column: str, value=None, statistic=None):
     if value is not None:
         df[column] = df[column].fillna(value)
     return df
+
+
+@pf.register_dataframe_method
+def then(df: pd.DataFrame, func) -> pd.DataFrame:
```
ericmjl: Wondering if we should allow for `*args, **kwargs` in here?

zbarry: +1

szuckerman: I don't think so, since the purpose is to pass a one-argument function to `then`. It's similar to `apply` where any parameters or such should be done in the referenced function.

Unless you guys think otherwise.

ericmjl: I would personally like it, but then I think to prevent premature overengineering, I'd just merge this version first, and we can add in args/kwargs later.

ericmjl:
Ok, going to merge, @szuckerman!

janitor/functions.py
zbarry: Could always do:
```python
def func(df, thing1, thing2):
    dostuff()

df.blah().then(lambda df: func(df, 1, 'asdf'))
```

if you want to use extra args for now.

szuckerman: So I'm a bit torn...

I like that it doesn't have arguments so it's forced to read like English: `.then(do_this_to_the_df)`, but arguments obviously makes it more flexible.

However, one can always use `functools.partial` to coerce a function to a one-argument function (as I usually do with `pd.Series.apply`).

So the trade off would be: arguments in `.then()` or force people to use `functools.partial`?

ericmjl: `functools.partial` is pretty intermediate-level stuff!

ericmjl:
Hohoho, I just used this function today, @szuckerman :smile:

PR #133, opened 2019-02-21
Rolling PR for 0.16.5

ericmjl:
Rolling PR for 0.16.5. Any new merges into `dev` will automatically be reflected here.

WIP in title ensures we're always reminded that this is not ready to go until we're ready to cut a new release!


ericmjl:
Friends, I'm itching to merge this into master and cut a release.

Going forth, I think we should just cut a new release for each new logical feature added. I'll just bump the patch version for logically small things.

PR #140, opened 2019-03-17
Transform columns

ericmjl:
This PR introduces a plural function for `transform_column`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ docs:
 isort:
 	isort -r . -y -up -tc
 
-check: test docs isort format
+check: test docs isort format lint
 	echo "checks complete"
 
 style: isort format
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -4,9 +4,9 @@ General purpose data cleaning functions.
 import datetime as dt
 import re
 import warnings
+from fnmatch import translate
 from functools import partial, reduce
 from typing import Dict, Iterable, List, Union
-from fnmatch import translate
 
 import numpy as np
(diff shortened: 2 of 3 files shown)


ericmjl:
@szuckerman only the currency API request test is failing here. Can you see what might be going on? If that can't be solved, I'd be inclined to have you just hit the merge button :smile:.

szuckerman:
It looks like there's something going on with redirects and the currency API.

There are some solutions mentioned on [this stack overflow question](https://stackoverflow.com/questions/23651947/python-requests-requests-exceptions-toomanyredirects-exceeded-30-redirects), but I'm wondering if it's not just something that will resolve itself?

It appears that the easiest solution is to add a header with the request.

ericmjl:
> It appears that the easiest solution is to add a header with the request.

Agreed, after reading the SO post and response.

Let's add that in a separate PR. I'll raise an issue.

szuckerman:
After looking into it more, I'm reluctant to add the `User-Agent` header since:

1. It looks like the build is passing now
2. I don't want to force the `User-Agent` to be some set agent (like Chrome, etc) if someone is using something else and put that in the code. If someone wants to spoof it himself, that's his prerogative. Similarly we could just use "pyjanitor" as a user-agent, but that might make more problems in the future and might not be a best practice.
3. It might be just an Amazon issue from how the stack overflow question was phrased. The `User-Agent` might not address our issue.

ericmjl:
I'll trust your best judgment on this!

PR #163, opened 2019-05-06
Update requirements.txt

HectorM14:
Testing failed since I didn't have `hypothesis` and `requests` in my environment. `hypothesis` version was taken from the `.yml` file. `chemistry.py` also fails since I don't have `rdkit`, but that seems too niche and difficult to install, so I assume it's okay for those tests to fail for the rest of us.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,6 @@ scikit-learn>=0.19.1
 xlrd>=0.9.0
 pandas-flavor==0.1.2
 biopython
+requests
+hypothesis>=4.4.0
+


ericmjl:
> chemistry.py also fails since I don't have rdkit, but that seems too niche and difficult to install, so I assume it's okay for those tests to fail for the rest of us.

Totally agree, don't worry about those tests locally. They will pass on Travis, though.

------

Currently, Travis is failing because it's doing `pip install -e .` rather than `python setup.py develop`. Can you pull `dev` into your branch so that the Travis build script is correct?

By the end of this sprint at PyCon, we will be moving off Travis to Azure Pipelines, which are more user-friendly for multiple OSes.

ericmjl:
Meanwhile, I wouldn't worry about the tests failing on this PR, so I will go ahead and merge.

HectorM14:
Great. Should I have PR'd into `dev` instead of `master`?

When is the sprint? Should I hold off cleaning up naming until after? (Or try to get them in before?)

ericmjl:
> Should I have PR'd into dev instead of master?

Yes, but don't worry about this particular PR ;). @szuckerman made the suggestion to move have a `dev` branch, and we've stuck with that ever since.

> When is the sprint? Should I hold off cleaning up naming until after? (Or try to get them in before?)

We can do it concurrently, don't worry! That's because everyone will be working on a different portion of the codebase. I will be marking issues as "taken", and you can take on this one.

HectorM14:
Got it. I'll update and later merge into `dev`.
Thanks for the pointers. I'll keep working on naming.

PR #168, opened 2019-05-06
Fix naming conventions in expand_column()

### LINKED ISSUE #59: [ENH] Naming conventions
I wanted to discuss naming conventions for the various functions and arguments for consistency. `expand_column` has a parameter `column` whereas `add_column` had `col_name`.

Also, is `_column` necessary in each function? Would it be ok to just have an `add()` or `transform()` method? In general I'm more on the side of more verbose function names, but just wanted to throw the idea out there.

Similarly, when following the format of `limit_column_characters` functions like `change_type` should probably be named `change_column_type`.

I'm not married to any of this (except the function arguments, those should be consistent) but wanted to get peoples' ideas.

HectorM14:
This PR resolves #59. 

# PR Description

- Updates argument `column` to `column_name`
- Updates variable `expanded` to `expanded_df`
- Accommodates backwards compatibility by accepting old argument `column`

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -631,7 +631,7 @@ def fill_empty(df, columns, value):
 
 
 @pf.register_dataframe_method
-def expand_column(df, column, sep, concat=True):
+def expand_column(df, sep, column_name=None, concat=True, **kwargs):
     """
     Expand a categorical column with multiple labels into dummy-coded columns.
 
@@ -641,7 +641,8 @@ def expand_column(df, column, sep, concat=True):
 
     .. code-block:: python
 
-        df = expand_column(df, column='col_name',
+        df = expand_column(df,
+                           column_name='col_name',
                            sep=', ')  # note space in sep
 
     Method chaining example:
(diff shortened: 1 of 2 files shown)


ericmjl:
@HectorM14 at this point, it's failing only on code style.

HectorM14:
Sure.  Is it appropriate to break the usage in the docstring  into multiple lines as `black` would do if it were code?:

```python
"""
    .. code-block:: python

        import pandas as pd
        import janitor
        df = pd.DataFrame(...).expand_column(df, column_name='col_name', sep=', ')
"""
```

```python
"""
    .. code-block:: python

        import pandas as pd
        import janitor
        df = pd.DataFrame(...).expand_column(
            df, column_name='col_name', sep=', '
        )
"""
```

Otherwise, this may be a reason to use `col_name` instead of `column_name`

ericmjl:
> Is it appropriate to break the usage in the docstring into multiple lines as black would do if it were code?

Yes, it definitely is! We use `black` internally, and I think it's a very good and opinionated code formatted that @ambv created. Keeping the code style consistent across examples is a good idea as well, because it can be educational for new users.

HectorM14:
`black` is a great tool. `blackcellmagic` for Jupyter notebooks is a nice way to keep them clean without an IDE. I'll follow the multi-line example usage found in the rest of the project.

ericmjl:
@HectorM14 this looks like it's ready to merge, I think. Could you confirm?

HectorM14:
Ready to merge! I'll look more closely for future PR's so it's smoother.

ericmjl:
Ok! Thanks a lot, @HectorM14!

PR #205, opened 2019-05-06
Add ability to groupby and agg, but chain it within the same dataframe

### LINKED ISSUE #204: GroupBy, aggregate, but keep within the original DataFrame
I often want aggregates over different values in a dataframe, but still want to see those values alongside non-aggregate values, which can require multiple non-chainable steps of code.

Wonder if anyone has an example of this, or has an opinion?

jonnybazookatone:
This PR resolves #204

# PR Description

- An example of how you could think about addressing this problem.
- Added a simple test
- Note: I haven't replicated all the kwargs across `GroupBy` and `agg` because it may not be the way in you which you want things to work.

# Relevant Reviewers
- @ericmjl

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2645,13 +2645,7 @@ def groupby_agg(
 
     Allow one to chain a groupby and a merge
 
-    Functional usage example:
-
-    .. code-block:: python
-
-        df = groupby_agg(df, by='col1', agg='mean', new_column='col1_mean')
-
-    Method chaining example:
+    Example:
 
     .. code-block:: python
 
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2645,7 +2645,13 @@ def groupby_agg(
 
     Allow one to chain a groupby and a merge
 
-    Example:
+    Without this function, we would have to break out of method chaining:
+
+    .. code-block:: python
+        df_grp = df.groupby(...).agg(...)
+        df = df.merge(df_grp, ...)
+
+    Now, this function can be method-chained:
 
     .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


janitor/functions.py
```diff
@@ -2630,3 +2630,51 @@ def to_datetime(df: pd.DataFrame, column: str, **kwargs) -> pd.DataFrame:
...
+        new_column: str,
+        agg_column: str,
+        agg: Union[Callable, str, List, Dict],
+        axis: int=0
+) -> pd.DataFrame:
+    """
+
+    Allow one to chain a groupby and a merge
+
+    Functional usage example:
```
```suggestion
    Without this function, we would have to break out of method chaining:

    .. code-block:: python

        df_grp = df.groupby(...).agg(...)
        df = df.merge(df_grp, ...)

    Now, this function can be method-chained:
```

janitor/functions.py
```diff
@@ -2630,3 +2630,51 @@ def to_datetime(df: pd.DataFrame, column: str, **kwargs) -> pd.DataFrame:
...
+
+    .. code-block:: python
```
ericmjl: I'm trying to nudge people away from functional usage of pyjanitor and towards the method-chainable API, so perhaps we can kickstart the removal of the functional API examples (even though it's technically still possible) starting from here?

janitor/functions.py
jonnybazookatone: I can add this, or just remove completely if you want people to move away?

ericmjl: This part was added so as to contrast how without your contrib, we'd have to break out of method chaining. I would love for this part to be kept, so that it's evident for a new person that this function makes life much easier!

jonnybazookatone: Added the text you suggested

ericmjl:
Thanks @jonnybazookatone!

PR #363, opened 2019-05-16
[ENH] Add take_first function

dsaxton:
This PR is to add the function `take_first()` as an expression for `sort_values().drop_duplicates()`.


ericmjl:
Thank you, @dsaxton!

PR #365, opened 2019-05-17
[ENH] Add apply_over function

dsaxton:
This makes it easier to implement window functions in `pandas` when we want to do grouped aggregation without reducing the rows of our `DataFrame`.

### DIFF
--- a/tests/functions/test_apply_over.py
+++ b/tests/functions/test_apply_over.py
@@ -1,0 +1,13 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+
+@pytest.mark.functions
+def test_apply_over():
+    df = pd.DataFrame({"a": ["dog", "dog", "cat", "cat"], "b": [1, 2, 3, 4]})
+
+    res = df.apply_over(func=np.max, col="b", by="a", name="max")
+    exp = pd.concat((df, pd.Series([2, 2, 4, 4], name="max")), axis=1)
+
+    assert res == exp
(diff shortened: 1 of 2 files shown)


ericmjl:
@dsaxton I think I was a bit quick to hit the merge button on #363. Turns out Pipelines didn't error out when the tests failed for take_first and apply_over. 

I will provide a code review shortly.

tests/functions/test_apply_over.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@pytest.mark.functions
+def test_apply_over():
+    df = pd.DataFrame({"a": ["dog", "dog", "cat", "cat"], "b": [1, 2, 3, 4]})
+
+    res = df.apply_over(func=np.max, col="b", by="a", name="max")
+    exp = pd.concat((df, pd.Series([2, 2, 4, 4], name="max")), axis=1)
+
+    assert res == exp
```
ericmjl: This test should use `pd.testing.assert_frame_equal`, which is the proper way to conduct the test. Could you change this for the test for `take_first`?

ericmjl: Actually, don't worry about `take_first`; I'll handle that one.

dsaxton:
> I might not be understanding this function correctly; could you perhaps elaborate on how this differs from `groupby_agg`, which was contributed a while ago?
> 
> Also, can you add the function to an appropriate section under `docs/reference/general_functions.rst`?

Ah, hadn't seen `groupby_agg`.  You're right, this function is basically doing the same thing so I'll go ahead and close this.

ericmjl:
Thanks for thinking of the idea, and making the PR, nonetheless!

PR #379, opened 2019-05-27
[INF] Test only on latest Python version

ericmjl:
# PR Description

- I have removed py36 from the build matrix. This should reduce the number of slots used per PR for CI builds, and hence speed up the time taken when there are multiple builds taking place.


ericmjl:
Given that this passes on Linux py37, I am going to merge.

PR #463, opened 2019-07-14
[BUG] reset_index in intro notebook

zbarry:
This was a bug introduced because `reset_index_inplace()` was deprecated and removed, I'm guessing.

@ericmjl


ericmjl:
Closed because the fix was made in Zach's food coma mode.

PR #472, opened 2019-07-15
[DOC] Standardized Chemistry Function DocStrings 

loganthomas:
# PR Description

Please describe the changes proposed in the pull request: 

- Ensured all functions within `chemistry.py` had the following:
  - Functional usage examples 
  - Method chaining usage examples
  - Parameters
  - Returns

- Made sure previous `biology.py` functions followed same conventions as those above (within `chemistry.py` file) 

**This PR resolves #462**

# PR Checklist

Please ensure that you have done the following:

1. [X] PR in from a fork off your branch. Do not PR from `<your_username>`:master, but rather from `<your_username>`:<branch_name>.

2. [X] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

## Quick Check

To do a very quick check that everything is correct, follow these steps below:

- [X] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build
    
Once done, please check off the check-box above.
    
If `make check` does not work for you, you can execute the commands listed in the Makefile individually.

## Code Changes

If you are adding code changes, please ensure the following:

- [X] Ensure that you have added tests.
- [X] Run all tests (`$ pytest .`) locally on your machine.
    - [X] Check to ensure that test coverage covers the lines of code that you have added.
    - [X] Ensure that all tests pass.

## Documentation Changes

If you are adding documentation changes, please ensure the following:

- [X] Build the docs locally.
- [X] View the docs to check that it renders correctly.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #485, opened 2019-07-20
[ENH] Add inflation adjustment function #384

### LINKED ISSUE #384: [ENH] Inflation Adjustment
# Brief Description

Adding inflation adjustment would be a nice addition to the `finance` tools in `pyjanitor`. The package already has an external API it uses for currency conversion, so adding another external API wouldn't be totally unreasonable.

[StatBureau](https://www.statbureau.org/en/inflation-api) provides an Inflation API, but it requires non-commercial use. There is probably something out there we can use.

Is there interest in adding this to `pyjanitor`, and is anyone eager to try implementing it?

# Example API

```python
# inflation adjust currency given currency name and years
df.adjust_inflation(column_name=['USD_2018'],
                    currency='USD',
                    reference_year=1967,
                    current_year=2018,
                    new_column_name=['USD_1967'])

```

rahosbach:
# PR Description

- Added a new method `inflate_currency` (along with a couple helper methods) to finance.py() that inflates a column of currency values from `currency_year` to `to_year` based on the specified country of the currency.  This method relies on the World Bank's Indicator API, which is based on country  name (not currency type).
- Added a new test module called test_inflate_currency.py that provides test coverage for almost all of the added lines for the inflate_currency() method and its helper functions.
- Ran all linting, styling, and unit test tests; everything passes.

**This PR resolves #384.**

# PR Checklist

(I'm already added to the contributor's list)

## Quick Check

    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build
    
## Code Changes

# Relevant Reviewers

- @ericmjl
- @hectormz


rahosbach:
I made a couple of changes to hopefully increase the test coverage.  I think all added lines for this PR are now covered, except for the if construct that checks that the API is actually sending back a valid response (i.e., code 200).  I've yet to have an instance where an invalid response code was received.

janitor/finance.py
```diff
@@ -204,3 +487,291 @@ def convert_currency(
...
+        'Kosovo': 'XKX', 'Yemen, Rep.': 'YEM', 'South Africa': 'ZAF',
+        'Zambia': 'ZMB', 'Zimbabwe': 'ZWE'}
+    :param currency_year: The currency year to inflate from.
+        The year should be 1960 or later.
+    :param to_year: The currency year to inflate to.
+        The year should be 1960 or later.
+    :param make_new_column: Generates new column for inflated currency if
+        True, otherwise, inflates currency in place.
+
+    :Setup:
```
ericmjl: The structure "setup-example-output" is, at the moment, done in Jupyter notebooks. In the API docs, in which the docstrings are rendered, we only show an example of how to call the function, without showing the necessary inputs and outputs. Would you be kind enough to move this example to a notebook instead, so that the docstrings can be standardized on just showing the method-chaining and functional examples?

tests/finance/test_inflate_currency.py
```diff
@@ -0,0 +1,137 @@
+from datetime import date, datetime
+
+import pytest
+import requests
+
+from janitor.finance import _inflate_currency, inflate_currency
+
+
+@pytest.mark.finance
+def test_make_currency_inflator_api_request():
```
ericmjl: Something came to mind on how to improve tests that require internet connectivity. Maybe we can do "skip if cannot ping some generic, always-available server"?

Here's a suggestion.

```suggestion
@pytest.mark.skipif(...)  # some boolean that indicates whether we are able to ping internet or not?
def test_make_currency_inflator_api_request():
```

It's ok if you don't have an answer to this one - feel free to resolve it and we can move on without changing the test.

tests/finance/test_inflate_currency.py
```diff
@@ -0,0 +1,137 @@
...
+    inflator = _inflate_currency("USA", currency_year=2018, to_year=2015)
+    df = dataframe.inflate_currency(
+        "a",
+        country="USA",
+        currency_year=2018,
+        to_year=2015,
+        make_new_column=False,
+    )
+    updatedval = df["a"].sum()
+    assert (initialval * inflator) == pytest.approx(updatedval)
```
ericmjl: Superb! :thumbsup:

janitor/finance.py
rahosbach: @ericmjl Just to be clear, are you suggesting that:

1. I pull out the setup and output portions of the doc string entirely (leave in the example), and
2. Put all three (setup, example, and output) in a separate notebook located under /examples/notebooks.

tests/finance/test_inflate_currency.py
rahosbach: I'm not sure off-hand how to handle this.  I'm going to resolve for now, and we can always add in such functionality later if/when we figure it out.

hectormz:
This is a really nice PR with tests @rahosbach ! I'm looking through myself to see if there are any other needed/worthwhile changes beyond what @ericmjl  has pointed out.

janitor/finance.py
```diff
@@ -204,3 +487,291 @@ def convert_currency(
...
+        )
+
+    inflator = (
+        inflation_dict_ready[to_year] / inflation_dict_ready[currency_year]
+    )
+    return inflator
+
+
+@pf.register_dataframe_method
+@deprecated_alias(colname="column_name")
```
hectormz: `inflate_currency()` shouldn't need the `deprecated_alias` decorator since it is a new method. You only need it if it once accepted `colname` in the past

janitor/finance.py
ericmjl: @rahosbach yes, that’s in the right direction. Maybe a more specific thing would be:

- Remove setup and output portions.
- Follow the rest of the examples in the biology and chemistry submodules, where I know that @loganthomas has started standardizing the docstrings (see PR #472 and #461). That is where we would like to be:
    1. One line description
    1. Additional notes
    1. Method chaining usage example
    1. Functional usage example
    1. Parameter description
    1. Return description

If you would like to have an example available, my loosely held opinion is that the best place is the notebook examples. My other loosely held opinion is that docstrings should be as small as possible, but not smaller than the aforementioned list. Hence, real-life examples go in notebooks, rather than the docstrings. This makes maintenance of docstrings a bit easier, and reduces the number of lines of docstrings we need to scroll through before we hit the code. If we were to start expanding the docstrings, I would probably want to leverage another sprint to expand them all at once, so until we make that decision, for now it should be the minimal list above.

rahosbach: Thanks, @ericmjl.  I updated the doc string using the example from the biology submodule.  I did not create an example for this PR.  I figure in the future I will create another PR (with an associated issue) that contains a single example notebook that covers both of the finance submodule's methods: convert_currency() and inflate_currency().

I think this PR should be ready now.

janitor/finance.py
rahosbach: Thanks @hectormz.  I wasn't entirely sure what that decorator was for, but I saw it attached to many existing methods.  I've removed it from this new inflate_currency() method.

janitor/finance.py
ericmjl: Awesome stuff, I just reviewed the changes, and they look great to me!

ericmjl:
Awesome stuff, @rahosbach! And thanks for reviewing, @hectormz!

hectormz:
Great PR @rahosbach ! Thanks for picking up the issue 👏

PR #507, opened 2019-07-28
[ENH] Added convert_units() method in engineering submodule

### LINKED ISSUE #486: [ENH] Unit Conversion
# Brief Description

I would like to propose a set of `engineering` methods that would convert units (e.g., Fahrenheit to Celsius, or Pascals to mmHg, etc.) as well as prefixes (e.g., milligrams to kilograms, kWh to MWh, etc.).

This could go down a big rabbit hole.  But maybe we could rely on [Pint](https://pint.readthedocs.io/en/0.9/), or a similar package.

# Example API

```python
# Convert a column of temperatures in Celsius to Fahrenheit
df.convert_units(column_name='temperature_c',
                 existing_units='C',
                 to_units='F',
                 create_new_column=False)
```

rahosbach:
# PR Description

- Created engineering.py submodule.
- Added convert_units() method to engineering submodule that converts a numeric column from one unit to another using the [unyt](https://unyt.readthedocs.io/en/stable/) package.
- Added a new "engineering" folder under the tests directory containing a file for testing the convert_units() method (achieves 94% coverage).
- Updated the README.rst file to include a reference to the new engineering submodule.
- Updated pytest.ini to add a mark for engineering tests.
- Updated requirements.txt and requirements-dev.txt to include the unyt package.
- Updated CHANGELOG.rst with the improvements from this PR.
- Added engineering.rst and updated index.rst for inclusion of new submodule and method into docs.

**This PR resolves #486.**

# PR Checklist

(I'm already on the contributors list)

## Quick Check

    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build
    
## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -162,7 +162,7 @@ Current functionality includes:
   into dummy-encoded variables
 - Concatenating and deconcatenating columns, based on a delimiter
 - Syntactic sugar for filtering the dataframe based on queries on a column
-- Experimental submodules for finance, biology, and engineering
+- Experimental submodules for finance, biology, chemistry, and engineering
 
 API
 ---
--- a/README.rst
+++ b/README.rst
@@ -162,7 +162,7 @@ Current functionality includes:
   into dummy-encoded variables
 - Concatenating and deconcatenating columns, based on a delimiter
 - Syntactic sugar for filtering the dataframe based on queries on a column
-- Experimental submodules for finance and biology
+- Experimental submodules for finance, biology, and engineering
 
 API
 ---
--- a/janitor/engineering.py
+++ b/janitor/engineering.py
@@ -6,6 +6,8 @@ import numpy as np
 import pandas as pd
 import pandas_flavor as pf
 
+from janitor import check
+
 from .utils import import_message
 
 try:
--- a/janitor/engineering.py
+++ b/janitor/engineering.py
@@ -94,14 +94,34 @@ def convert_units(
         raise TypeError(f"{column_name} must be a numeric column.")
 
     # Check that existing_units and to_units are of the same type
-    existing_type = eval("unyt." + existing_units + ".dimensions")
-    to_type = eval("unyt." + to_units + ".dimensions")
-    if existing_type != to_type:
-        raise TypeError(
-            f"{existing_units} are not of the same type as {to_units}."
-            f"{existing_units} are of type {existing_type}."
-            f"{to_units} are of type {to_type}."
-        )
+    # Need to handle area and volume units differently
+    if ('**' in existing_units) & ('**' in to_units):
+        # Check for same exponent on units
+        if existing_units.split('**')[1] == to_units.split('**')[1]:
+            existing_type = eval("unyt." + existing_units.split('**')[0] + ".dimensions")
+            to_type = eval("unyt." + to_units.split('**')[0] + ".dimensions")
+
+            if existing_type != to_type:
+                raise TypeError(
+                    f"Base unit of {existing_units} are not of the same type as {to_units}."
+                    f"Base unit of {existing_units} are of type {existing_type}."
+                    f"Base unit of {to_units} are of type {to_type}."
+                )
+        else:
+            raise TypeError(
+                f"Exponent on {existing_units} is not the same type as 
+                f"the exponent on {to_units}."
+            )
+    else:
+        existing_type = eval("unyt." + existing_units + ".dimensions")
+        to_type = eval("unyt." + to_units + ".dimensions")
+
+        if existing_type != to_type:
+            raise TypeError(
+                f"{existing_units} are not of the same type as {to_units}."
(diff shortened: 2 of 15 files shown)


README.rst
```diff
@@ -162,7 +162,7 @@ Current functionality includes:
   into dummy-encoded variables
 - Concatenating and deconcatenating columns, based on a delimiter
 - Syntactic sugar for filtering the dataframe based on queries on a column
-- Experimental submodules for finance and biology
+- Experimental submodules for finance, biology, and engineering
```
ericmjl:
```suggestion
- Experimental submodules for finance, biology, chemistry, and engineering
```

janitor/engineering.py
```diff
@@ -0,0 +1,135 @@
...
+    # Check that column_name is a numeric column
+    if not np.issubdtype(df[column_name].dtype, np.number):
+        raise TypeError(f"{column_name} must be a numeric column.")
+
+    # Check that existing_units and to_units are of the same type
+    # Need to handle area and volume units differently
+    if ("**" in existing_units) & ("**" in to_units):
+        # Check for same exponent on units
+        if existing_units.split("**")[1] == to_units.split("**")[1]:
+            existing_type = eval(
```
ericmjl: I'm a bit wary of evaluating strings, even though I know this is a relatively restricted thing. 

Is there an alternative to evaluating the raw string? Could we maybe use a mapping (i.e. dictionary), e.g. like how it is done in janitor.chemistry?

(If my question doesn't make sense, please do point it out where I might be assuming something wrong!)

rahosbach: I'm not entirely sure what you are thinking of in terms of code insecurities; I'd love to hear an example if you have one (mainly because I haven't really considered this issue in my coding before).

At any rate, that whole section of code in the method was meant to just provide more concise, better feedback in the case of a user trying to convert between incompatible units.  Of course, the unyt package will throw an error if the provided units are not both valid and compatible.  Therefore, I've removed that section of my code and just allowed the unyt error message to prevail (in this case, it will throw a UnitConversionError).

As a side effect, removing those lines of code dropped the test code coverage from 94% to 90%.  The only untested part of the submodule is a try-error construct in the case that the user does not have unyt installed.  I'm not sure if there's an easy way to have Pytest check for somthing like that (maybe a decorator that assumes a most basic python environment?).

anzelpwj:
Would `ast.literal_eval` work instead of `eval`. Could resolve Eric's concern if so.

anzelpwj:
Actually, I think what you need is `getattr`

janitor/engineering.py
ericmjl: @rahosbach I think the main concern I have stems from injecting malicious code into an `eval` block. Given the code you’ve written, it’s hard for that to happen, but I think we can eliminate this attack surface by avoiding the use of `eval` in general. 

This SO comment has a great explanation why: https://stackoverflow.com/a/9384005

@anzelpwj has a nifty suggestion below to use `hasattr` and `getattr`. Perhaps we could use that instead, to safely access a module’s functions?

rahosbach: Thanks @anzelpwj and @ericmjl.  I misinterpreted what your concern was at the start, but I can see why using eval() could be problematic (and thanks for the SO link).

The good news is that, since we are now just relying on unyt's error catching to throw UnitConversion errors whenever invalid or incompatible units are provided, we can just change the syntax that we use as follows:
**old**: `eval('unyt.' + existing_units)`
**new**: `unyt.Unit(existing_units)`

That was the only instance of eval() that still remained, and it's now gone.

anzelpwj: I will say that `NamedTuple` (Python core library even!) uses `eval`, so it definitely has a time and a place, but no need to run it if it's not needed.

requirements.txt
```diff
@@ -4,4 +4,5 @@ scikit-learn>=0.19.1
 xlrd>=0.9.0
 pandas-flavor==0.1.2
 biopython
-requests
\ No newline at end of file
+requests
+unyt
```
ericmjl: @rahosbach I just caught this one - we should keep `unyt` outside of `requirements.txt`, but it's ok for `unyt` to be in `requirements-dev.txt`. 

In a later PR, I will be removing `biopython` from `requirements.txt`, because it is technically an optional dependency.

ericmjl: On the other hand, it should be added to `environment.yml`, because the Azure pipeline build uses all packages to run tests.

rahosbach: OK.  I removed unyt from `requirements.txt` and added it to `environment-dev.yml`.  I didn't see an `environment.yml` file in the repo...

ericmjl:
Okie dokes, as soon as all tests pass, we’re good to go!

ericmjl:
Awesome stuff. Thanks @rahosbach!

PR #550, opened 2019-09-02
[ENH] Add decorators to skip nans and errors in iteration

### LINKED ISSUE #549: [ENH] Decorator utility for handling NAs and errors (for transform_column)
# Brief Description

If say `df["x"]` is a string column with NAs, then the routine `df["x"].apply(lambda x : x+"1")` doesn't work due to the NAs. It would be useful to have a decorator such that 
```python
df.transform_column("x", skipna(lambda x : x + "1"))
``` 
would work

# Example API

```python
from janitor.func_utils import skipna, skiperror

df.transform_column("x", skipna(lambda x : x + "1"))
df.transform_column("y", skiperror(func))

@skiperror
def foo(x):
    ...
    pass
```

# Example implementation

```python

def skipna(f):
    def _wrapped(x):
        if type(x) is float and np.isnan(x) or x is None:
            return np.nan
        else:
            return f(x)

    return _wrapped


def skiperror(f, return_val=None, return_x=False):
    def _wrapped(x):
        try:
            return f(x)
        except:
            if return_x:
                return x
            return return_val

    return _wrapped
```

jiafengkevinchen:
# PR Description

- Implements the functions proposed in #549 
- Provides decorators `.utils.skipna` and `.utils.skiperror` for skipping np.nan and errors

**This PR resolves #549 **

1. [x ] PR in from a fork off your branch. Do not PR from `<your_username>`:master, but rather from `<your_username>`:<branch_name>.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/utils.py
+++ b/janitor/utils.py
@@ -347,7 +347,7 @@ def skiperror(
     def _wrapped(x, *args, **kwargs):
         try:
             return f(x, *args, **kwargs)
-        except:
+        except Exception:
             if return_x:
                 return x
             return return_val
--- a/janitor/utils.py
+++ b/janitor/utils.py
@@ -4,6 +4,7 @@ import functools
 import warnings
 from typing import Callable, Dict, List, Union
 
+import numpy as np
 import pandas as pd
 
 from .errors import JanitorError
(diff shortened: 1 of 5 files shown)


tests/utils/test_skipna.py
```diff
@@ -0,0 +1,23 @@
...
+from janitor.utils import skipna
+import numpy as np
+
+import pytest
+
+
+@pytest.mark.functions
+def test_skipna():
+    df = pd.DataFrame({"x": ["a", "b", "c", np.nan], "y": [1, 2, 3, np.nan]})
+    func = lambda s: s + "1"
```
ericmjl: I like the use of lambda! But to keep the codebase beginner-friendly, can we instead use an internal function definition?

tests/utils/test_skiperror.py
```diff
@@ -0,0 +1,29 @@
...
+from janitor.utils import skiperror
+import numpy as np
+
+import pytest
+
+
+@pytest.mark.functions
+def test_skiperror():
+    df = pd.DataFrame({"x": [1, 2, 3, "a"], "y": [1, 2, 3, "b"]})
+    func = lambda s: s + 1
```
ericmjl: I like the use of lambda! But to keep the codebase beginner-friendly, can we instead use an internal function definition? i.e.:

```python
def test_skiperror():
...
    def func(s):
        return s + 1
```

tests/utils/test_skiperror.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # Verify that applying function causes error
```
ericmjl: Lines 14-18: These should be caught using a `with pytest.raises` block. You can find an example [here](https://stackoverflow.com/a/56569533/1274908).

tests/utils/test_skipna.py
```diff
@@ -0,0 +1,23 @@
...
+
+    # Verify that applying function causes error
+    try:
```
ericmjl: (Copy/pasted from other block)

Lines 14-18: These should be caught using a `with pytest.raises` block. You can find an example [here](https://stackoverflow.com/a/56569533/1274908).

janitor/utils.py
```diff
@@ -289,3 +290,66 @@ def check_column(
...
+        should be returned
+    :param return_val: the value to be returned when an error hits.
+        Ignored if return_x is True
+    :returns: _wrapped, the wrapped function
+    """
+
+    def _wrapped(x, *args, **kwargs):
+        try:
+            return f(x, *args, **kwargs)
+        except:
```
ericmjl: I think here we can use a generic Exception:

```suggestion
        except Exception:
```

zjpoh:
@ericmjl @jiafengkevinchen I will do that tonight. Then we can have a clearer plan when to have the next release. Sorry for the bottleneck....

ericmjl:
No worries, @zjpoh! Btw, please don't feel pressured to work on this - `pyjanitor` has always been a volunteer project, and I'm more than happy that there's contributions from all over the place coming in :smile:. If you feel like you need a break, just let me know; I'm totally OK with releasing a new version even if some parts of it are not complete. The package is always WIP :smile:.

PR #560, opened 2019-09-06
[ENH] Change morgan_fingerprints to use chirality by default

### LINKED ISSUE #558: [ENH] Enable chiral morgans by default
# Brief Description

I would like to propose changing `janitor.chemistry.morgan_fingerprints` to pass `useChirality=True` to both `GetMorganFingerprintAsBitVect` and `GetHashedMorganFingerprint `. This should be exposed directly as default to `True` in the Python API.

Clayton-Springer:
This PR closes #558.

Changes made:

* We added `useChirality=True`.
* We do not surface this as an option to end-users because this should be on by default.


ericmjl:
Thanks @Clayton-Springer!

PR #571, opened 2019-09-16
[ENH] Added delete_columns option to coalesce function

### LINKED ISSUE #561: [ENH] Coalesce function that doesn't delete columns
# Brief Description

It seems the coalesce function must always replace all input columns with one output column. While this makes sense from the meaning of "coalesce" I can think of cases when this is not the desired outcome.

For example, I have a column D that is a good default value for a few other columns and I might want to coalesce([A,D],A), coalesce([B, D], B) etc. This would break currently as D would disappear after the first coalesce

In this case, I could use A.fillna(D, inplace=True) but if I have multiple default columns (e.g. I want to coalesce([A,B,D]) and coalesce([B,C,D]) then it becomes messy. Maybe there is a better way I am not aware of.

I'm happy to try a PR for this but wanted to confirm that this is a reasonable thing to do.

# Example API

I would propose adding a delete_columns option which is defaulted to be True so existing behaviour is preserved. Or keep_columns which defaults to False.

```python
def coalesce(
    df: pd.DataFrame,
    column_names: Union[str, Iterable[str], Any],
    new_column_name: str = None,
    delete_columns: bool = True
) -> pd.DataFrame:
...
if delete_columns:
     df = df.drop(columns=column_names)
```
It would still replace an existing column if no column name is specified or the new column already exists.

gddcunh:
# PR Description

- Updated functions.py to modify coalesce function (and docstring) to include delete_columns option
- added test case
- updated AUTHORS and CHANGELOG

**This PR resolves #561 **

# PR Checklist

## Quick Check

    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build
    
## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -83,3 +83,5 @@ Contributors
 - `@anzelpwj <https://github.com/anzelpwj>`_ | `contributions <https://github.com/qtson/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3anzelpwj>`_
 - `@Ram-N <https://github.com/Ram-N>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3ARam-N>`_
 - `@eyaltrabelsi <https://github.com/eyaltrabelsi>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?q=is%3Aclosed+mentions%3A eyaltrabelsi>`_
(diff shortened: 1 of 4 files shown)


ericmjl:
@gddcunh there appears to be a merge conflict that needs to be resolved. 

Are you familiar with this? If not, do let me know, happy to point you to instructions.

gddcunh:
Yup, I can merge the conflicts and resubmit

gddcunh:
Okay I think it should be done. I had originally resolved the CHANGELOG to be in 0.18.2 but realized that it has already been released so it should now be in the next release section.

Issue #605, opened 2019-10-31
[INF] Time to cut a new release


ericmjl:
Friends, it's time to release 0.18.3 now :smile:. We have accumulated enough changes that I think it warrants an update.

ericmjl:
It's out in the wild! Anybody want to do a Twitter post? :smile: :tada:

hectormz:
I don't have the same following as @ericmjl , but I can tweet about the new release! Which features do we want to highlight?

hectormz:
Can we put in a quick commit to update the CHANGELOG from On Deck to 0.18.3?

ericmjl:
Please go for it!

PR #606, opened 2019-10-31
Release #605

ericmjl:
PR to bump version.

Snuck in one other change: I realized that  `isort` will always conflict with `black` under the current defaults. Hence, I removed `isort`, and that means we will only use `black` pre-commit checks.


zbarry:
Cool. Would it be possible for me to finish the xarray stuff before cutting a release?

ericmjl:
I was thinking of making that 0.19.0! You ok with that, @zbarry?

zbarry:
Yup

PR #620, opened 2019-12-25
[ENH] Add ECDF function

### LINKED ISSUE #583: [ENH] Empirical CDF calculation
Because @ericmjl & I keep copypasta'ing it across all my notebooks

```python
def calc_single_ecdf(data):
    """Compute the Empirical Cumulative Distribution Function for an Array."""
    # Number of points n
    n = len(data)

    # x data for the CDF 
    x = np.sort(data)

    # y data for the CDF
    y = np.arange(1, n + 1 ) / n

    return x, y
```

ericmjl:
Closes #583.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,7 @@ new version (on deck)
 =====================
 - [DOC] Edited transform_column dest_column_name kwarg description to be clearer on defaults by @evan-anderson.
 - [ENH] Replace ``apply()`` in favor of ``pandas`` functions in several functions. @hectormz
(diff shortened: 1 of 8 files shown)


ericmjl:
@hectormz I don’t think I put in a nan/inf check in there. It’s a good point though. Let me add in a few tests to see what happens.

ericmjl:
OK, I now remembered why the nan/inf checks were important. The series/array length calculation is not correct if we allow nan/inf values inside the array. I added an explicit error-raising step if null values exist.

PR #593, opened 2019-10-16
[DOC] Move #noqa comment to end of docstrings to avoid docs pollution

hectormz:
# PR Description

Please describe the changes proposed in the pull request: 

- Move #noqa: E501 comments from end of long urls in docstring to end of docstring
- Will prevent #noqa comments appearing in online docs
- hat tip to @bdice for solution!

**This PR resolves #555 .**

# PR Checklist

Please ensure that you have done the following:

1. [X] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [X] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

3. [X] Add a line to `CHANGELOG.rst` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

## Quick Check

To do a very quick check that everything is correct, follow these steps below:

- [X] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build
    
Once done, please check off the check-box above.
    
If `make check` does not work for you, you can execute the commands listed in the Makefile individually.

## Documentation Changes

If you are adding documentation changes, please ensure the following:

- [X] Build the docs locally.
- [X] View the docs to check that it renders correctly.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #621, opened 2019-12-26
[ENH] Implement find_replace API change

### LINKED ISSUE #255: [ENH] find_replace on multiple columns
modify find_replace to be applicable to a list of columns

ericmjl:
This PR now lets us `find_replace` using the following pattern:

```python
df = pd.DataFrame(...)

df = df.find_replace(column1={"hello": "world"}, column2={"what's up": "world"})
```

I have left out `find_replace_all`, because `df.replace()` natively fulfills this functionality.

This PR closes #255.

h/t @nickdelgrosso for the API change suggestion!

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -3039,7 +3039,20 @@ def find_replace(df, match: str = "exact", **mappings):
     pandas' ``df.replace()`` function provides the appropriate functionality.
     You can find more detail on the replace_ docs.
 
+    This function only works with column names that have no spaces
+    or punctuation in them.
+    For example, a column name ``item_name`` would work with ``find_replace``,
+    because it is a contiguous string that can be parsed correctly,
+    but ``item name`` would not be parsed correctly by the Python interpreter.
+
+    If you have column names that might not be compatible,
+    we recommend calling on ``clean_names()`` as the first method call.
+    If, for whatever reason, that is not possible,
+    then ``_find_replace()`` is available as a function
+    that you can do a pandas pipe_ call on.
+
     .. _replace: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.replace.html
+    .. _pipe: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.pipe.html
 
     :param df: A pandas DataFrame.
     :param match: Whether or not to perform an exact match or not.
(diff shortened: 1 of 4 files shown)


ericmjl:
Requested reviews from those who recently edited the files touched (based on GitHub's suggestions).

janitor/functions.py
```diff
@@ -3003,27 +2997,64 @@ def find_replace(
...
+        df = df.find_replace(
+            match='regex',
+            order={'coffee$': 'latte'},
+        )
+
+    To perform a find and replace on the entire dataframe,
+    pandas' ``df.replace()`` function provides the appropriate functionality.
+    You can find more detail on the replace_ docs.
+
+    .. _replace: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.replace.html  # noqa: E501
```
hectormz: Do you want to move `# noqa: E501` to the end of the docstring (line 3040) like #593

tests/functions/test_find_replace.py
```diff
@@ -50,6 +50,4 @@ def test_find_replace_regex(df_orders):
 @pytest.mark.functions
 def test_find_replace_regex_match_rases_error(df_orders):
```
hectormz: Can we change `rases` to `raises` in test name?

janitor/functions.py
ericmjl: Yes, good catch. I have been sloppy about this. This is even I have to have my code reviewed :smile: (I probably sound like a broken tape recorder by now.) Let me fix it.

tests/functions/test_find_replace.py
ericmjl: Good catch.

hectormz:
@ericmjl If I'm looking at this correctly, this will require column names that are provided to be capable of being variable names/keyword arguments? So a column `"num_dice"` or `"numdice"` can be provided, but not `"num dice"`? If someone has incompatible column names, they should use the underlying `_find_replace()`?

ericmjl:
> @ericmjl If I'm looking at this correctly, this will require column names that are provided to be capable of being variable names/keyword arguments? 

Yes.

> So a column `"num_dice"` or `"numdice"` can be provided, but not `"num dice"`? 

I think that is right. I will have to check what happens, and should probably add a test.

> If someone has incompatible column names, they should use the underlying `_find_replace()`?

Probably yes, and this behaviour should be documented.

------

@hectormz thanks for catching these issues! It's very helpful. Can you request changes on this, so that I don't accidentally merge?

hectormz:
@ericmjl done! I'm guessing Python just won't parse the keyword, so we probably just need to add documentation.

ericmjl:
@hectormz thanks for the good suggestion. Pushed the change up, and I'm going to merge this in.

PR #624, opened 2020-01-15
[ENH] Issue 622: Jitter Method

### LINKED ISSUE #622: [ENH] Proposing a "jitter" function
# Brief Description

Had to use this at work recently, where I had to jitter the data that we had to slightly anonymize it.

All that we did here was to add Gaussian noise scaled by a fraction of the magnitude of each  value.

# Example API

This could operate at two levels: at the pandas Series and DataFrame levels. I implemented the "Series" version at work, but not the "DataFrame" one.

As a Series method, we would do something like this:

```python
from typing import Optional, Tuple

def jitter(s: pd.Series, scale_magnitude: np.number, clip: Optional[Tuple[float, float]] = None):
    """
    Jitter a pandas Series by applying Gaussian noise.
    
    By default, the jitter function takes the numeric value of each series element as the mean,
    and uses a multiplier of the numeric value (scale_magnitude) as the standard deviation.
    """
    loc = 0
    scale = s * scale_magnitude
    noise = np.random.normal(loc=0, scale=scale)
    result = s + noise
    if clip:
        result = np.clip(result, *clip)
        
    return result
```

I think it might be possible to do the analogous thing for DataFrames.

The API would look like the following:

```python
# series  API
s = df['column_name']
s.jitter()

# dataframe API
df = pd.DataFrame(...).jitter("column_name", scale_magnitude=0.1, clip=(0, None))
```

What do others think?

rahosbach:
# PR Description

- Added jitter() method to functions.py.
- Added related tests in test_jitter.py.
- Updated docs in general_functions.rst.
- Updated CHANGELOG.rst.
- Linting and Black formatting updated some other xarray modules as well (not sure why that is, but they are included here).

**This PR resolves #622.**

# PR Checklist

## Quick Check

    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build

**Everything ran, and tests passed for what I added; but, some xarray tests failed that I did not touch.**

## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers

- @ericmjl
- @anzelpwj

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -3756,15 +3756,16 @@ def jitter(
             random_state=None,
         )
 
-    A new column will be created containing the values of the original column 
-    with Gaussian noise added.  
+    A new column will be created containing the values of the original column
+    with Gaussian noise added.
     For each value in the column, a Gaussian distribution is created
     having a location (mean) equal to the value
-    and a scale (standard deviation) equal to `scale`. 
-    A random value is then sampled from this distribution, 
-    which is the jittered value.  
-    If a tuple is supplied for `clip`, 
-    then any values of the new column less than `clip[0]` will be set to `clip[0]`,
+    and a scale (standard deviation) equal to `scale`.
+    A random value is then sampled from this distribution,
+    which is the jittered value.
+    If a tuple is supplied for `clip`,
+    then any values of the new column less than `clip[0]`
+    will be set to `clip[0]`,
     and any values greater than `clip[1]` will be set to `clip[1]`.
     Additionally, if a numeric value is supplied for `random_state`,
     this value will be used to set the random seed used for sampling.
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -18,7 +18,6 @@ from typing import (
     Union,
 )
 
-
 import numpy as np
 import pandas as pd
 import pandas_flavor as pf
@@ -3715,8 +3714,8 @@ def jitter(
     column_name: Hashable,
     dest_column_name: str,
     scale: np.number,
-    clip: Optional[Iterable[np.number, np.number]]=None,
-    random_state: np.number=None,
+    clip: Optional[Iterable[np.number, np.number]] = None,
+    random_state: np.number = None,
 ) -> pd.DataFrame:
     """
     Adds Gaussian noise (jitter) to the values of a column.
@@ -3789,7 +3788,7 @@ def jitter(
     """
 
     # Check types
-    check('scale', scale, [int, float])
+    check("scale", scale, [int, float])
 
     if scale <= 0:
         raise ValueError("`scale` must be a numeric value greater than 0.")
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -3707,3 +3707,100 @@ def toset(series: pd.Series) -> Set:
     """
 
     return set(series.tolist())
+
+
+@pf.register_dataframe_method
+def jitter(
+    df: pd.DataFrame,
+    column_name: Hashable,
+    dest_column_name: str,
+    scale: np.number,
+    clip: Optional[Iterable[np.number, np.number]]=None,
+    random_state: np.number=None,
+) -> pd.DataFrame:
+    """
+    Adds Gaussian noise (jitter) to the values of a column.
+
+    Functional usage example:
+
+    .. code-block:: python
+
+        import pandas as pd
+        import janitor as jn
+
+        df = pd.DataFrame(...)
+
+        df = jn.functions.jitter(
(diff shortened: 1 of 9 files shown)


janitor/functions.py
```diff
@@ -3707,3 +3707,102 @@ def toset(series: pd.Series) -> Set:
...
+
+        df = df.jitter(
+            column_name='values',
+            dest_column_name='values_jitter',
+            scale=1.0,
+            clip=None,
+            random_state=None,
+        )
+
+    A new column will be created containing the values
```
ericmjl: To keep this section easily editable, can we switch over to [semantic line breaks](https://sembr.org/)?

To make things easy for you, here's my proposed change:

```
    A new column will be created containing the values of the original column 
    with Gaussian noise added.  
    For each value in the column, a Gaussian distribution is created
    having a location (mean) equal to the value
    and a scale (standard deviation) equal to `scale`. 
    A random value is then sampled from this distribution, 
    which is the jittered value.  
    If a tuple is supplied for `clip`, 
    then any values of the new column less than `clip[0]` will be set to `clip[0]`,
    and any values greater than `clip[1]` will be set to `clip[1]`.
    Additionally, if a numeric value is supplied for `random_state`,
    this value will be used to set the random seed used for sampling.
```

janitor/functions.py
```diff
@@ -3707,3 +3707,102 @@ def toset(series: pd.Series) -> Set:
...
+        the jittered values to, default to None.
+    :param random_state: A interger or 1-d array value used to set the random
+        seed, default to None.
+
+    :returns: A pandas DataFrame with a new column containing Gaussian-
+        jittered values from another column.
+    """
+
+    # Check types
+    check("scale", scale, [int, float])
```
ericmjl: 👍

janitor/functions.py
```diff
@@ -3707,3 +3707,102 @@ def toset(series: pd.Series) -> Set:
...
+
+    if scale <= 0:
+        raise ValueError("`scale` must be a numeric value greater than 0.")
+    values = df[column_name]
+    if random_state:
+        np.random.seed(random_state)
+    result = np.random.normal(loc=values, scale=scale)
+    if clip:
+        if len(clip) != 2:
```
ericmjl: Can we add a quick comment line here?

```suggestion
        # Validate `clip` parameter.
        if len(clip) != 2:
```

ericmjl:
The automatic checks that are currently failing are only pycodestyle. Can you run a quick "make format" on the codebase?

janitor/functions.py
rahosbach: Sure thing!  And thanks for adding that hyperlink; I hadn't heard of semantic line breaks before.

janitor/functions.py
rahosbach: Will do.

ericmjl:
Looks like there are a few more minor problems with code style. 

If you click on "Details", and then "View more details on Azure Pipelines" (small grey lettered text), you'll be able to see where the issues are.

Here's one of them for your convenience

https://dev.azure.com/ericmjl/pyjanitor/_build/results?buildId=1342&view=logs&jobId=acc1347f-8235-55b0-95c7-0e2189e61659&j=acc1347f-8235-55b0-95c7-0e2189e61659&t=262b017d-5922-5965-7c5c-6f9a55a67380

ericmjl:
@anzelpwj, before I hit the big green button tomorrow evening, do you have any other comments?

anzelpwj:
The three things I would like to see addressed:
- Change randomstate type to Union(int, None)
- Change `if random_state` to `if random_state is not None`
- Add a unit test so that we ensure that the jitter is centered around the values and has a standard deviation around the scale value.

rahosbach:
Good thoughts, Paul.  I've just pushed up some commits that address those changes and should still pass all checks.  I'll admit that I'm not enough of a statistician to know if there's a better way to check that the mean and standard deviation of the values we're getting are within acceptable bounds: I just did a simple error calculation and allowed a result within 5% of our expectation (you'll see what I mean in the test_jitter.py module).

rahosbach:
@ericmjl The test that is currently failing is apparently a server connection issue with the WB API for the finance.inflate_currency() method (as far as I can tell).  I just ran a simple query in my browser and the API is sending me a valid response; so, I'm not sure why this error occurred or what may be happening.

ericmjl:
@rahosbach not a problem, let me just re-build that CI to be 100% sure. (Usually the macOS and Linux jobs track one another w.r.t. pass/fail, so since macOS passes, I'm sure Linux will too.)

Issue #187, opened 2019-05-06
Improve the examples in janitor.functions.py


dave-frazzetto:
Some docs in the functions in the janitor.functions.py file present runnable code examples, others don't. It could be made consistent through all functions (unless trivial)

ericmjl:
Indeed! I'd be happy to accept individual PRs per example, with new issues are raised per example to keep track of progress.

SorenFrohlich:
I have started changing the doc-strings of some of the more advanced functions in functions.py

ericmjl:
Thanks @SorenFrohlich! Looking forward to your PRs :smile:.

samukweku:
Hi Team, great work with the pyjanitor package. I noticed that some functions do not have examples attached to them. Examples would make the functions, in my opinion, much clearer. I would love to take this up, if no one is currently working on it, as it is my first time making a contribution to open source.

ericmjl:
@samukweku we would love to see a contribution from you! Examples are usually the hardest for maintainers to write, so I’d love to see what you’ve got!

Going forward, let’s build it up one by one. We can use this issue to track progress. Which do you think would be the easiest for you to get started? Let’s take it from there!

samukweku:
Cool. I will start off with groupby_agg and create an example. By the way, for the groupby_agg function, one of the parameters listed is axis, which when used returns this error: 'TypeError: groupby_agg() got an unexpected keyword argument 'axis'. Wondering if that argument should be there in the first place.

samukweku:
Hi  @ericmjl and the rest of the team. I need your advice on an error I am getting:

Any idea what make: *** [Makefile:23: test] Killed means?
make: *** [Makefile:20: html] Error 137 ?
 How do I fix the errors? 

(pyjanitor-dev) sam@pop-os:~/github_cloned_projects/pyjanitor$ make check
Running test suite...
pytest --cov-report html
##vso[task.logissue type=warning;]The 'junit_family' default value will change to 'xunit2' in pytest 6.0.
Add 'junit_family=xunit1' to your pytest.ini file to keep the current format in future versions of pytest and silence this warning.
============================= test session starts ==============================
platform linux -- Python 3.7.6, pytest-5.3.4, py-1.8.1, pluggy-0.13.1
rootdir: /home/sam/github_cloned_projects/pyjanitor, inifile: pytest.ini
plugins: azurepipelines-0.8.0, hypothesis-5.2.0, cov-2.8.1
[...]

PR #626, opened 2020-01-22
[DOC]: Added an example for groupby_agg in general functions

samukweku:
# PR Description

- Added an example for groupby_agg in general functions
-
-

**This PR is for #187 **

# PR Checklist

## Quick Check

    - black formatting
    - pycodestyle checking
    - running the test suite
    - docs build

## Documentation Changes

- [ ] Build the docs locally.

# Relevant Reviewers

@ericmjl 

- @ericmjl

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -82,6 +82,7 @@ Contributors
 - `@anzelpwj <https://github.com/anzelpwj>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3Aanzelpwj>`_
 - `@Ram-N <https://github.com/Ram-N>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3ARam-N>`_
 - `@eyaltrabelsi <https://github.com/eyaltrabelsi>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3Aeyaltrabelsi>`_
+- `@samukweku <https://github.com/samukweku>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3Asamukweku>`_
 - `@gddcunh <https://github.com/gddcunh>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3Agddcunh>`_
 - `@DollofCuty <https://github.com/DollofCuty>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3ADollofCuty>`_
 - `@bdice <https://github.com/bdice>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3Abdice>`_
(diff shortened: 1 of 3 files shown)


janitor/functions.py
```diff
@@ -3236,6 +3236,49 @@ def groupby_agg(
     new_col = df.groupby(by)[agg_column_name].transform(agg)
     df_new = df.assign(**{new_column_name: new_col})
     return df_new
+    """
```
ericmjl: The contents of this docstring should go in the block right below the function. That way it will get recognized as proper/valid docstring.

janitor/functions.py
```diff
@@ -3236,6 +3236,49 @@ def groupby_agg(
...
+    :Setup:
```
ericmjl: I think this, and the `:Example:` and `:Output:` syntax are invalid Sphinx syntax. If it is indeed invalid, it was my fault for letting it through previously.

I actually like the example you provided, and I'm convinced we should include it, but it does contain a bunch of long  overhead that may hamper the readability of the doc page.

There's two ways I can think of to get around this:

1. Move the example into its own Jupyter notebook, and then link the Jupyter notebook instead (while also building the Jupyter notebooks as part of the docs).
2. Add a bit of prose under the first code block above, and then add in all of these code blocks you've added as a single code block with Python comments annotating what's happening in-line. 

As a maintainer, I would prefer the first, because it helps keep the design pattern where quick usage reminders are shown on the API docs, but details show up in a notebook (for which we can also use Binder to serve up). 

What do you say, @samukweku?

samukweku: thanks for the feedback; I'll go with the first option  - move the example to a jupyter notebook. Any examples in the docs where this has been done so I can model my approach on that?

ericmjl: @samukweku awesomeness! 

To start, you'd put the notebooks in the [notebooks directory](https://github.com/ericmjl/pyjanitor/tree/dev/examples/notebooks).

I think that's it - we already have a bunch of infrastructure that lets us automatically pick out notebooks and build them.

To write a good notebook, make sure the purpose is written clearly, and that it has sufficient prose for others to understand. That's all we ask for :smile:.

ericmjl: And btw, you can continue working on this branch to get it done, don't worry about creating a new one.

samukweku: @ericmjl 

made the requested changes - took out the example block and created a notebook for the groupby_agg example. Let me know your thoughts.

ericmjl: @samukweku thanks for getting back! Have you pushed up the changes yet? They aren't yet reflected on the pull request. (If you're working on the `samukweku:myedits_pyjanitor` branch, and push any changes up to the same branch, they will magically be reflected on the pull request :tada:.)

samukweku: @ericmjl I think I got this time around. Small wins!

ericmjl:
Wonderful work, @samukweku! There's only one thing that's might need to be fixed left, it's that the file permissions have been changed from 644 to 755. Mainly this is just hampering my review of the PR, because there's a lot of files (289 of them) that have been changed.

Are you on a *nix system (Linux/macOS) or Windows? I might be able to help you figure out how to fix the problem with a new push.

samukweku:
thanks for the feedback @ericmjl ! made the changes to changelog.rst and authors.rst. I can see a message saying 'conflicting files' though. any reason why it would show as conflicting instead of say 'review'

ericmjl:
Let me see, there probably is something going on with the authors file. (I can’t see which on while on my phone.) don’t worry, I think I should be able to take care of it. 

Thanks again for the contribution, and for your patience!

ericmjl:
Ahhh, as it turned out, it was the Changelog! It's easily fixed though. Once all the tests re-pass, I will hit merge. Thanks @samukweku!

ericmjl:
:tada: thanks @samukweku! :tada: Going to hit the big green merge button! :tada:

samukweku:
Woohoo!  Hi @ericmjl , I will create some more examples for other functions. One suggestion : Is it ok to create a link from the functions.py to the notebooks, so any user reading the docs can easily click on and be redirected to the examples.

ericmjl:
Yes, definitely! I love the idea. It should probably be the way we provide usage examples, I think.

PR #653, opened 2020-05-06
[ENH] Upgrades to transform_column

ericmjl:
# PR Description

- This PR upgrades transform_column to be able to handle series-wise transforms via a boolean transform.
- As such, we can get access to series string methods for vectorized string processing.

**This PR does not resolves an existing issue.**

# PR Checklist

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [x] Build the docs locally.


ericmjl:
Leaving this here for anybody who would like to review! I've fixed all of the problems. If I don't get a review by Thursday or Friday then I'll go ahead and merge it.

hectormz:
@ericmjl I will give it a look before Thursday! But anyone else is welcome to review before then too

tests/functions/test_transform_column.py
```diff
@@ -4,18 +4,26 @@
...
+    """
+    Test replacing data of the original column.
+
+    The function that is used for testing here
+    must be able to operate elementwise
+    and on a single pandas Series.
+    We use np.log10 as an example
+    """
+
+    df = dataframe.transform_column("a", np.log10, elementwise=elementwise)
```
zbarry, line 18: A check to make sure that when `not elementwise`, an entire column is indeed passed:

```python
if not elementwise:
    def _test_is_series(series):
        assert isinstance(series, pd.Series)
        return series
    dataframe.transform_column("a", _test_is_series, elementwise=elementwise)
    
```

janitor/functions.py
```diff
@@ -2201,11 +2201,50 @@ def transform_column(
...
+            modified_x = ... # do stuff here
+            return modified_x
+
+        df.transform_column(column_name="my_column", function=elementwise_func)
+
+    On the other hand, columnwise application of a function
+    behaves as if the function takes in a pandas Series
+    and emits back a sequence that is of identical length to the original.
+    One place where this is desirable
+    is to gain access to `pandas` native string methods,
```
hectormz, line 2233: Should this be \``pandas\`` or did you want this to be italics?

I believe there other cases in the project where \` is used instead of \`` in docstrings due to slips between .md and .rst. But it's possible the author _did_ intend to generate italic text.

hectormz:
Still looking through, but here are some general thoughts/questions:
* Setting `elementwise` default to `True` maintains compatibility with existing usage. In your experience (or view of how this method is used) do you think users are more likely to pass in functions that are elementwise, or can operate on a whole series?
    * I think I am more likely to use `numpy` functions that operate on a series, and have would opted for `elementwise=False` as default, but I don't know what others actually use. I'm probably in the minority since the point of this function is to make any transformations, especially elementwise ones, much easier to perform.
* Should `nbstripout` be applied to all notebooks going forward? That's easy enough to do in a single PR, but wanted to make that clear.
* I find `lambda` functions are something that people either use a lot and are familiar with them, or don't know they exist at all. Is it worth adding a line in `transform_column.ipynb` that clearly states that we have created a `lambda` function to perform a custom operation on every element of the series, and in this case is the epitome of an elementwise operation?

tests/functions/test_transform_column.py
ericmjl: Cool beans, thanks for the test suggestion!

janitor/functions.py
ericmjl: Ahhhh, I kinda sorta want to move over the docs to mkdocs to take advantage of Markdown docstrings, but I'm hesitant to undo so many contributors' work making the docs work with Sphinx.

ericmjl: In any case, good catch, I'll change it to double backticks.

hectormz: I'm not volunteering right now, but supposedly it could be converted [via pandoc](https://news.ycombinator.com/item?id=17717513). 

Would you be looking for a conversion, or complete re-write?

ericmjl:
Hey @hectormz! Thanks for taking the time to do the review :smile:. Really appreciate it. I’m going to respond to your questions bit by bit.

> Setting elementwise default to True maintains compatibility with existing usage. 

You are correctly pointing out that I was first and foremost concerned with maintaining compatibility for now. 

> In your experience (or view of how this method is used) do you think users are more likely to pass in functions that are elementwise, or can operate on a whole series?

I am not quite sure. Mostly what was on my mind is to expand the options to use pandas’ string-vectorized methods in a verb-y and chain-able fashion.

> Should nbstripout be applied to all notebooks going forward? That's easy enough to do in a single PR, but wanted to make that clear.

Yes, that is right, nbstripout should be applied to all notebooks going forth. Version control is intended to be for only the things that are human-curated, while the auto-generated stuff (cell outputs; derived HTML docs) should be always re-executed. At the same time, I think we have not yet enabled execution of notebooks as part of the docs build, so that will have to get up and running first.

Upon writing that out, I think we should disable nbstripout until we know that auto-execution works as expected. I also think we should move the repo over to the pyjanitor-devs group first, as that will give us all a chance to help out with the infrastructure piece.

> Is it worth adding a line in transform_column.ipynb that clearly states that we have created a lambda function to perform a custom operation on every element of the series, and in this case is the epitome of an elementwise operation?

Yes, that’s a very good point. I will go in and add that.

ericmjl:
On second thought, I think I just figured out what my next big writing/docs project should be :smile:.

hectormz:
@ericmjl 😂 what's the next topic, or I'll have to wait?

I think it makes sense to maintain compatibility. I was just thinking aloud that vectorized operations would be the norm/preferred, and only for elementwise operations would I want to provide an additional argument.

This whole thing did make me realize I don't know too much about how our docs are made. Especially from the notebooks. I didn't think we did much with Travis anymore after switching to Azure.


Is the plan to increment the `interrogate` threshold as we update the docstrings?

ericmjl:
> what's the next topic, or I'll have to wait?

You've probably seen the updated [Bayesian Analysis Recipes](https://ericmjl.github.io/bayesian-analysis-recipes) collection, or the development site that I've been using (it's for Patreon supporters only right now - thank you for joining in 😄!) to update Network Analysis Made Simple. `mkdocs-material` really enables so much, like admonitions and built-in static site search, all written with Markdown, that I think we should move over our docs to mkdocs-material.

I don't mind at all taking the time to comb through the docs, as I think the lack of a proper style guide means I should just hunker down and write it, so that the docs are much more standardized. That will make things easier for others to contribute.

The existing docs pages can remain, probably will also restructure them (pun unintended) to make things a bit easier to navigate.

> Is the plan to increment the interrogate threshold as we update the docstrings?

Yep, that is right, @hectormz! 

OK, seeing as how everything is passing, I am going to hit the merge button now.

hectormz:
👍 awesome @ericmjl !

If you're thinking about potential changes to the docs, one thing I've been thinking about is providing an alternate/additional view for users to discover functions. I think a user looking at our API pages can see all the functions `janitor` has, and what they do, but we could also have a version where someone is browsing the page, or coming from a search engine with the question, 

*"How do I remove duplicate columns in pandas?"*

and can then find the function they want, potentially see what it would take in pure `pandas`, and how we do it in `pyjanitor`.

PR #662, opened 2020-05-29
[INF] First attempt at a dev container

ericmjl:
The purpose of this PR is to provide a development container that new users can opt in to use.

I have made the following choices:

1. We build the conda environment in the container on Travis, repeatedly and automatically. This way, nobody should need to build it locally.
1. Users can then pull the Docker container and develop code from within it.

This is all intended to work with VSCode Remote (Container) extension: https://code.visualstudio.com/docs/remote/containers


hectormz:
This is pretty cool! Finally using Docker and thought I'd check this out! 🎉

PR #673, opened 2020-06-15
[ENH] updates to update_where and find_replace functions

samukweku:
# PR Description

Please describe the changes proposed in the pull request:

- Changed update_where conditions to query style mode
- Removed find_replace dependency on update_where function
-

**This PR resolves #663 .**

# PR Checklist

Please ensure that you have done the following:

1. [x] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [x] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

3. [ ] Add a line to `CHANGELOG.rst` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

## Quick Check

To do a very quick check that everything is correct, follow these steps below:

- [x] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

Once done, please check off the check-box above.

If `make check` does not work for you, you can execute the commands listed in the Makefile individually.

## Code Changes

If you are adding code changes, please ensure the following:

- [x] Ensure that you have added tests.
- [x] Run all tests (`$ pytest .`) locally on your machine.
    - [ ] Check to ensure that test coverage covers the lines of code that you have added.
    - [ ] Ensure that all tests pass.

## Documentation Changes

If you are adding documentation changes, please ensure the following:

- [ ] Build the docs locally.
- [ ] View the docs to check that it renders correctly.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #680, opened 2020-06-22
[INF] Require pandas >= 1.0.0

hectormz:
# PR Description

- Updates minimum `pandas` version from 0.24 to 1.0.0

**This PR does not directly resolve an issue, but arose from conversation with @ericmjl.**

**Our CI uses the latest version of `pandas`, and this simplifies (likely) future compatibility issues.**

As some background, this initially arose because we have some tests that import from `pandas._testing`, which doesn't exist <1.0. I think we should import from `pandas.testing` anyway, so the `pandas` version becomes less of an issue. However, if we're claiming to be `pandas>=0.24` compatible, we might want to consider adding a CI job that will test on `pandas==0.24` in addition to `pandas==latest`.

# PR Checklist

1. [X] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

3. [X] Add a line to `CHANGELOG.rst` under the latest version header (i.e. the one that is "on deck") describing the contribution.

## Quick Check

- [X] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

# Relevant Reviewers

- @ericmjl
- @zbarry 
- @szuckerman

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ new version (on deck)
   or boolean indexing via the ``loc`` method.
   Also updated ``find_replace`` function to use the ``loc`` method directly,
   instead of routing it through the ``update_where`` function. @samukweku
+- [INF]: Update ``pandas`` minimum version to 1.0.0. @hectormz
 
 v0.20.7
 =======
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -25,7 +25,7 @@ dependencies:
 - nox
 - numpy>=1.16
 - pandas-flavor
-- pandas>=0.24.0
+- pandas>=1.0.0
 - pip
 - pipreqs
 - pre-commit
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 natsort
 numpy>=1.14.1
 pandas-flavor==0.2.0
-pandas>=0.24.0
+pandas>=1.0.0
 requests
 scikit-learn>=0.19.1
 xlrd>=0.9.0


hectormz:
Not sure about the "10 fewer hit lines", but this seems to match what happened in #673

ericmjl:
> Not sure about the "10 fewer hit lines", but this seems to match what happened in #673

Yeah, not worried about this one! :smile: I approve!

ericmjl:
As a matter of habit, let's get one more eye on PRs, so that there's a culture of 2 or more reviews per PR. @anzelpwj, reviewing this PR should fit your schedule, I think, as it's of a very narrow scope. Would you like to do the honors?

szuckerman:
I’m just seeing this now, and take my comment with a grain of salt since I haven’t been contributing much to the project recently, but I’m not sure it’s the best thing to require the most recent version of a library. 

There are some users who might not be able to upgrade as often as you’d think and this might make the library inaccessible to them. 

But, narrowing in on a version does make it easier for development and if the majority of users will just “pip install” the latest version of pyjanitor anyway, this might not be a big deal. Just a few thoughts.

hectormz:
Hi @szuckerman that's a valid point. We weren't really sure where the `pyjanitor` contributors and users stood regarding `pandas` requirements. 

As some background, this initially arose because we have some tests that import from `pandas._testing`, which doesn't exist <1.0. I think we should import from `pandas.testing` anyway, so the `pandas` version becomes less of an issue. However, if we're claiming to be `pandas>=0.24` compatible, we might want to consider adding a CI job that will test on `pandas==0.24` in addition to `pandas==latest`.

I'll update the original PR comment with this info.

ericmjl:
> There are some users who might not be able to upgrade as often as you’d think and this might make the library inaccessible to them.

That is true, and I agree with you. 

> But, narrowing in on a version does make it easier for development and if the majority of users will just “pip install” the latest version of pyjanitor anyway, this might not be a big deal. Just a few thoughts.

This is also true. 

On balance, taking into account the fact that pandas 1.x has been pretty backwards-compliant with the old 0.2x APIs, I think it's okay to pin to 1.0, which is a pretty early version.

ericmjl:
I don't want this one to hang too long, so I'm going to merge.

PR #684, opened 2020-06-27
[TST] Add automatic doc checking for general functions.

### LINKED ISSUE #683: Ad-hoc script to search for functions not yet added to the General Functions API Page
This is a reference to #682 - a script I used to search for functions not yet added to the API page. 

```from pathlib import Path
import re

test_folder = Path("./tests/functions")

#extract the names from the test_functions folder
function_names = [
    name.stem.split("_", 1)[-1].strip() for name in test_folder.iterdir()
]

docs = Path("./docs/reference/general_functions.rst")

pattern = re.compile("\s{4}[a-zA-Z_]+")

#get the names in the general_functions page
with docs.open() as doc:
    doc_functions = [
        pattern.search(line).group().strip()
        for line in doc
        if pattern.search(line)
    ]

#get the functions that have not yet been added
missing_funcs = set(function_names).difference(doc_functions)

print(missing_funcs)```

### PR DESCRIPTION
**Pull request by ericmjl**
# PR Description 

- Adding a script that can be run on PRs to check for automatic docs coverage.
- h/t @samukweku for posting the script.
- Documentation will be forthcoming.

**This PR resolves #683.**

### CODE REVIEW

File: scripts/check-autodoc.py [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+import re
+from pathlib import Path
+from typing import List, Tuple
+
+
+def extract_function_names(
+    test_folder: Path, exclude_names: List[str]
+) -> List[str]:  # skipqc
+    """
+    Extract function names from the list of functions.
```
hectormz, line 21: Do you want to put this docstring on one line following PEP 257?

scripts/check-autodoc.py
```diff
@@ -0,0 +1,77 @@
...
+    """Check if path does not contain a list of names."""
+    for name in names:
+        if name in str(path):
+            return False
+    return True
+
+
+def extract_folder_names(test_dir: Path) -> Tuple[Path, str]:
+    """
+    Extract folder names.
```
hectormz, line 54: Do you want to move this first line of docstring to same line as `"""`?

hectormz:
Hi @ericmjl , so this works by scanning the `tests/functions` folder and comparing to `.rst`. Would it be more direct to get the list of functions in `janitor/functions.py`?

scripts/check-autodoc.py
ericmjl: I think I just let black do whatever it needed to.

samukweku:
I took the `tests` folder path since every function must have tests associated with it, and at the time  I wrote the ad hoc script, it was a simpler path. Do you think there are any advantages to pulling the information directly from `janitor functions` module? Just curious. By the way, It seems you have a good grasp of all the PEPs. I just saw a comment of yours where you mentioned PEP287.

scripts/check-autodoc.py
hectormz: Got it. I believe that `black` doesn't touch docstrings and doesn't have an opinion on them. (I tested with a 300 character long docstring on a single line)

In the grand scheme of things, I think this is a minor aspect of docstrings, but might be worth trying to align the codebase.

Following PEP 257, we could put one-liners on one line, and 

>Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

I think I personally notice the different styles in the codebase when scanning through functions, and could just do an issue/PR myself to align them to whichever style is decided.

hectormz:
@samukweku got it. My first inclination is to get the list of functions directly, especially if pulling from tests relies on test coverage. But if this method is easier, that's understandable. I guess I should look into how easy/bad it would be to get the function list from `functions.py` myself! 😅

environment-dev.yml
```diff
@@ -31,6 +31,7 @@ dependencies:
 - pre-commit
 - pyflakes
 - pylint
+- pyprojroot
```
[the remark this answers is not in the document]
ericmjl: I didn't, thanks for the catch!

scripts/check-autodoc.py
ericmjl: Good point. Consistency does matter. Pushing up a change.

ericmjl:
@samukweku yes, once I read your code, I realized what you've got would be a great strategy. The reason is because it's easier to parse filenames than to parse the code; after all it's not _purely_ the functions that we want to be documented that are there. 

@hectormz in case you want to try parsing the source, I think the only thing we'd have to do is find every function that is decorated with `@pf.register_dataframe_method`, and then we'd be golden. But I wasn't sure how to do that, and searching on the interwebs, [it looks complicated](https://stackoverflow.com/questions/5910703/how-to-get-all-methods-of-a-python-class-with-given-decorator) haha 😅

hectormz:
Ah, true! I started to look into it, but didn't think that we were looking for decorated functions 😬

hectormz:
This setup works for the general `pyjanitor` functions, but the same pattern can be extended to other modules too, right?

ericmjl:
> This setup works for the general pyjanitor functions, but the same pattern can be extended to other modules too, right?

In theory yes, but in practice, unnecessary I think. The others are automatically populated!

ericmjl:
With two reviews in (h/t @hectormz and @samukweku), one of you please do the honors! Actually, @samukweku, I'd like to invite you to do so, if you'd please!

samukweku:
I feel like I just hit the gong at the NYSE!

PR #704, opened 2020-07-26
[ENH] Pyjanitor fill method

### LINKED ISSUE #700: Forward/Backward Fill
# Brief Description

I would like to propose a forward/backward fill method, similar to ``fill`` in tidyr. In Pandas, forward/backward fill can be achieved, but you cannot select directions for different columns within a dataframe.

# Example API
    @pf.register_dataframe_method
    def fill(df, columns = ..., directions = sequence of directions(up, down, updown, downup):
        """
        Forward or backward fill on a column or columns of a dataframe.
        """
        df[column] = df[column].ffill() #forward fill will be the default
        return df

```python
example data : 
df = pd.DataFrame(data={'a':[1,2,3,None],'b':[4,5,None,6],'c':[None,None,7,8]})

          a	b	c
0	1.0	4.0	NaN
1	2.0	5.0	NaN
2	3.0	NaN	7.0
3	NaN	6.0	8.0

#downward fill (backward fill in Pandas) on only the `a` column:
df.fill(columns="a", directions = "down")

# downward fill on  ``a``, upward fill on  ``c`` :

# the idea is the directions will be paired with the columns
# similar to what exists when sorting multiple columns in Pandas' sort_values 
# with booleans passed to the ``ascending`` argument
df.fill(columns=["a", "c"], directions = ["down", "up"])



# currently in Pandas, ffill and bfill are applied to all the columns that have null values, 
# you cannot select which columns to be forward or backward filled.
```

Still undecided on the name, as I do not feel it clearly conveys what it should do without reading the function description.

samukweku:
# PR Description

- A  method-chainable ``fill`` method  for forward and backward fills on selected columns of a dataframe.

**This PR resolves #700 .**

3. [ ] Add a line to `CHANGELOG.rst` under the latest version header (i.e. the one that is "on deck") describing the contribution.

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [ ] Build the docs locally.

# Relevant Reviewers

- @ericmjl


janitor/functions.py
```diff
@@ -4104,3 +4104,210 @@ def process_text(
...
     return df
+
+
+@pf.register_dataframe_method
+def fill(
+    df: pd.DataFrame,
+    columns: Union[Tuple[str], List[str], str],
+    directions: Union[Tuple[str], List[str], str] = "down",
+    limit: Optional[int] = None,
+) -> pd.DataFrame:
```
ericmjl, line 4115: The `fill` function falls into the class of functions where we can trivially map column name to keyword if we want to express that the operation is applied over multiple columns, with potentially different directions on each column. This pattern is used in `find_replace`, and I think it can be done here too. (It seems clearer and clearer to me that where appropriate, this should be the pattern used.)

I think the most Python-idiom-respecting API might look something like this:

```python
def fill(
    df: pd.DataFrame,
    directions: Dict[Hashable, str],
    limit: Optional[int] = None,
) -> pd.DataFrame:
```

The use of a string that is split by commas is not something I would encourage. This is equivalent to passing in a mapping of x->y, and where mappings are used, I would advocate for dictionaries instead.

janitor/functions.py
```diff
@@ -4104,3 +4104,210 @@ def process_text(
...
+        if length_directions_arg > 1:
+            raise ValueError(
+                """
+                The number of arguments in the columns and
+                directions do not match.
+                """
+            )
+        directions = directions * length_columns_arg
+
+    fill_pairs = dict(zip(columns, directions))
```
ericmjl, line 4297: With the API change, a lot of these checks can be removed. I think the only ones that would be necessary when using a dictionary are that:

- It is not an empty dictionary
- All keys exist as columns in the dataframe,
- All values are one of the four valid options.

We'd have fewer lines of code to maintain this way.

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+        {"rank": 4, "pet_type": np.nan, "breed": "Exotic", "owner": np.nan},
+        {"rank": 5, "pet_type": np.nan, "breed": "Siamese", "owner": np.nan},
+        {
+            "rank": 6,
+            "pet_type": "Cat",
+            "breed": "American Short",
+            "owner": "adaora",
+        },
+    ]
+)
```
ericmjl, line 51: For maintenance (and cleanliness in the test), can we make this one a fixture? I think the way to do this is:

```suggestion
@pytest.fixture
def fill_dummy_data():
    df = pd.DataFrame(
        [
            {
                "rank": 1,
                "pet_type": np.nan,
                "breed": "Boston Terrier",
                "owner": "sam",
            },
            {
                "rank": 2,
                "pet_type": np.nan,
                "breed": "Retrievers (Labrador)",
                "owner": "ogor",
            },
            {
                "rank": 3,
                "pet_type": np.nan,
                "breed": "Retrievers (Golden)",
                "owner": "nathan",
            },
            {
                "rank": 4,
                "pet_type": np.nan,
                "breed": "French Bulldogs",
                "owner": np.nan,
            },
            {"rank": 5, "pet_type": np.nan, "breed": "Bulldogs", "owner": np.nan},
            {"rank": 6, "pet_type": "Dog", "breed": "Beagles", "owner": np.nan},
            {"rank": 1, "pet_type": np.nan, "breed": "Persian", "owner": np.nan},
            {
                "rank": 2,
                "pet_type": np.nan,
                "breed": "Maine Coon",
                "owner": "ragnar",
            },
            {"rank": 3, "pet_type": np.nan, "breed": "Ragdoll", "owner": np.nan},
            {"rank": 4, "pet_type": np.nan, "breed": "Exotic", "owner": np.nan},
            {"rank": 5, "pet_type": np.nan, "breed": "Siamese", "owner": np.nan},
            {
                "rank": 6,
                "pet_type": "Cat",
                "breed": "American Short",
                "owner": "adaora",
            },
        ]
    )
    return df
```

And then in each of the tests, you would pass in `fill_dummy_data` as the argument, as follows:

```python
def test_fill_column(fill_dummy_data):
    ...
```

`pytest` will automagically pass in the result of `fill_dummy_data`. Doing the test as a fixture makes the dataframe safe from global mutation (in case it might ever happen), so we don't accidentally change the structure of the dataframe in memory when using it in a test that executes after the mutation.

ericmjl: I'll be leaving "fixture please" comments (copy/pasted) so you can have a pseudo-todo list on the PR tracker.

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_fill_column():
```
ericmjl, line 54: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Fill down on a single column with default direction."""
+    expected = df.copy()
+    expected.loc[:, "pet_type"] = expected.loc[:, "pet_type"].ffill()
+    result = df.fill(columns="pet_type")
+    assert_frame_equal(result, expected)
+
+
+def test_fill_column_up():
```
ericmjl, line 62: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Fill up on a single column."""
+    expected = df.copy()
+    expected.loc[:, "pet_type"] = expected.loc[:, "pet_type"].bfill()
+    result = df.fill(columns="pet_type", directions="up")
+    assert_frame_equal(result, expected)
+
+
+def test_fill_column_updown():
```
ericmjl, line 70: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Fill upwards, then downwards on a single column."""
+    expected = df.copy()
+    expected.loc[:, "pet_type"] = expected.loc[:, "pet_type"].bfill().ffill()
+    result = df.fill(columns="pet_type", directions="updown")
+    assert_frame_equal(result, expected)
+
+
+def test_fill_column_down_up():
```
ericmjl, line 78: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Fill downwards, then upwards on a single column."""
+    expected = df.copy()
+    expected.loc[:, "pet_type"] = expected.loc[:, "pet_type"].ffill().bfill()
+    result = df.fill(columns="pet_type", directions="downup")
+    assert_frame_equal(result, expected)
+
+
+def test_fill_multiple_columns():
```
ericmjl, line 86: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Fill on multiple columns with a single direction."""
+    expected = df.copy()
+    expected.loc[:, ["pet_type", "owner"]] = expected.loc[
+        :, ["pet_type", "owner"]
+    ].ffill()
+    result = df.fill(columns="pet_type,owner")
+    assert_frame_equal(result, expected)
+
+
+def test_fill_multiple_columns_multiple_directions():
```
ericmjl, line 96: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Fill on multiple columns with different directions."""
+    expected = df.copy()
+    expected.loc[:, "pet_type"] = expected.loc[:, "pet_type"].ffill()
+    expected.loc[:, "owner"] = expected.loc[:, "owner"].bfill()
+    result = df.fill(columns=("pet_type", "owner"), directions="down,up")
+    assert_frame_equal(result, expected)
+
+
+def test_fill_uneven_lengths():
```
ericmjl, line 105: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Raise Value Error if number of directions is greater than one and
+    is unequal to number of columns."""
+    with pytest.raises(ValueError):
+        df.fill(columns=["pet_type", "owner"], directions=["up", "up", "down"])
+
+
+def test_wrong_column_instance():
```
ericmjl, line 112: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Raise Type Error if columns argument is not a List/Tuple/str."""
+    with pytest.raises(TypeError):
+        df.fill(columns={"pet_type"})
+
+
+def test_wrong_direction_instance():
```
ericmjl, line 118: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Raise Type Error if directions argument is not a List/Tuple/str."""
+    with pytest.raises(TypeError):
+        df.fill(columns=["pet_type"], directions={"pet_type": "up"})
+
+
+def test_wrong_column_name():
```
ericmjl, line 124: fixture please

tests/functions/test_fill.py
```diff
@@ -0,0 +1,133 @@
...
+    """ Raise Value Error if wrong column name is provided."""
+    with pytest.raises(ValueError):
+        df.fill(columns="PetType")
+
+
+def test_wrong_direction():
```
ericmjl, line 130: fixture please

samukweku:
@ericmjl , thanks for taking out the time to review the code. I wonder if it is better to lump the code, somewhat, into fill_empty:

`df = fill_empty(df, column_names=['col1', 'col2'], value=0, directions=None)`

if the ``value`` argument is not None, then directions must be None and vice versa. It might make it simpler and remove the need for one more function that might be confusing name wise. The drawback might be that I may not be able to pass in a dictionary for the mapping. @HectorM14 @zbarry @szuckerman , and others, I would also love your thoughts on this.

For the names suggested, ``fill_direction`` seems most promising.

hectormz:
@samukweku I'm just catching up to speed on the issue/PR.

FYI, you've been tagging me at `@HectorM14` which is my old account. Tag me at `@hectormz` if you want to get my attention 😁

samukweku:
> @samukweku I'm just catching up to speed on the issue/PR.
> 
> FYI, you've been tagging me at `@HectorM14` which is my old account. Tag me at `@hectormz` if you want to get my attention

Haha!, my bad. Now I know how the right way to get your attention

samukweku:
Hi @ericmjl what are your thoughts on my suggestion?

ericmjl:
The PR's been open for a few days, and we've not had much in the way of replies.

@samukweku, I propose we go forward with `fill_direction` as the function name. (By convention, most function names are two words, after all.) This is probably more informative than lumping it into `fill_empty`. In any case, I am not sure that we're near a 1.x release, so if this turns out to be an API wart after a while, we can always change it later. (Famous last words - I know that people end up just adapting, ha ha ha.)

samukweku:
the failing test is from ``test_filter_date.py``, specifically the ``test_filter_date_column_name``. I think it should raise a keyerror, since the column ``42`` does not exist in the dataframe. Not sure how to fix it

ericmjl:
@samukweku I think I see what’s going on! Could you change the expected error raised from `TypeError` to `KeyError`? That probably will fix it. (A local check before pushing can confirm for you whether this was right or not.)

samukweku:
I did change it to `KeyError` but if failed on the local test, whereas `TypeError` doesn't trigger an error

ericmjl:
@samukweku hmm, I'm quite baffled myself. Can you decorate it with `@pytest.mark.xfail`, and put a reason on the xfail decorator as "Unknown error; KeyError fails locally but ValueError fails on CI"? 

https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail

hectormz:
@ericmjl @samukweku I'm intrigued! I'll look at this quickly myself

hectormz:
@samukweku @ericmjl I figured it out 🤔

So Pandas 1.1.0 came out in the past few days, and the CI is grabbing the latest version. You likely have something like 1.0.5 locally.

According to the release notes:
> KeyErrors raised by loc specify missing labels¶
> Previously, if labels were missing for a .loc call, a KeyError was raised stating that this was no longer supported.
> Now the error message also includes a list of the missing labels (max 10 items, display width 80 characters). See GH34272.

So, we should either be flexible here, or set a maximum or minimum pandas version for now (1.1.0)

samukweku:
@hectormz Thanks for this. I thought the CI worked with the version in the environment. Good learning point for me.

hectormz:
Are you still using the Docker container? The CI grabs the latest version it can, while following the min/max version rules we set. These things pop up for me when I install once locally, and don't keep the packages up to date

samukweku:
@hectormz Yes, I'm using the docker container.

ericmjl:
@hectormz and @samukweku I think I have an idea. I'm also using the Docker container. There's probably a cached layer that's causing this error to not be revealed. I'm going to clear out my cache:

```
docker system prune -a -f
```

and see what happens.

ericmjl:
OK! Looping back and confirming that pandas 1.1.0 was the reason why we have had issues.

I suspect pandas 1.0.5 was what got resolved by conda, but in the CI system, 1.1.0 got installed because we pip install the package.

I’m going to suggest we mark the test as xfail with 1.1.0 as the reason why it’s failing. It’s not crucial at the moment, I think, for that test to be working, and with the new `.loc` changes in pandas, I think it means we can have some other package-wide changes underneath the hood (i.e. by not needing to do checks to make sure a column is present.)

ericmjl:
@samukweku it's past 24 hours! @hectormz, @zbarry or @szuckerman would you like to do the honors of hitting the big green button?

hectormz:
@ericmjl I can do it, but @samukweku can you add a one line summary docstring? (Unless my phone is not showing it properly)

samukweku:
> @ericmjl I can do it, but @samukweku can you add a one line summary docstring? (Unless my phone is not showing it properly)

Hi @hectormz , not sure what you mean. Where should the one line summary docstring be located?

janitor/functions.py
```diff
@@ -4104,3 +4104,143 @@ def process_text(
...
+def fill_direction(
+    df: pd.DataFrame,
+    directions: Dict[Hashable, str],
+    limit: Optional[int] = None,
+) -> pd.DataFrame:
+    """
+    Provides a method-chainable function for filling missing values in selected
+    columns, using the next or previous entry. The columns are paired with the
+    directions in a dictionary. It is a wrapper for ``pd.Series.ffill`` and
+    ``pd.Series.bfill``.
```
hectormz, line 4119: @samukweku This is what I would be looking for:

```suggestion
    """Provide a method-chainable function for filling missing values in selected columns.
    
    Missing values are filled using the next or previous entry.
    The columns are paired with the directions in a dictionary.
    It is a wrapper for ``pd.Series.ffill`` and ``pd.Series.bfill``.
```

janitor/functions.py
```diff
@@ -4104,3 +4104,143 @@ def process_text(
...
+    # as Pandas loc will raise a KeyError if columns provided do not exist
+    wrong_columns_provided = set(directions).difference(df.columns)
+    if any(wrong_columns_provided):
+        if len(wrong_columns_provided) > 1:
+            outcome = ", ".join(f"'{word}'" for word in wrong_columns_provided)
+            raise ValueError(
+                f"Columns {outcome} do not exist in the dataframe"
+            )
+        outcome = "".join(wrong_columns_provided)
+        raise ValueError(f"Column {outcome} does not exist in the dataframe")
```
hectormz, line 4216:
```suggestion
        raise ValueError(f"Column {outcome} does not exist in the dataframe.")
```

janitor/functions.py
samukweku: @hectormz gotcha.

hectormz:
@samukweku I realize I still haven't resolved the issue/PR on docstrings, so it should be clearer after that!

ericmjl:
Anytime you’d like to, @hectormz, please do the honors!

PR #706, opened 2020-07-27
[ENH] first pass on fill_missing_timestamps

### LINKED ISSUE #705: Add the ability to flag and fill if timestamps are missing in a pandas data frame
# Brief Description

Following up on https://github.com/ericmjl/pyjanitor/issues/703, this issue seeks to introduce the ability to flag if timestamps are missing in a given data frame, provided the frequency of the time series data in the data frame  

I would like to propose...

# Example API
```python
import pandas as pd 
import janitor 

def _flag_missing_timestamps(df, frequency, first_time_stamp=None, last_time_stamp=None):
    """
    Utility function to test that if input time_series is missing any timestamps relative to expected timestamps
    generated based on the first_time_stamp, last_time_stamp and frequency.

    :param pd.DataFrame df: data frame which needs to be tested for missing timestamps
    :param int frequency: frequency i.e. sampling frequency of the data, expressed in seconds. For e.g. if the data has
    hourly frequency, then a value of 60 * 60 = 3600 is expected
    :param pd.Timestamp first_time_stamp: timestamp at which the time_series is expected to start from. Defaults to None
    If no input is provided, assumes the minimum value in time_series
    :param pd.Timestamp last_time_stamp: timestamp at which the time_series is expected to end with. Defaults to None
    If no input is provided, assumes the minimum value in time_series
    :return: boolean flag indicating whether there are any missing timestamps in the input time_series
    :rtype: bool
    """

@pf.register_dataframe_method
def fill_missing_timestamps(df, frequency, column_name=None, first_time_stamp=None, last_time_stamp=None): 
    """
    Re-index the data frame if there are missing timestamps found in the data frame.  
    
    :param pd.DataFrame df: 
    :param int frequency: frequency i.e. sampling frequency of the data, expressed in seconds. For e.g. if the data has
    hourly frequency, then a value of 60 * 60 = 3600 is expected
    :param str column_name: name of the column which has time series if not the index. Defaults to None. By default, the 
    index is used for checking for the timestamps, unless a value other than None is assigned to column_name 
    :param pd.Timestamp first_time_stamp: timestamp at which the time_series is expected to start from. Defaults to None
    If no input is provided, assumes the minimum value in time_series
    :param pd.Timestamp last_time_stamp: timestamp at which the time_series is expected to end with. Defaults to None
    If no input is provided, assumes the minimum value in time_series 
    :return: reindexed data frame if it turns out to have missing timestamps, else returns the input data frame as is 
    :rtype: pd.DataFrame
    """

```

UGuntupalli:
Closes #705

# PR Description

- Introducing a "timeseries.py" module with data checking and data cleaning functions geared towards time series data 
- Adding two functions (a utility function and an actual user-facing function) to fill_missing_timestamps in a timeseries data set 

**This PR resolves #705.**

# PR Checklist

         I think I am doing this, but I am really not sure, so please guide me through and I will be happy to resubmit. 

## Quick Check
**_I have made code and documentation changes, however I don't have enough experience in adding unit tests. I just added one unit test using an existing example from the code base, but I need some help with ```pytest```. I have used unittest very briefly in the past, but any guidance regarding the testing piece will be helpful for me to progress further._**  

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

```
tests\conftest.py:7: in <module>
    from janitor.testing_utils import date_data
E   ModuleNotFoundError: No module named 'janitor'
```

## Code Changes

- [ ] Ensure that you have added tests.
- [ ] Run all tests (`$ pytest .`) locally on your machine.
    - [ ] Check to ensure that test coverage covers the lines of code that you have added.
    - [ ] Ensure that all tests pass.

## Documentation Changes

- [ ] Build the docs locally.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -91,3 +91,4 @@ Contributors
 - `@smu095 <https://github.com/smu095>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3smu095>`_
 - `@VPerrollaz <https://github.com/VPerrollaz>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3AVPerrollaz>`_
 - `@UGuntupalli <https://github.com/UGuntupalli>`_ | `contributions < https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3AUGuntupalli >`_
+- `@mphirke <https://github.com/mphirke>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3Amphirke>`_
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -90,3 +90,4 @@ Contributors
 - `@evan-anderson <https://github.com/evan-anderson>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3evan-anderson>`_
 - `@smu095 <https://github.com/smu095>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3smu095>`_
 - `@VPerrollaz <https://github.com/VPerrollaz>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3AVPerrollaz>`_
+- `@UGuntupalli <https://github.com/UGuntupalli>`_ | `contributions < https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3AUGuntupalli >`_
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ new version (on deck)
 - [ENH] Added ``process_text`` wrapper function for all Pandas string methods. @samukweku
 - [TST] Only skip tests for non-installed libraries on local machine. @hectormz
 - [DOC] Fix minor issues in documentation. @hectormz
+- [ENH] Introducing timeseries module. Adding ```fill_missing_timestamps``` function @UGuntupalli 
 
 
 v0.20.7
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ new version (on deck)
 - [TST] Updated test_expand_grid to account for int dtype difference in Windows OS @samukweku
 - [TST] Make importing ``pandas`` testing functions follow uniform pattern. @hectormz
 - [ENH] Added ``process_text`` wrapper function for all Pandas string methods. @samukweku
+- [TST] Only skip tests for non-installed libraries on local machine. @hectormz
 
 
 v0.20.7
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ new version (on deck)
 - [TST] Make importing ``pandas`` testing functions follow uniform pattern. @hectormz
 - [ENH] Added ``process_text`` wrapper function for all Pandas string methods. @samukweku
 - [TST] Only skip tests for non-installed libraries on local machine. @hectormz
+- [DOC] Fix minor issues in documentation. @hectormz
 
 
 v0.20.7
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ new version (on deck)
 - [DOC] Add the Optional type for all signatures of the API. @VPerrollaz
 - [TST] Updated test_expand_grid to account for int dtype difference in Windows OS @samukweku
 - [TST] Make importing ``pandas`` testing functions follow uniform pattern. @hectormz
+- [ENH] Added ``process_text`` wrapper function for all Pandas string methods. @samukweku
 
 
 v0.20.7
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,11 @@
 new version (on deck)
 =====================
+- [ENH] Updated groupby_agg function to account for null entries in the ``by`` argument. @samukweku
+- [ENH] Added function ``groupby_topk`` to janitor functions @mphirke
+
+
+v0.20.8
+=======
 - [ENH] Upgraded ``update_where`` function to use either the pandas query style,
   or boolean indexing via the ``loc`` method.
   Also updated ``find_replace`` function to use the ``loc`` method directly,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ new version (on deck)
 - [ENH] Added ``process_text`` wrapper function for all Pandas string methods. @samukweku
 - [TST] Only skip tests for non-installed libraries on local machine. @hectormz
 - [DOC] Fix minor issues in documentation. @hectormz
+- [ENH] Added ``fill_direction`` function for forward/backward fills on missing values
+  for selected columns in a dataframe. @samukweku
 - [ENH] Introducing timeseries module. Adding ```fill_missing_timestamps``` function @UGuntupalli 
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,8 @@ new version (on deck)
 - [DOC] Fix minor issues in documentation. @hectormz
 - [ENH] Added ``fill_direction`` function for forward/backward fills on missing values
   for selected columns in a dataframe. @samukweku
-- [ENH] Introducing timeseries module. Adding ```fill_missing_timestamps``` function @UGuntupalli 
+- [ENH] Simpler logic and less lines of code for expand_grid function @samukweku
+ 
 
 
 v0.20.7
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -57,10 +57,10 @@ def _flag_missing_timestamps(
     # Check if they are the same
     comparison_index = expected_timestamps.difference(actual_timestamps)
     if comparison_index.empty:
+        result['flag'] = False
+    else:
         result['flag'] = True
         result['new_index'] = expected_timestamps
-    else:
-        result['flag'] = False
 
     # Return the result as a Named Tuple
     return MissingTimeStampFlag._make(result)
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -19,18 +19,22 @@ def _flag_missing_timestamps(
     """
     Test if timestamps are missing
 
-    Utility function to test if input data frame is missing any timestamps relative to expected timestamps.
-    They are generated based on the first_time_stamp, last_time_stamp and frequency.
+    Utility function to test if input data frame
+    is missing timestamps relative to expected timestamps.
+    They are generated based on the first_time_stamp,
+    last_time_stamp
+    and frequency.
 
-    :param df: data frame which needs to be tested for missing timestamps
-    :param frequency: frequency i.e. sampling frequency of the data.
-        A list of acceptable frequency strings are available here:
-        (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
-    :param column_name: name of the column which has time series if not the index.
-    :param first_time_stamp: timestamp at which the time_series is expected to start from.
-    :param last_time_stamp: timestamp at which the time_series is expected to end with.
+    :param df: data frame to test for missing timestamps
+    :param frequency: frequency i.e. sampling frequency
+        Acceptable frequency strings are available
+        [Reference](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases/):
+    :param column_name: name of the column which has time series if not the index
+    :param first_time_stamp: timestamp the time_series is expected to start from
+    :param last_time_stamp: timestamp the time_series is expected to end with
     :return: namedtuple with 3 attributes namely flag, raw_data and new_index
-        1. flag - boolean set to True if there are missing timestamps, else set to False
+        1. flag - boolean set to True if there are missing timestamps,
+            else set to False
         2. raw_data - input data frame as is without any modifications
         3. new_index - pd.DateTimeIndex that can be used to set the new index.
             Defaults to None.
@@ -49,10 +53,10 @@ def _flag_missing_timestamps(
 
     # Get actual timestamps
     if column_name:
-        df = df.set_index(column_name, inplace=True)
+        df = df.set_index(column_name)
 
     df = df.sort_index(inplace=True)
-    actual_timestamps = df.index.values
+    actual_timestamps = df.index.array
 
     # Check if they are the same
     comparison_index = expected_timestamps.difference(actual_timestamps)
@@ -74,16 +78,16 @@ def fill_missing_timestamps(
     last_time_stamp: pd.Timestamp = None,
 ) -> pd.DataFrame:
     """
-    Fills data frame with missing timestamps if found to be missing.
+    Fills data frame with missing timestamps if missing
 
-    Test the data frame for missing timestamps.
-    If timestamps are missing, then Re-indexes the data frame.
-    If timestamps are not missing, original data frame will be returned.
+    Test the data frame for missing timestamps
+    If timestamps are missing, Re-indexes the data frame
+    If timestamps are not missing, returns original data frame
 
     :param df: data frame which needs to be tested for missing timestamps
     :param frequency: frequency i.e. sampling frequency of the data.
-        A list of acceptable frequency strings are available here:
-        (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
+        Acceptable frequency strings are available
+        [Reference](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases/):
     :param column_name: name of the column which has time series if not the index.
         Defaults to None.
         By default the index is used for checking for the timestamps,
@@ -123,6 +127,6 @@ def fill_missing_timestamps(
 
     # Return result based on whether timestamps are missing or not
     if timestamps_missing_flag["flag"]:
-        df = df.set_index(timestamps_missing_flag["new_index"], inplace=True)
+        df = df.set_index(timestamps_missing_flag["new_index"])
         return df
     return timestamps_missing_flag["raw_data"]
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -67,7 +67,8 @@ def _flag_missing_timestamps(
     result["new_index"] = expected_timestamps
 
     # Return the result as a Named Tuple
-    return MissingTimeStampFlag._make(list(result.values()))
+    values = [v for k, v in result.items()]
+    return MissingTimeStampFlag._make(values)
 
 
 @pf.register_dataframe_method
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -37,15 +37,15 @@ def _flag_missing_timestamps(
             Assigned a value only when flag is set to True
     """
     # Declare a named tuple to hold results
-    MissingTimeStampFlag = namedtuple('MissingTimeStampFlag', ['flag', 'raw_data', 'new_index'])
-    result = {
-        'flag': None,
-        'raw_data': df.copy(deep=True),
-        'new_index': None
-    }
+    MissingTimeStampFlag = namedtuple(
+        "MissingTimeStampFlag", ["flag", "raw_data", "new_index"]
+    )
+    result = {"flag": None, "raw_data": df.copy(deep=True), "new_index": None}
 
     # Generate expected timestamps
-    expected_timestamps = pd.date_range(start=first_time_stamp, end=last_time_stamp, frequency=frequency)
+    expected_timestamps = pd.date_range(
+        start=first_time_stamp, end=last_time_stamp, frequency=frequency
+    )
 
     # Get actual timestamps
     if column_name:
@@ -57,10 +57,9 @@ def _flag_missing_timestamps(
     # Check if they are the same
     comparison_index = expected_timestamps.difference(actual_timestamps)
     if comparison_index.empty:
-        result['flag'] = False
-    else:
-        result['flag'] = True
-        result['new_index'] = expected_timestamps
+        result["flag"] = False
+    result["flag"] = True
+    result["new_index"] = expected_timestamps
 
     # Return the result as a Named Tuple
     return MissingTimeStampFlag._make(result)
@@ -109,7 +108,9 @@ def fill_missing_timestamps(
     if column_name:
         test_new_index_data_type = is_datetime(df[column_name])
         if not test_new_index_data_type:
-            raise TypeError("\n column_name should refer to a column whose data type is datetime")
+            raise TypeError(
+                "\n column_name should refer to a column whose data type is datetime"
+            )
     if not first_time_stamp:
         first_time_stamp = df.index.min()
     if not last_time_stamp:
@@ -117,16 +118,11 @@ def fill_missing_timestamps(
 
     # Test if there are any timestamps missing
     timestamps_missing_flag = _flag_missing_timestamps(
-        df,
-        frequency,
-        column_name,
-        first_time_stamp,
-        last_time_stamp
+        df, frequency, column_name, first_time_stamp, last_time_stamp
     )
 
     # Return result based on whether timestamps are missing or not
-    if timestamps_missing_flag['flag']:
-        df = df.set_index(timestamps_missing_flag['new_index'], inplace=True)
+    if timestamps_missing_flag["flag"]:
+        df = df.set_index(timestamps_missing_flag["new_index"], inplace=True)
         return df
-    else:
-        return timestamps_missing_flag['raw_data']
+    return timestamps_missing_flag["raw_data"]
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -23,6 +23,15 @@ def fill_missing_timestamps(
     this function will reindex the dataframe.
     If timestamps are not missing,
     then the function will return the dataframe unmodified.
+    
+    Example usage:
+    
+    .. code-block:: python
+
+        df = (
+            pd.DataFrame(...)
+            .fill_missing_timestamps(frequency="1H")
+        )
 
     :param df: Dataframe which needs to be tested for missing timestamps
     :param frequency: frequency i.e. sampling frequency of the data.
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -69,7 +69,7 @@ def _get_missing_timestamps(
     frequency: str,
     first_time_stamp: pd.Timestamp = None,
     last_time_stamp: pd.Timestamp = None,
-):
+) -> pd.DataFrame:
     """
     Return the timestamps that are missing in a dataframe.
 
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -4,8 +4,6 @@ Time series-specific data testing and cleaning functions.
 
 import pandas as pd
 import pandas_flavor as pf
-from pandas.api.types import is_datetime64_any_dtype as is_datetime
-from collections import namedtuple
 from janitor import check
 
 
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -29,9 +29,9 @@ def _flag_missing_timestamps(
     :param frequency: frequency i.e. sampling frequency
         Acceptable frequency strings are available
         `Reference <https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases/>`_
-    :param column_name: name of the column which has time series if not the index
-    :param first_time_stamp: timestamp the time_series is expected to start from
-    :param last_time_stamp: timestamp the time_series is expected to end with
+    :param column_name: column which has time series if not the index
+    :param first_time_stamp: timestamp expected to start from
+    :param last_time_stamp: timestamp expected to end with
     :return: namedtuple with 3 attributes namely flag, raw_data and new_index
         1. flag - boolean set to True if there are missing timestamps,
             else set to False
@@ -88,14 +88,14 @@ def fill_missing_timestamps(
     :param frequency: frequency i.e. sampling frequency of the data.
         Acceptable frequency strings are available
         `Reference <https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases/>`_
-    :param column_name: name of the column which has time series if not the index.
+    :param column_name: column which has time series if not index
         Defaults to None.
         By default the index is used for checking for the timestamps,
         unless a value other than None is assigned to column_name
-    :param first_time_stamp: timestamp at which the time_series is expected to start from.
+    :param first_time_stamp: timestamp expected to start from
         Defaults to None.
         If no input is provided assumes the minimum value in time_series
-    :param last_time_stamp: timestamp at which the time_series is expected to end with.
+    :param last_time_stamp: timestamp expected to end with.
         Defaults to None.
         If no input is provided, assumes the maximum value in time_series
     :return: reindexed data frame if it turns out to have missing timestamps,
@@ -113,7 +113,7 @@ def fill_missing_timestamps(
         test_new_index_data_type = is_datetime(df[column_name])
         if not test_new_index_data_type:
             raise TypeError(
-                "\n column_name should refer to a column whose data type is datetime"
+                f""" \n {column_name} data type must be datetime"""
             )
     if not first_time_stamp:
         first_time_stamp = df.index.min()
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -28,7 +28,7 @@ def _flag_missing_timestamps(
     :param df: data frame to test for missing timestamps
     :param frequency: frequency i.e. sampling frequency
         Acceptable frequency strings are available
-        [Reference](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases/):
+        `Reference <https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases/>`_
     :param column_name: name of the column which has time series if not the index
     :param first_time_stamp: timestamp the time_series is expected to start from
     :param last_time_stamp: timestamp the time_series is expected to end with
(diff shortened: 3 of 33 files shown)


AUTHORS.rst
```diff
@@ -90,3 +90,5 @@ Contributors
 - `@evan-anderson <https://github.com/evan-anderson>`_ | `contributions <https://github.com/ericmjl/pyjanitor/pulls?utf8=%E2%9C%93&q=is%3Aclosed+mentions%3evan-anderson>`_
 - `@smu095 <https://github.com/smu095>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3smu095>`_
 - `@VPerrollaz <https://github.com/VPerrollaz>`_ | `contributions <https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3AVPerrollaz>`_
+- `@UGuntupalli <https://github.com/UGuntupalli>`_ | `contributions <https://github.com/UGuntupalli/pyjanitor/tree/issue_705_flag_and_fill_timestamps>`_
+    - Introducing data checking and data cleaning functions for timeseries data
```
ericmjl: The `AUTHORS.rst` is intended just to list the authors. 

Here's the proposed change:

```suggestion
- `@UGuntupalli <https://github.com/UGuntupalli>`_ | `contributions < https://github.com/ericmjl/pyjanitor/issues?q=is%3Aclosed+mentions%3AUGuntupalli >`_
```

CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ new version (on deck)
 - [ENH] Added ``process_text`` wrapper function for all Pandas string methods. @samukweku
 - [TST] Only skip tests for non-installed libraries on local machine. @hectormz
 - [DOC] Fix minor issues in documentation. @hectormz
+- [ENH] Introducing timeseries module. Adding ```fill_missing_timestamps``` function @UGuntupalli
```
ericmjl: 🎉 🎉 🎉 🎉

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+def _flag_missing_timestamps(
+    df: pd.DataFrame,
+    frequency: str,
+    column_name: str,
+    first_time_stamp: pd.Timestamp,
+    last_time_stamp: pd.Timestamp,
+) -> namedtuple:
+    """
+    Utility function to test if input data frame is missing any timestamps relative to expected timestamps
+    generated based on the first_time_stamp, last_time_stamp and frequency.
```
ericmjl: First line should be a concise description. 

Here's a proposed change:

```suggestion
    Flag timestamps that  are missing.

    Utility function to test if input data frame is missing any timestamps
    relative to expected timestamps 
    generated based on the first_time_stamp, last_time_stamp and frequency.
```

Also incorporated [semantic line breaks](https://sembr.org/) into that line. I probably didn't include this in the contribution docs, though I should. Would you be kind enough to fix the others? Using best judgment here is good enough, don't need to be _too_ strict about it.

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+
+    :param pd.DataFrame df: data frame which needs to be tested for missing timestamps
+    :param str frequency: frequency i.e. sampling frequency of the data, expressed in seconds. A list of acceptable
+    frequency strings are available here
+    (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
+    :param str column_name: name of the column which has time series if not the index.
+    :param pd.Timestamp first_time_stamp: timestamp at which the time_series is expected to start from.
+    :param pd.Timestamp last_time_stamp: timestamp at which the time_series is expected to end with.
```
ericmjl: Types not needed, as they are in the function signature already.

```suggestion
    :param df: data frame which needs to be tested for missing timestamps
    :param frequency: frequency i.e. sampling frequency of the data, expressed in seconds. A list of acceptable
    frequency strings are available here
    (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
    :param column_name: name of the column which has time series if not the index.
    :param first_time_stamp: timestamp at which the time_series is expected to start from.
    :param last_time_stamp: timestamp at which the time_series is expected to end with.
```

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+    last_time_stamp: pd.Timestamp,
+) -> namedtuple:
+    """
+    Utility function to test if input data frame is missing any timestamps relative to expected timestamps
+    generated based on the first_time_stamp, last_time_stamp and frequency.
+
+    :param pd.DataFrame df: data frame which needs to be tested for missing timestamps
+    :param str frequency: frequency i.e. sampling frequency of the data, expressed in seconds. A list of acceptable
+    frequency strings are available here
+    (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
```
ericmjl: Proposing [semantic line break](https://sembr.org/) changes:

```suggestion
    :param str frequency: frequency i.e. sampling frequency of the data,
        expressed in seconds.
        A list of acceptable frequency strings are available here:
        https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases
```

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+    :param str column_name: name of the column which has time series if not the index.
+    :param pd.Timestamp first_time_stamp: timestamp at which the time_series is expected to start from.
+    :param pd.Timestamp last_time_stamp: timestamp at which the time_series is expected to end with.
+    :return: namedtuple with 3 attributes namely flag, raw_data and new_index
+        1. flag - boolean set to True if there are missing timestamps, else set to False
+        2. raw_data - input data frame as is without any modifications
+        3. new_index - pd.DateTimeIndex that can be used to set the new index, defaults to None, assigned a value only
+        when flag is set to True
+    :rtype: namedtuple
```
ericmjl: Not necessary, as the return type is in the function signature.
```suggestion
```

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+        'raw_data': df.copy(deep=True),
+        'new_index': None
+    }
+
+    # Generate expected timestamps
+    expected_timestamps = pd.date_range(start=first_time_stamp, end=last_time_stamp, frequency=frequency)
+
+    # Get actual timestamps
+    if column_name:
+        df.set_index(column_name, inplace=True)
```
ericmjl: This line will end up mutating the original dataframe. I think as a rule of thumb it's probably better to avoid mutating the original. Could we do a reassignnment instead?

```suggestion
        df = df.set_index(column_name)
```

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+
+    df.sort_index(inplace=True)
```
ericmjl: Same comment about mutating vs. not mutating the original object.

```suggestion
    df = df.sort_index()
```

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+def fill_missing_timestamps(
+    df: pd.DataFrame,
+    frequency: str,
+    column_name: str = None,
+    first_time_stamp: pd.Timestamp = None,
+    last_time_stamp: pd.Timestamp = None,
+) -> pd.DataFrame:
+    """
+    Fills data frame with missing timestamps if found to be missing. Re-index the data frame if there are missing
+    timestamps else no changes wil be made and the original data frame will be returned.
```
ericmjl: Same comment as above w.r.t.  docstring first lines. (I just remembered, the convention is detailed in [here](https://www.python.org/dev/peps/pep-0257/).) Also, could you introduce SEMBR changes please?

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+    frequency strings are available here
+    (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
+    :param str column_name: name of the column which has time series if not the index. Defaults to None. By default, the
+    index is used for checking for the timestamps, unless a value other than None is assigned to column_name
+    :param pd.Timestamp first_time_stamp: timestamp at which the time_series is expected to start from. Defaults to None
+    If no input is provided, assumes the minimum value in time_series
+    :param pd.Timestamp last_time_stamp: timestamp at which the time_series is expected to end with. Defaults to None
+    If no input is provided, assumes the maximum value in time_series
+    :return: reindexed data frame if it turns out to have missing timestamps, else returns the input data frame as is
+    :rtype: pd.DataFrame
```
ericmjl: A few things to look out here:

1. Types not necessary.
2. Indent line continuations, this makes recognizing the individual `:param:`s much easier.
3. SEMBR.

janitor/timeseries.py
```diff
@@ -0,0 +1,123 @@
...
+        df,
+        frequency,
+        column_name,
+        first_time_stamp,
+        last_time_stamp
+    )
+
+    # Return result based on whether timestamps are missing or not
+    if timestamps_missing_flag['flag']:
+        df.set_index(timestamps_missing_flag['new_index'], inplace=True)
```
ericmjl:
```suggestion
        df = df.set_index(timestamps_missing_flag['new_index'])
```

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
ericmjl, line 1: I sense this might be your first attempt at writing tests, is this right, @UGuntupalli? If so, you've made a great first try - not many will get this far!

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,24 @@
...
+        )
+        assert fill_missing_timestamps(
+            pd.DataFrame([]), '1B', 1, None, None
+        )
+        assert fill_missing_timestamps(
+            pd.DataFrame([]), '1B', 'DateTime', '2020-01-01', None
+        )
+        assert fill_missing_timestamps(
+            pd.DataFrame([]), '1B', 'DateTime', pd.Timestamp(2020, 1, 1), '2020-02-01'
+        )
```
ericmjl: In this block, the `asserts` aren't necessary. That is because a simple execution of the function `fill_missing_timestamps` will result in `TypeError`s for each of the examples you've raised.

I would suggest also parametrizing the test rather than calling on `fill_missing_timestamps` five times. My other comment above will have shown you how to do that.

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+import pandas as pd
+from janitor.timeseries import fill_missing_timestamps
+
+
+@pytest.mark.timeseries
+def test_datatypes_check():
```
ericmjl: I'd like to suggest parametrizing the test. The full docs on how to do so are available [here](https://docs.pytest.org/en/stable/parametrize.html), but here's a lightweight example.

```suggestion
@pytest.mark.parametrize(
    "df,frequency,column_name,first_time_stamp,last_time_stamp",
    [
        (4, '1T', None, None, None),
        (pd.DataFrame([]), 2, None, None, None),
        (pd.DataFrame([]), '1B', 1, None, None),
        (pd.DataFrame([]), '1B', 'DateTime', '2020-01-01', None),
        (pd.DataFrame([]), '1B', 'DateTime', pd.Timestamp(2020, 1, 1), '2020-02-01'),
]
def test_datatypes_check(df, frequency, column_name, first_time_stamp, last_time_stamp):
```

In doing so, you can make the test block:

```python
with pytest.raises(TypeError):
    fill_missing_timestamps(df, frequency, column_name, first_time_stamp, last_time_stamp)
```

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,24 @@
...
+        assert fill_missing_timestamps(
+            pd.DataFrame([]), '1B', 1, None, None
+        )
+        assert fill_missing_timestamps(
+            pd.DataFrame([]), '1B', 'DateTime', '2020-01-01', None
+        )
+        assert fill_missing_timestamps(
+            pd.DataFrame([]), '1B', 'DateTime', pd.Timestamp(2020, 1, 1), '2020-02-01'
+        )
+
```
ericmjl, line 44: Apart from the error-checking test above, I think it may help to provide a test that is small enough for others to read and execute inside a test. Could you add that in please?

tests/timeseries/test_fill_missing_timestamps.py
UGuntupalli: Yes, this is my first time by myself, especially with pytest. Appreciate your patience in guiding me through the testing piece, I hope to quickly pick up atleast the basics to make successful contribution

tests/timeseries/test_fill_missing_timestamps.py
UGuntupalli: Perhaps something like this is what you have in mind ? 

```python
@pytest.mark.timeseries
def test_fill_missing_timestamps():
    assert fill_missing_timestamps(
        pd.DataFrame([]), '1B', 'DateTime', pd.Timestamp(2020, 1, 1), pd.Timestamp(2020, 2, 1)
    )
```

ericmjl: Almost! The `assert fill_missing_timestamps...` isn’t necessary, because the return value of `fill_missing_timestamps` is not the thing we need tested, but rather, _properties_ of the return value. For example, an appropriate assertion here is to assert that the timestamps that are added are correct.

I would suggest the following pattern sketch:

```python
result = fill_missing_timestams(...)
assert result[“something”] == something
```

tests/timeseries/test_fill_missing_timestamps.py
ericmjl: You're doing great, @UGuntupalli!

ericmjl: You're doing great!

UGuntupalli:
@ericmjll, 
       When you have the opportunity, please let me know if there is anything else missing from this PR that I should be taking care of

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,41 @@
...
+)
+def test_datatypes_check(df, frequency, column_name, first_time_stamp, last_time_stamp):
+    with pytest.raises(TypeError):
+        fill_missing_timestamps(df, frequency, column_name, first_time_stamp, last_time_stamp)
+
+
+@pytest.mark.timeseries
+def test_fill_missing_timestamps():
+    # Remove random row from the data frame
+    df1 = test_df.drop(test_df.sample())
```
ericmjl: Can we make the test_df a fixture that gets passed in? This is generally best practice, to avoid the situation where an object being tested accidentally gets modified in one test before getting used in another test.

The docs for fixtures are here: https://docs.pytest.org/en/stable/fixture.html

And the general way we’d write fixtures is as follows:


```python
@pytest.fixture
def my_dataframe():
    # stuff
    return mydf


# argument must be named the same as the pytest fixture
def test_my_function(my_dataframe):
    # stuff, use my_dataframe as you would a dataframe object
    assert # stuff 
```

UGuntupalli:
@ericmjl, 
         Performed the 3 corrections recommended: 

1. ran flake 
2. added pytest.fixture 
3. removed/restructured if-else blocks 

    Still running into a build error on linux and macos.  I unfortunately don't have either OS to test the same. In my attempt to fix this error, I came across this, but it is not really clear what is needed to fix these. (https://github.com/microsoft/azure-pipelines-tasks/issues/10680)

janitor/timeseries.py
```diff
@@ -0,0 +1,128 @@
...
+    result = {"flag": None, "raw_data": df.copy(deep=True), "new_index": None}
+
+    # Generate expected timestamps
+    expected_timestamps = pd.date_range(
+        start=first_time_stamp, end=last_time_stamp, frequency=frequency
+    )
+
+    # Get actual timestamps
+    if column_name:
+        df = df.set_index(column_name, inplace=True)
```
ericmjl: The `inplace=True` was not necessary. This is because adding `inplace=True` returns nothing, so the assignment of `df = df.set_index(..., inplace=True)` would make `df` become a `None` object.

```suggestion
        df = df.set_index(column_name)
```

janitor/timeseries.py
```diff
@@ -0,0 +1,128 @@
...
+
+    df = df.sort_index(inplace=True)
+    actual_timestamps = df.index.values
```
ericmjl: Flake8 is suggesting that using `.values` isn't recommend practice, and that instead we should be accessing `.array`.

I'm basing this off the CI system:

```
./janitor/timeseries.py:55:25: PD011 Use '.array' or '.to_array()' instead of '.values'; 'values' is ambiguous
```

Suggested change:

```suggestion
    actual_timestamps = df.index.array
```

janitor/timeseries.py
```diff
@@ -0,0 +1,128 @@
...
+        last_time_stamp = df.index.max()
+
+    # Test if there are any timestamps missing
+    timestamps_missing_flag = _flag_missing_timestamps(
+        df, frequency, column_name, first_time_stamp, last_time_stamp
+    )
+
+    # Return result based on whether timestamps are missing or not
+    if timestamps_missing_flag["flag"]:
+        df = df.set_index(timestamps_missing_flag["new_index"], inplace=True)
```
ericmjl: Same as above.

```suggestion
        df = df.set_index(timestamps_missing_flag["new_index"])
```

janitor/timeseries.py
```diff
@@ -0,0 +1,128 @@
...
+    frequency: str,
+    column_name: str,
+    first_time_stamp: pd.Timestamp,
+    last_time_stamp: pd.Timestamp,
+) -> namedtuple:
+    """
+    Test if timestamps are missing
+
+    Utility function to test if input data frame is missing any timestamps relative to expected timestamps.
+    They are generated based on the first_time_stamp, last_time_stamp and frequency.
```
ericmjl: One suggestion for a semantic line-breaks-compatible docstring

```suggestion
    Utility function to test if input data frame
    is missing any timestamps relative to expected timestamps.
    They are generated based on the first_time_stamp,
    last_time_stamp
    and frequency.
```

janitor/timeseries.py
```diff
@@ -0,0 +1,128 @@
...
+
+    :param df: data frame which needs to be tested for missing timestamps
+    :param frequency: frequency i.e. sampling frequency of the data.
+        A list of acceptable frequency strings are available here:
+        (https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
```
ericmjl: Hyperlinks should follow [Sphinx-style syntax][sphinx]. This slipped from my view; thankfully, linters like Flake8 help catch this :).

[sphinx]: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks

UGuntupalli:
@ericmjl, 
      Gave it 2 more shots, still failing. It is definitely educational, the only question/ suggestion I would have is it possible to view the logs when I commit and the build fails. I have a feeling it could be because of the link in the doc string, but I have no way to verify it. And this prolongs the time for completion of the issue. 

     Independent of that, could you point me to some resources on how to set such automated checks and balances for each commit ? From my limited understanding, I could realized you are using: 
1. Flake8 for PEP compliance 
2. Azure Devops
3. Travis CI/CD for continuous integration 

     We can have that conversation offline based on when you have some free time too.

ericmjl:
@UGuntupalli actually, you’ve fixed all of the problems! :smile:

If you look at the Azure DevOps pipeline, you’ll see that the main place it’s failing right now is something that will be fixed in @samukweku PR. So let’s hang tight here, once his PR is merged in, I’m going to have you merge in the latest `dev` branch. 

Are you familiar with how to do this? If not, here’s the main ideas:

1. Make sure you have a git remote pointing to ericmjl/pyjanitor, e.g. called `upstream`.
1. On your local machine, checkout dev branch: `git checkout dev`
1. Pull latest changes: `git fetch upstream && git merge upstream/dev` —> this should pull latest dev from my branch to your local, and merge it.
1. Now, checkout to the branch `issue_705_flag_and_fill_timestamps`
1. Merge local dev into your branch: `git merge dev`.
1. Finally, push up the changes: `git push`

That should bring the latest dev changes back into your branch, which the CI systems will then go ahead and check again.

Yes, there’s a lot of services out there that help our open source projects automatically check code all the time. We can chat another time offline. My preferred contact method is at the bottom of my [personal website](ericmjl.github.io) (protected away from public view because of the number of emails I get!).

UGuntupalli:
I got some notifications that he finished his merges and I attempted to follow the steps, currently it is running into some merge conflicts. Let me know when I can attempt it or if you can push this over the line ?

samukweku:
Hi @UGuntupalli just checking if you rebased

UGuntupalli:
@samukweku,
         Sam yes. I ran through steps 1-4 and started seeing issues on step 5 of Eric's outline. Happy to try again

ericmjl:
@UGuntupalli actually, I think you've also succeeded in pulling in the latest changes! :tada:

I looked at the build logs again -- I think you should be able to view them too, right? Click around on the "Details", and keep digging through the links on the page, until you get taken to this interface:

image

That's where all the build errors will show up. 

In this case, the build is failing [here](https://dev.azure.com/ericmjl/pyjanitor/_build/results?buildId=1867&view=logs&j=acc1347f-8235-55b0-95c7-0e2189e61659&t=486f34f1-eda6-516a-fc5d-d8195128dacd&l=127).

Does executing that test locally work for you?

UGuntupalli:
@ericmjl, 
       I am pretty sure it is not as straightforward as running the script ? I am out of my depth here with testing. How do I run the test locally ?

UGuntupalli:
@ericmjl,
           This error doesn't make sense to me. The flake recommendation is to use ```.to_array()``` while extracting the values of a dictionary ? As far as I am aware, the common documented way to get the values of a dictionary is using ```.values()````, let me know if you disagree. Now all the local tests work and are successful

samukweku:
@UGuntupalli the linter is confusing the `.values` as a method for an array, instead of a dictionary. I usually go full mode and iterate through the `.items` method. I know there is also a way to disable the error, using #noqa or other; never really worked for me.

ericmjl:
@UGuntupalli got it! And thanks @samukweku for chiming in!

My suggestion here is to always use `.items()` when looping over a dictionary, as when we write code that looks like this:

```python
for k, v in something.items():
    # stuff
```

it signals immediately to a code reader that `something` is a dictionary. I’m not aware of any other data structure in Python that allows immediate signalling of its type.

ericmjl:
Hmmm, now test coverage seems to be the issue. (This is important because test coverage is a proxy for how extensively tested the code is. Not perfect, but useful nonetheless.)

@UGuntupalli I'm ran `make test` on your fork's branch, and see where code coverage might need to be improved. The following lines aren't currently executed by the test suite:

```bash
janitor/timeseries.py                    40      6    85%   57, 65, 116-118, 135
```

Those lines correspond to the following `if` blocks:

Line 57:

```python
    # Get actual timestamps
    if column_name:
        df = df.set_index(column_name)
```

Line 65:

```python
    if comparison_index.empty:
        result["flag"] = False
```

Line 116-118:

```python
    if column_name:
        test_new_index_data_type = is_datetime(df[column_name])
        if not test_new_index_data_type:
            raise TypeError(
                f""" \n {column_name} data type must be datetime"""
            )
```

Line 135:

```python
    # Return result based on whether timestamps are missing or not
    if timestamps_missing_flag.flag:
        df = df.reindex(timestamps_missing_flag.new_index)
        return df
    return timestamps_missing_flag.raw_data
```

Could you add some tests that cover these `if` blocks? I think once those are covered, we'll be golden!

ericmjl:
@UGuntupalli just a heads-up, I've cut a new release, so please pull in the latest changes from `dev` again!

UGuntupalli:
Eric, 

> Hmmm, now test coverage seems to be the issue. (This is important because test coverage is a proxy for how extensively tested the code is. Not perfect, but useful nonetheless.)
> 
> @UGuntupalli I'm ran `make test` on your fork's branch, and see where code coverage might need to be improved. The following lines aren't currently executed by the test suite:
> 
> ```shell
> janitor/timeseries.py                    40      6    85%   57, 65, 116-118, 135
> ```
> 
> Those lines correspond to the following `if` blocks:
> 
> Line 57:
> 
> ```python
>     # Get actual timestamps
>     if column_name:
>         df = df.set_index(column_name)
> ```
> 
> Line 65:
> 
> ```python
>     if comparison_index.empty:
>         result["flag"] = False
> ```
> 
> Line 116-118:
> 
> ```python
>     if column_name:
>         test_new_index_data_type = is_datetime(df[column_name])
>         if not test_new_index_data_type:
>             raise TypeError(
>                 f""" \n {column_name} data type must be datetime"""
>             )
> ```
> 
> Line 135:
> 
> ```python
>     # Return result based on whether timestamps are missing or not
>     if timestamps_missing_flag.flag:
>         df = df.reindex(timestamps_missing_flag.new_index)
>         return df
>     return timestamps_missing_flag.raw_data
> ```
> 
> Could you add some tests that cover these `if` blocks? I think once those are covered, we'll be golden!

Testing is not my strong suite and this is a little overwhelming. Can you please provide examples on what tests you are looking for ?

ericmjl:
Yep, not a problem! Sorry if it felt like I was leaving you in the lurch, I had three incoming requests regarding pyjanitor that I had to handle, which means I wasn't really paying attention to individual contexts. 

Here's a minimal example on testing with conditionals. That should be enough to give you the general idea. 

Let's say you have a function:

```python
def relu(x):
    if x > 0:
        return x
    return 0
```

This is a classic function in neural networks, one of those activation functions.

If we were getting started, we might write a test that covers the negative case, to make sure that they all are zero. That's the most important property of the ReLU function. The test might look like this:

```python
@pytest.mark.parametrize('x', np.linspace(-1000, 0))
def test_relu_negatives(x):
    assert relu(x) == 0
```

However, that ignores the `if` block. We need to set up another test that covers the positive numbers case.

```python
@pytest.mark.parametrize('x', np.linspace(0, 1000))
def test_relu_positives(x):
    assert relu(x) == x
```

By adding in the second test, we now have a test suite that hits every single line of the function. Previously, by testing only the negatives, the `return x` line wasn't hit, because every example passed in failed the conditional `if x > 0:`.

In your case, we need a few more tests in which each of the following conditionals evaluate to True:

- `if comparison_index.empty`
- `if column_names`
- `if timestamps_missing_flag.flag`

To satisfy the code coverage piece, you'll probably need to add three more tests, one test function for each of them.

Does that make sense? If not, let me know what's tripping you up, I'll do my best to explain.

Btw, @samukweku, if you have a bit of bandwidth, do you think you could come up with some specific examples for @UGuntupalli? No pressure though, I just have to move on to other matters this evening.

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,61 @@
...
+    df, frequency, column_name, first_time_stamp, last_time_stamp
+):
+    with pytest.raises(TypeError):
+        fill_missing_timestamps(
+            df, frequency, column_name, first_time_stamp, last_time_stamp
+        )
+
+
+@pytest.mark.timeseries
+def test_fill_missing_timestamps(my_dataframe):
```
ericmjl: Let's get a docstring here that describes the intent behind the test.

tests/timeseries/test_fill_missing_timestamps.py
```diff
@@ -0,0 +1,61 @@
...
+        (
+            pd.DataFrame([]),
+            "1B",
+            "DateTime",
+            pd.Timestamp(2020, 1, 1),
+            "2020-02-01",
+        ),
+    ],
+)
+def test_datatypes_check(
```
ericmjl: Let's get a docstring here that describes the intent behind the test.

Probably something like: "Check that TypeError is raised when invalid types are passed in."

ericmjl:
Looping back, to keep the PR moving. I found another chunk of time to think of an example, @UGuntupalli. Currently, you have the test `test_fill_missing_timestamps(my_dataframe)`, but in it, the case where `column_name` is set is not tested. If you redesign that test such that `column_name` becomes parametrized, that would help increase the test coverage, because that if block would get executed.

To handle the case where `if comparison_index.empty`, we need one test where there is no difference between `expected_timestamps` and `actual_timestamps`. A trivial case would be to pass in a DataFrame that has no missing timestamps, and then that if block will get executed.

These probably can be two extra test functions in the test suite. Hopefully this helps you here, @UGuntupalli!

samukweku:
@ericmjl , just going through the code, and your suggestions are spot on!

UGuntupalli:
@ericmjl  & @samukweku , 
         Of the 3 tests, I was able to add 1.  I attempted to add a test to capture the other two cases, but I am unable to get it to run and I don't know how to fix it. 

```python
@pytest.mark.timeseries
def test_comparison_index(my_time_series_dataframe):
    result = fill_missing_timestamps(my_time_series_dataframe, frequency="1H")
    assert my_time_series_dataframe.equals(result)


@pytest.mark.parametrize(
    "df,frequency,column_name,first_time_stamp,last_time_stamp",
    [
        (my_dataframe,
         "1H",
         "DateTime",
         pd.Timestamp(2019, 1, 1, 0, 0, 0),
         pd.Timestamp(2019, 2, 11, 15, 0, 0)
         ),
    ],
)
def test_index_assignment(
        df, frequency, column_name, first_time_stamp, last_time_stamp
):
    if column_name:
        df = df.set_index(column_name)
    # result = fill_missing_timestamps(
    #     df, frequency, column_name, first_time_stamp, last_time_stamp
    # )
    assert df[column_name] == pd.date_range("1/1/2019", periods=1000, freq="1H")
```

ericmjl:
@UGuntupalli great stuff! Can you push up the changes so we can see how they run? Seeing the CI system's output helps a ton with the review.

I think I forgot this piece in the last review comment I left, but could you add in a docstring describing what each test function is intended to check? If you look at a few of the other tests, you should be able to see the level of detail that is helpful. (Hynek Schwalak has a great [blog post](https://hynek.me/articles/document-your-tests/) on why documenting tests is helpful!)

janitor/timeseries.py
```diff
@@ -0,0 +1,135 @@
...
+        df = df.set_index(column_name)
+
+    df = df.sort_index()
+    actual_timestamps = df.index.array
+
+    # Check if they are the same
+    comparison_index = expected_timestamps.difference(actual_timestamps)
+    if comparison_index.empty:
+        result["flag"] = False
+    result["flag"] = True
```
ericmjl: @UGuntupalli I found a chunk of time to go through the code in more detail, and I found this to be a bug.
It should have been the other way around:

```python
result["flag"] = True
if comparison_index.empty:
    result["flag"] = False
```

This bug resulted in `result["flag"]` never being false.

I took a deeper look at your functions' implementations, and I am going to propose a rework of the function. `_flag_missing_timestamps` seems to be doing one too many things here, and the `namedtuple` data structure, I think, might be unnecessary.

No worries, this is a fun puzzle to try to untangle. My goal here is to get functions to do one thing and _only_ one thing; I'll be eliminating the helper namedtuple, as it seems to be holding state where not necessary.

ericmjl:
@UGuntupalli I went through the codebase and looked at it again. Indeed, there was a bit of too much complexity present in `_flag_missing_timestamps`, so I took the liberty of reworking the code a bit.

Firstly, I gathered from our original discussion that you wanted some way to `_flag_missing_timestamps`. Logically, this can never be displayed in the original dataframe, as the missing timestamps in the original are by definition going to be a subset of what's expected.

I thus restructured the code such that `_flag_missing_timestamps` can be reimplemented later in terms of more primitive operations that only involve dataframes.

For now, there's two functions:

1. A much more simplified `fill_missing_timestamps`, which assumes that the index is TimeStamp type, so no reindexing by a column happens;
2. A private `_get_missing_timestamps` function, which returns a DataFrame that has all of the missing timestamps.

Testing whether there are missing timestamps or not is a matter of calling on the private function and testing the "truthiness" of its return dataframe. An empty dataframe should evaluate to `False`, and a non-empty one should be `True`. 

To help move the PR along, I created two GitHub Gists, which contain the reworked tests that you can copy directly in.

- [`test_fill_missing_timestamps.py`](https://gist.github.com/ericmjl/4a4fa228cfde41ba5ff52b9281fa240b)
- [`timeseries.py`](https://gist.github.com/ericmjl/6684ff43da033a1baec706a34df0c0f1)

If you're agreeable with this, I propose we go with that for now. What `_flag_missing_timestamps` should or should not return has to be more clearly defined, in terms of Python objects that get returned. I'd be in favour of a simpler, functional style that avoids creating new categories of objects that hold state. This would make the codebase much easier to maintain. 

I can't push directly to your branch, I think, so at this point, the ball is back in your court! Let's finish this off 😄! The core of the idea is a really good one, I'd love to see it merged in soon!

ericmjl:
@UGuntupalli I found I could actually edit directly - just had to do it via the GitHub interface. 

I took the liberty of placing in the simplified versions of the functions. As it stands, the current state of the code fits much better with the rest of the `pyjanitor` design philosophy - it's a simpler function that does one and only one thing well. Wanted to thank you for taking the initiative for putting in the PR! Working through the PR, in this case, helped with refining the ideas down to its most essential piece, and that I think helped clarify many things that were originally muddy in my head.

@samukweku and others on the team, if you'd like to, please take a look at the updated code and chime in. Otherwise, I will leave it for two more days to rest, before merging.

UGuntupalli:
@ericmjl , 
        Thank you for giving this a shot.  May I ask why the builds are still failing ?

ericmjl:
Right now, they're failing on flake8. This is probably a natural consequence of me working via the GitHub text editor rather than in VSCode, where all of my auto code-formatting tools live :smile:. No worries - can you do me a favour? Pull down the latest changes, format the code, and push them up again please?

Btw, you can click in to see the details:

image


And then: 

image


And then finally, having clicked on "View more details on Azure":

image

UGuntupalli:
Tried. Failing in some notebooks. 

tQo0S1C1dF

ericmjl:
No problem, since the unit tests are passing, I'm OK merging for now. I'm doing some maintenance on a separate branch that will fix the notebooks.

Thanks again, @UGuntupalli! I know this process seemed long, but I hope it was educational! Especially when it comes to deciding what a function is supposed to do. At least for me, the simplification of the function was a really good exercise. 

Writing tests is not a habit most data scientists have, but as I've developed on pyjanitor and other data projects at work, I realized soon that knowing how to test my own code well brought me clarity to how I thought about my data problem at hand. I now consider it an essential skill for data science, and any intern that passes by my hand learns how to do it at a basic level.

As for frustrations from the CI system - I've been there before (back when I did my first contribs to matplotlib). 5 years on, I know why they're important, but I still know they're frustrating to satisfy, especially when errors crop up that are out of our own domain.

In any case, I'm going to merge! Thanks again for initiating the PR and patiently walking it through to the end!

PR #714, opened 2020-08-08
[ENH] Added function groupby_topk

### LINKED ISSUE #710: [ENH]: Groupby Top K
# Brief Description

I would like to propose a function that allows users to take in a dataframe and return another dataframe that has the top "k" values after grouping by another column.

This has come up multiple times at work, when I have to pick out "top protein mutants in each cluster".

Posting this here in case anybody wants to take a formal stab at it, otherwise I'll get to it later.

# Example API

```python
def groupby_topk(d: pd.DataFrame, groupby_column_name: Hashable, sort_column_name: Hashable, k: int, sort_values_kwargs: Dict) -> pd.DataFrame:
    """
    Return a dataframe that has the top "k" values per group.
    
    Some docstrings to flesh out:
    - Groups: defined by the "groupby" column.
    
    Taken inspiration from https://stackoverflow.com/a/27844045/1274908
    """
    return d.groupby(groupby_column_name).apply(lambda df: df.sort_values(sort_column_name, **sort_values_kwargs).head(k))
```

Give it a thumbs up if you're interested in seeing this too!

mphirke:
# PR Description

- Creation of function groupby_topk that was suggested in https://github.com/ericmjl/pyjanitor/issues/710
- The function groups the dataframe by a specified column, then sorts it by another specified column and returns the top k entries.

**This PR resolves #710 **

# PR Checklist

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

    - [ ] Ensure that all tests pass. Spark tests are failing on my machine despite no changes to them. I assume this is because I don't have Spark/JVM, rest all work. I will try again when if I can install Spark.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 new version (on deck)
 =====================
+- [ENH] Updated groupby_agg function to account for null entries in the ``by`` argument. @samukweku
 - [ENH] Added function ``groupby_topk`` to janitor functions @mphirke
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 new version (on deck)
 =====================
+- [ENH] Added function ``groupby_topk`` to janitor functions @mphirke
 
 
 v0.20.8
@@ -21,7 +22,6 @@ v0.20.8
 - [ENH] Added ``fill_direction`` function for forward/backward fills on missing values
   for selected columns in a dataframe. @samukweku
 - [ENH] Simpler logic and less lines of code for expand_grid function @samukweku
-- [ENH] Added function ``groupby_topk`` to janitor functions @mphirke
 
 
 v0.20.7
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ v0.20.8
 - [ENH] Added ``fill_direction`` function for forward/backward fills on missing values
   for selected columns in a dataframe. @samukweku
 - [ENH] Simpler logic and less lines of code for expand_grid function @samukweku
+- [ENH] Added function ``groupby_topk`` to janitor functions @mphirke
 
 
 v0.20.7
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -4256,11 +4256,16 @@ def groupby_topk(
     sort_values_kwargs: Dict = None,
 ) -> pd.DataFrame:
     """
-    Return a dataframe that has the top `k` values grouped by
-    `groupby_column_name` and sorted by `sort_column_name`. Additional
-    parameters to the sorting (such as ascending=True) can be passed using
-    `sort_values_kwargs`. List of all sort_values() parameters can be found at
-    https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html
+    Return top `k` rows from a groupby of a set of columns.
+
+    Returns a dataframe that has the top `k` values grouped by `groupby_column_name`
+    and sorted by `sort_column_name`.
+    Additional parameters to the sorting (such as ascending=True)
+    can be passed using `sort_values_kwargs`.
+
+    List of all sort_values() parameters can be found here_.
+
+    .. _here: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html
 
 
     .. code-block:: python
@@ -4334,11 +4339,9 @@ def groupby_topk(
         `groupby_column_name` column with each group sorted along the
         column `sort_column_name`.
     :raises: ValueError if `k` is less than 1.
-    :raises: ValueError if `groupb y_column_name` does not exist in dataframe.
-    :raises: ValueError if `sort_column_name` does not exist in dataframe.
     :raises: ValueError if `groupby_column_name` is same as `sort_column_name`.
     :raises: KeyError if `inplace:True` is present in `sort_values_kwargs`.
-    """
+    """  # noqa: E501
 
     # Convert the default sort_values_kwargs from None to empty Dict
     sort_values_kwargs = sort_values_kwargs or {}
@@ -4347,18 +4350,6 @@ def groupby_topk(
     if k < 1:
         raise ValueError(
             "Numbers of rows per group to be returned must be greater than 0."
-        )
-
-    # Check if groupby_column_name exists in dataframe.
-    if groupby_column_name not in df.columns:
-        raise ValueError(
-            f"Column {groupby_column_name} does not exist in the dataframe."
-        )
-
-    # Check if sort_column_name exists in dataframe.
-    if sort_column_name not in df.columns:
-        raise ValueError(
-            f"Column {sort_column_name} does not exist in the dataframe."
         )
 
     # Check if group by column and sort column are unique.
@@ -4370,7 +4361,7 @@ def groupby_topk(
         "inplace" in sort_values_kwargs.keys()
         and sort_values_kwargs["inplace"]
     ):
-        raise KeyError("Cannot use inplace : True in sort_values_kwargs.")
+        raise KeyError("Cannot use `inplace=True` in `sort_values_kwargs`.")
 
     return df.groupby(groupby_column_name).apply(
         lambda d: d.sort_values(sort_column_name, **sort_values_kwargs).head(k)
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -4339,7 +4339,6 @@ def groupby_topk(
         `groupby_column_name` column with each group sorted along the
         column `sort_column_name`.
     :raises: ValueError if `k` is less than 1.
-    :raises: ValueError if `groupby_column_name` is same as `sort_column_name`.
     :raises: KeyError if `inplace:True` is present in `sort_values_kwargs`.
     """  # noqa: E501
 
@@ -4351,10 +4350,6 @@ def groupby_topk(
         raise ValueError(
             "Numbers of rows per group to be returned must be greater than 0."
         )
-
-    # Check if group by column and sort column are unique.
-    if sort_column_name == groupby_column_name:
-        raise ValueError("Group by column and sort column cannot be same.")
 
     # Check if inplace:True in sort values kwargs because it returns None
     if (
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -1901,7 +1901,7 @@ def limit_column_characters(
     col_names = [col_name[:column_length] for col_name in col_names]
 
     col_name_set = set(col_names)
-    col_name_count = dict()
+    col_name_count = {}
 
     # If no columns are duplicates, we can skip the loops below.
     if len(col_name_set) == len(col_names):
@@ -3280,12 +3280,12 @@ class DataDescription:
     def __init__(self, data):
         """Initialize DataDescription class."""
         self._data = data
-        self._desc = dict()
+        self._desc = {}
 
     def _get_data_df(self) -> pd.DataFrame:
         df = self._data
 
-        data_dict = dict()
+        data_dict = {}
         data_dict["column_name"] = df.columns.tolist()
         data_dict["type"] = df.dtypes.tolist()
         data_dict["count"] = df.count().tolist()
@@ -4360,11 +4360,16 @@ def groupby_topk(
         `groupby_column_name` column with each group sorted along the
         column `sort_column_name`.
     :raises: ValueError if `k` is less than 1.
+    :raises: ValueError if `groupby_column_name` not in dataframe `df`.
+    :raises: ValueError if `sort_column_name` not in dataframe `df`.
     :raises: KeyError if `inplace:True` is present in `sort_values_kwargs`.
     """  # noqa: E501
 
     # Convert the default sort_values_kwargs from None to empty Dict
     sort_values_kwargs = sort_values_kwargs or {}
+
+    # Check if groupby_column_name and sort_column_name exists in the dataframe
+    check_column(df, [groupby_column_name, sort_column_name])
 
     # Check if k is greater than 0.
     if k < 1:
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -4245,3 +4245,128 @@ def fill_direction(
                 df.loc[:, column].ffill(limit=limit).bfill(limit=limit)
             )
     return df
+
+
+@pf.register_dataframe_method
+def groupby_topk(
+    df: pd.DataFrame,
+    groupby_column_name: Hashable,
+    sort_column_name: Hashable,
+    k: int,
+    sort_values_kwargs: Dict = {},
+) -> pd.DataFrame:
+    """
+    Return a dataframe that has the top `k` values grouped by
+    `groupby_column_name` and sorted by `sort_column_name`. Additional
+    parameters to the sorting (such as ascending=True) can be passed using
+    `sort_values_kwargs`. List of all sort_values() parameters can be found at
+    https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html
+
+
+    .. code-block:: python
+
+        import pandas as pd
+        import janitor as jn
+
+        df = pd.DataFrame({'age' : [20, 22, 24, 23, 21, 22],
+                           'ID' : [1,2,3,4,5,6],
+                           'result' : ["pass", "fail", "pass",
+                                       "pass", "fail", "pass"]})
+
+        # Ascending top 3:
+        df.groupby_topk('result', 'age', 3)
+        #       age  ID  result
+        #result
+        #fail   21   5   fail
+        #       22   2   fail
+        #pass   20   1   pass
+        #       22   6   pass
+        #       23   4   pass
+
+        #Descending top 2:
+        df.groupby_topk('result', 'age', 2, {'ascending':False})
+        #       age  ID result
+        #result
+        #fail   22   2   fail
+        #       21   5   fail
+        #pass   24   3   pass
+        #       23   4   pass
+
+    Functional usage syntax:
+
+    .. code-block:: python
+
+        import pandas as pd
+        import janitor as jn
+
+        df = pd.DataFrame(...)
+        df = jn.groupby_topk(
+            df = df,
+            groupby_column_name = 'groupby_column',
+            sort_column_name = 'sort_column',
+            k = 5
+            )
+
+    Method-chaining usage syntax:
+
+    .. code-block:: python
+
+        import pandas as pd
+        import janitor as jn
+
+        df = (
+            pd.DataFrame(...)
+            .groupby_topk(
+            df = df,
+            groupby_column_name = 'groupby_column',
+            sort_column_name = 'sort_column',
+            k = 5
+            )
+        )
+
+    :param df: A pandas dataframe.
+    :param groupby_column_name: Column name to group input dataframe `df` by.
+    :param sort_column_name: Name of the column to sort along the
+        input dataframe `df`.
+    :param k: Number of top rows to return from each group after sorting.
+    :param sort_values_kwargs: Arguments to be passed to sort_values function.
+    :returns: A pandas dataframe with top `k` rows that are grouped by
+        `groupby_column_name` column with each group sorted along the
+        column `sort_column_name`.
+    :raises: ValueError if `k` is less than 1.
+    :raises: ValueError if `groupb y_column_name` does not exist in dataframe.
+    :raises: ValueError if `sort_column_name` does not exist in dataframe.
+    :raises: ValueError if `groupby_column_name` is same as `sort_column_name`.
+    """
+    # Check if k is greater than 0.
+    if k < 1:
+        raise ValueError(
+            "Numbers of rows per group to be returned must be greater than 0."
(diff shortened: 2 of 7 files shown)


mphirke:
Hi, I would like some help with the error message I am getting in the checks. I created the .rst file and ran make_docs which created the build folder in which I confirmed the .HTML file for the function is created. The .html is not detected when I try to commit it. Clearly I'm doing something wrong in the docs section but I'm not sure where the problem is.

This is the error that Azure pipeline gives:

```Traceback (most recent call last):
  File "scripts/check-autodoc.py", line 73, in <module>
    f"The following functions have not yet been added: {missing_funcs}"
Exception: The following functions have not yet been added: {'groupby_topk'}
```

`

ericmjl:
Hi @mphirke! Thanks for your work here! I think that script's not-so-easy-to-follow error message is my responsibility.

Can you do two things on your PR please?

Firstly, add the function name `groupby_topk` to an appropriate section under `docs/reference/general_functions.rst`.

Secondly, can you modify the error message in `check-autodoc.py` such that it more clearly spells out that we need to add the function to `general_functions.rst`?

We'll see what else the CI system says needs to be modified; it's there to maintain code quality in the library, so it's kind of picky, please be patient with it!

CHANGELOG.rst
```diff
@@ -21,6 +21,7 @@ v0.20.8
 - [ENH] Added ``fill_direction`` function for forward/backward fills on missing values
   for selected columns in a dataframe. @samukweku
 - [ENH] Simpler logic and less lines of code for expand_grid function @samukweku
+- [ENH] Added function ``groupby_topk`` to janitor functions @mphirke
```
ericmjl: This should go in the "new version on deck" section. Version 0.20.8 was just released.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    sort_column_name: Hashable,
+    k: int,
+    sort_values_kwargs: Dict = None,
+) -> pd.DataFrame:
+    """
+    Return a dataframe that has the top `k` values grouped by
+    `groupby_column_name` and sorted by `sort_column_name`. Additional
+    parameters to the sorting (such as ascending=True) can be passed using
+    `sort_values_kwargs`. List of all sort_values() parameters can be found at
+    https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html
```
ericmjl: Let's get this adhering to [semantic line breaks][sembr]!

[sembr]: https://sembr.org/

Here's my suggested change:

```suggestion
    Return top `k` rows from a groupby of a set of columns.
    
    Returns a dataframe that has the top `k` values grouped by `groupby_column_name`
    and sorted by `sort_column_name`.
    Additional parameters to the sorting (such as ascending=True)
    can be passed using `sort_values_kwargs`.
    
    List of all sort_values() parameters can be found here_.
    
    .. _here: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html
```

I did a few things in the suggestion:

1. Adhered to docstring conventions by making first line a single sentence.
2. Used ReStructured text URLs to reference the docs. (This is easier than copy/pasting the URL.) The documentation on how to insert a URL lives [here][rst_url].

[rst_url]: https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets

Finally, you might need a `# noqa` on the docstring because the lines _might_ extend beyond the 79 character limit. You'll see sufficient examples in `functions.py` that have this `# noqa` piece.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    :param groupby_column_name: Column name to group input dataframe `df` by.
+    :param sort_column_name: Name of the column to sort along the
+        input dataframe `df`.
+    :param k: Number of top rows to return from each group after sorting.
+    :param sort_values_kwargs: Arguments to be passed to sort_values function.
+    :returns: A pandas dataframe with top `k` rows that are grouped by
+        `groupby_column_name` column with each group sorted along the
+        column `sort_column_name`.
+    :raises: ValueError if `k` is less than 1.
+    :raises: ValueError if `groupb y_column_name` does not exist in dataframe.
```
```suggestion
    :raises: KeyError if `groupby_column_name` does not exist in dataframe.
```

I think it should be `KeyError`, rather than `ValueError`, as keys are how pandas access individual columns.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    :raises: ValueError if `sort_column_name` does not exist in dataframe.
```
ericmjl: Same as above.

```suggestion
    :raises: KeyError if `sort_column_name` does not exist in dataframe.
```

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    :raises: ValueError if `groupby_column_name` is same as `sort_column_name`.
```
```suggestion
    :raises: KeyError if `groupby_column_name` is same as `sort_column_name`.
```

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    if k < 1:
+        raise ValueError(
+            "Numbers of rows per group to be returned must be greater than 0."
+        )
+
+    # Check if groupby_column_name exists in dataframe.
+    if groupby_column_name not in df.columns:
+        raise ValueError(
+            f"Column {groupby_column_name} does not exist in the dataframe."
+        )
```
```suggestion
    if groupby_column_name not in df.columns:
        raise KeyError(
            f"Column {groupby_column_name} does not exist in the dataframe."
        )
```

Actually, I think we have a function that does this already.

It's in [`janitor.utils`](https://github.com/ericmjl/pyjanitor/blob/dev/janitor/utils.py#L317). And curiously, it raises a ValueError (really should be a KeyError).

In any case, can you use the `check_column` in place of the current checks that you have? This way we have fewer lines of code to maintain.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+
+    # Check if sort_column_name exists in dataframe.
+    if sort_column_name not in df.columns:
+        raise ValueError(
+            f"Column {sort_column_name} does not exist in the dataframe."
+        )
```
ericmjl: Just another comment to leave a visual reminder for you - change this one to use `check_column`.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    # Check if group by column and sort column are unique.
+    if sort_column_name == groupby_column_name:
+        raise ValueError("Group by column and sort column cannot be same.")
+
+    # Check if inplace:True in sort values kwargs because it returns None
+    if (
+        "inplace" in sort_values_kwargs.keys()
+        and sort_values_kwargs["inplace"]
+    ):
+        raise KeyError("Cannot use inplace : True in sort_values_kwargs.")
```
ericmjl: Minor change suggestion for the error message:

```suggestion
        raise KeyError("Cannot use `inplace=True` in `sort_values_kwargs`.")
```

scripts/check-autodoc.py
```diff
@@ -70,5 +70,7 @@ def extract_folder_names(test_dir: Path) -> Tuple[Path, str]:
 
 if len(missing_funcs) > 0:
     raise Exception(
-        f"The following functions have not yet been added: {missing_funcs}"
+        f'''The following functions have not yet been added: {missing_funcs}.
+         Please add the function names to docs/reference/general_functions.rst
+         under an appropriate section.'''
```
ericmjl: Black convention usually would suggest using the double quotes for docstrings:

```suggestion
        f"""The following functions have not yet been added: {missing_funcs}.
         Please add the function names to docs/reference/general_functions.rst
         under an appropriate section."""
```

ericmjl (changes requested):
@mphirke thank you for your work here! I found a chunk of time to do a PR review. This is high quality stuff you've got, thanks for taking the time here.

Most of the comments I have pertain to nit-picky stuff that come from my own knowledge of the rest of the codebase. Hope you don't mind them. 

I'm going to give the PR a few more days to get in others' reviews. Otherwise, I'll target Monday or Tuesday to merge the PR. 

Once again, great work!

samukweku (dismissed):
Hi @mphirke,  I will go through them now.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+
+    return df.groupby(groupby_column_name).apply(
```
samukweku, line 4387: Great work @mphirke !

 I suggest not using ``apply`` for this and use the built-in functions - apply is slower, about 2x slower when I ran a test on your example, and would likely decrease in speed as the data size increases. One possible solution is : 

    df.sort_values([group_column_name, sort_column_name], ascending = [False, True]).groupby(group_column_name, sort=False).head(k)

The `ascending` parameter can be modified dynamically, depending on  the ``sort_value_kwargs``

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+        )
+
+    # Check if groupby_column_name exists in dataframe.
+    if groupby_column_name not in df.columns:
+        raise ValueError(
+            f"Column {groupby_column_name} does not exist in the dataframe."
+        )
+
+    # Check if sort_column_name exists in dataframe.
+    if sort_column_name not in df.columns:
```
samukweku: I suggest removing this if condition - Pandas would trigger the error anyway, so let Pandas worry about that and include a test for it. I believe it will reduce the number of lines of code, and maintenance as well. (This is prompting me to revisit some of my own code and refactor!). In the test, it should be KeyError, as @ericmjl suggested.

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+    sort_values_kwargs = sort_values_kwargs or {}
+
+    # Check if k is greater than 0.
+    if k < 1:
+        raise ValueError(
+            "Numbers of rows per group to be returned must be greater than 0."
+        )
+
+    # Check if groupby_column_name exists in dataframe.
+    if groupby_column_name not in df.columns:
```
samukweku: Pandas will raise a KeyError if column does not exist, as such, I think we can safely remove this condition. I suggest a test is created for this.

tests/functions/test_groupby_topk.py
```diff
@@ -0,0 +1,67 @@
...
+    """Test descending group by, k=3"""
+    expected = df.groupby("result").apply(
+        lambda d: d.sort_values("age", ascending=False).head(3)
+    )
+    assert_frame_equal(
+        df.groupby_topk("result", "age", 3, {"ascending": False}), expected
+    )
+
+
+def test_wrong_groupby_column_name(df):
```
samukweku, line 38: It should be a KeyError exception, if we move the check to Pandas

tests/functions/test_groupby_topk.py
```diff
@@ -0,0 +1,67 @@
...
+    """Raise Value Error if wrong groupby column name is provided."""
+    with pytest.raises(ValueError):
+        df.groupby_topk("RESULT", "age", 3)
+
+
+def test_wrong_sort_column_name(df):
```
samukweku, line 44: This also should be a KeyError exception, if we allow Pandas to do the checks

janitor/functions.py
```diff
@@ -4245,3 +4245,133 @@ def fill_direction(
...
+        )
+
+    # Check if sort_column_name exists in dataframe.
+    if sort_column_name not in df.columns:
+        raise ValueError(
+            f"Column {sort_column_name} does not exist in the dataframe."
+        )
+
+    # Check if group by column and sort column are unique.
+    if sort_column_name == groupby_column_name:
```
samukweku: If you dont mind me asking, what is the danger in sorting and grouping on the same column?

samukweku (changes requested):
Great work @mphirke ! Let me know if any my comments or suggestions come across in a wrong/unpleasant way.

janitor/functions.py
mphirke: Hi, I did as requested and added # noqa E501. Hadn't heard of noqa and semantic line breaks before so this was very educational.

janitor/functions.py
mphirke: Hi, thanks for review. The method you mentioned is indeed faster, but it does not return the dataframe as grouped. For returning the dataframe as grouped, I believe the code will have to be modified to:

`df.sort_values([groupby_column_name,sort_column_name],**sort_values_kwargs).groupby(groupby_column_name).apply(lambda d: d.head(k))`

which requires application of apply and is slower on the datasets that I tested (Iris for small dataset and [2020 January NYC Yellow Taxis trip record dataset](https://www1.nyc.gov/site/tlc/about/tlc-trip-record-data.page) as medium sized dataset (550mb+).

[This is the notebook]( https://colab.research.google.com/drive/1o9_MMXM6JEhfZI4a2c_ZSRyBvMyA_3Qn?usp=sharing) where I've compare the three methods (pull request, suggestion and suggestion with `.apply(lambda d: d.head(k))` to return grouped dataframe.

janitor/functions.py
mphirke: I've done as you suggested.

janitor/functions.py
mphirke: Done as suggested.

janitor/functions.py
mphirke: There is no danger, but I initially thought it would be silly to do that since that would just return return a grouped dataframe sorted by the group...but it does seem like it could be useful in some cases now I think about it.

I think I should remove that check, then. Would you suggest the same?

scripts/check-autodoc.py
mphirke: I've replaced the single quotes by double quotes. It seems the auto styling does not run over the scripts so I missed it.

janitor/functions.py
mphirke: Changed.

janitor/functions.py
mphirke: As @samukweku suggested in his review https://github.com/ericmjl/pyjanitor/pull/714#discussion_r467526966, pandas gives a KeyError by itself when column is not found, so I removed these checks and added tests for checking for KeyError when wrong column is provided.

janitor/functions.py
mphirke: I've done as As @samukweku suggested in his review #714 (comment), pandas gives a KeyError by itself when column is not found, so I removed these checks and added tests for checking for KeyError when wrong column is provided.

Should I change it to use check_column and add the two KeyErrors or keep it the way it is now and let pandas give the KeyError? Asking since the reviews regarding these checks are different.

tests/functions/test_groupby_topk.py
mphirke: Changed to KeyError.

tests/functions/test_groupby_topk.py
mphirke: Changed to KeyError.

CHANGELOG.rst
mphirke: I changed this but there is a conflict now. I'll fix it in the next commit.

janitor/functions.py
mphirke: Missed this somehow. Will get it in the next commit soon along with the Changelog fix.

janitor/functions.py
samukweku: Thanks for the notebook. I went through it and your tests pan out. I do have some suggestions : 
@ericmjl, your input here would help. Do we want the grouping column as part of the index, or reintegrated back into the dataframe as a column? If we stick to the grouping column staying as the dataframe's index, then, I dont think it is necessary to have the same grouping column remain as part of the columns when it is already the index. Also, we can simply set the index instead of using `apply` : ``df.sort_values([group_column_name, sort_column_name], ascending = [False, True]).groupby(group_column_name, sort=False).head(k).set_index(group_column_name)``.

janitor/functions.py
ericmjl: Now that I've had a bit of time away from the problem for a bit, I think we should change the ValueError in `check` to a KeyError so that it matches what `pandas` gives us. Consistent error message types help simplify error checking, though our `check` function explicitly states the problem. For now, I'd suggest making everything consistent with using `check`; in a separate PR, we can make the codebase consistently return `KeyError`s by changing the underlying implementation of `check`.

scripts/check-autodoc.py
ericmjl: No worries! Looks like that's something for me to take care of.

janitor/functions.py
ericmjl: @samukweku I think it's best to stick with respecting whatever the original index of the dataframe is. At least for me, I found this to be a sane starting point - if pyjanitor functions change the index, then that is probably better done with a flag, or else documented properly. For now, I'm leaning slightly more in favour of "changing the index with a flag" rather than "changing the index by default".

janitor/functions.py
ericmjl: Yes, I agree with you, @mphirke, let's remove the check. I'm in favour of a simpler starter implementation that does a sane thing, which we can always reason about later. It's harder to delete code.

CHANGELOG.rst
mphirke: Fixed this.

janitor/functions.py
mphirke: Removed this check entirely on suggestions so this change is not relevant.

janitor/functions.py
mphirke: Removed this check from both the function as well as tests.

janitor/functions.py
mphirke: I've updated the notebook to include the suggestion to simply use "set_index" and it is faster than using apply as @samukweku said but it changes index, so @ericmjl , I've kept the function as it is for now. I could integrate a flag (something like `quick = True`), but I don't know whether it is required here. Your help would be appreciated.

janitor/functions.py
mphirke: I believe I got what you mean, but apologies if I misunderstood (My English is not my strongest point).

I think what you want me to do is to remove the checks in function but keep the checks in tests as `KeyError` so they will be consistent with `check_column` later on? If so, I've done this.

janitor/functions.py
mphirke: It seems the Deepsource check is failing because of 2 antipatterns on lines that I have not modified at all which is a bit weird, but the errors seem simple (`unnecessary `dict` call - write as literal` seems it's just `dict()` has to be changed to `{}` so it's faster). Should I fix them in this PR itself or ignore them?

ericmjl: Great work, @mphirke! Here's my thoughts below:

> I could integrate a flag (something like quick = True), but I don't know whether it is required here. Your help would be appreciated.

I'm in favour of keeping the function simple on the first PR, and then waiting to see whether there's sufficient asks for a faster version. Doing so helps with long-term maintenance of the library. 

> It seems the Deepsource check is failing because of 2 antipatterns on lines that I have not modified at all which is a bit weird, but the errors seem simple (unnecessary dict call - write as literal seems it's just dict() has to be changed to {} so it's faster). Should I fix them in this PR itself or ignore them?

I tried to PR in directly, but it looks like I can't. Yes, let's just sneak these into this PR :).

janitor/functions.py
ericmjl: No worries, @mphirke - looking back on what I wrote, I realize I wasn't communicating clearly either.

I think what was going through my head back then was this: `janitor.utils.check_column` will check whether column names are present in the dataframe or not for us, and returns a more useful error message than pandas' error message. As such, my proposal is:

1. Replace the lines in which you were checking for the presence of a column with the `check_column` function. 
2. Ignore the fact that `check_column` raises a ValueError. 
3. In the tests you have, for now, raise a ValueError rather than KeyError, because that's what `check_column` raises.
4. We can standardize the error type across the codebase back to KeyError in a separate PR.

janitor/functions.py
mphirke: Thanks for your help @ericmjl. That makes sense.

Fixed the DeepSource issue in the latest commit with the `check_column` change.

janitor/functions.py
mphirke: So it seems I understood it all wrong, my bad. Thanks a lot for clarifying. I've done as you suggested.

ericmjl (approved):
I'm good to go with the changes. Thanks @mphirke! 

@samukweku, would you like to do the honors? Also, it's worth a release at this point, I think, so that the distributed package matches up with the docs.

samukweku:
@ericmjl You want me to do the release? Love to! need your guidance on that though

ericmjl:
Yep! I'd love it if you would be willing to do the release!

The steps are documented on [the wiki]. I've done my best documenting each piece, but if you'd like to, we can walk through it over a pair call, with the goal being that you document whatever steps I've not written down. (I will insist on this call that you must insist firmly that I slow down if I am going too fast.)

You'll need a PyPI account. Can you email me your username so I can give you access to the package?

[the wiki]: https://github.com/ericmjl/pyjanitor/wiki/How-to-cut-a-new-release

mphirke:
Thanks for the guidance @ericmjl and @samukweku. Overall very educational and I'm looking forward to contribute some more in the future.

PR #718, opened 2020-08-12
[DOC] Maintenance

ericmjl:
# PR Description

- Fixed notebooks so that they work again. Their outputs have been cleared in preparation for switch over to mkdocs-material, which is frankly easier to maintain than sphinx.
- Some docstring indentation pieces were changed.
- Switched over to more explicit pandas indexing.

Requesting review from those who have touched the files I've touched.


samukweku:
@ericmjl 
I'll have a look at this

samukweku:
ok. How do I review notebook code? Tried using reviewNB, it just looks off. Plus I think a lot of the `str_remove`/`str_trim`, ... can be replaced with process_text. I'm not sure if it is possible to review it on VScode?

ericmjl:
@samukweku yep! I think you can clone the branch directly into a dev container!

samukweku:
@ericmjl , made some changes and left some comments on anime.ipynb. Have a look, I probably missed out some proper rewording

samukweku (approved):
LGTM. Need someone else to go through the notebooks

janitor/chemistry.py
```diff
@@ -114,17 +114,20 @@ def smiles2mol(
...
+                smiles_column_name='smiles',
+                mols_column_name='mols'
+            )
+        )
 
     A progressbar can be optionally used.
 
     - Pass in "notebook" to show a tqdm notebook progressbar. (ipywidgets must
-      be enabled with your Jupyter installation.)
+    be enabled with your Jupyter installation.)
```
hectormz, line 128: Is this change in indentation more proper/non-breaking? It currently renders correctly on the docs site. I never know with these things until I see it in action

ericmjl: It's a convention, mostly, nothing else. I haven't been strict about it, because there's no easy way to lint for it. But then again, I think there's a point to "if one can't lint for it, one shouldn't really bother"!

hectormz:
@ericmjl I have no other thoughts. But you should re-execute the two notebooks. The execution and other things has been messed up since your initial commit in this PR. I prepared a commit I was going to push, but my computer is upset with me at the moment

samukweku:
@hectormz , the mess might be my fault. @ericmjl , when you have a look, let me know where I bungled it, so I learn and avoid it in the future.

hectormz:
@samukweku initially I saw that the cells were executes out of order.  Not a big deal, but it's nice to have that clean on order. Also I'm not sure if you're using pre-commit. I think pre-commit is setup for the project to use nbstripout

samukweku:
Thanks @hectormz I use the dev container. How do I go about the pre commit hooks?

hectormz:
@samukweku I don't use the dev container so I'm not sure, but perhaps @ericmjl may know

ericmjl:
Hmm, looks like this is one thing to add to this maintenance PR - pre-commit hooks automatically enabled in the dev container!

samukweku:
@ericmjl the error seems to be from `test_convert_excel_date`. Since python version is 3.8, I suspect it should work. Unless there is something else. I triggered the build; it failed.

ericmjl:
This is old enough that I think we should close the PR.

PR #720, opened 2020-08-16
[ENH] Adding ability to sort timestamps monotonically

### LINKED ISSUE #707: Add the ability to test and sort timestamps to be monotonic in a pandas data frame
# Brief Description

Following up on #703, this issue seeks to introduce the ability to sort the timestamps in a pandas data frame monotonically

I would like to propose...

# Example API
```python
def _test_for_monotonicity(
        df: pd.DataFrame,
        column_name: str = None,
        direction: str = 'increasing'
) -> bool:
    """
    Tests input data frame for monotonicity.

    Check if the data is monotonically increasing or decreasing.
    Direction is dependent on user input.
    Defaults to increasing

    :param df: data frame to be tested for monotonicity
    :param column_name: needs to be specified if and only if the date time is not in index.
        Defaults to None.
    :param direction: specifies the direction in which monotonicity is being tested for.
        Defaults to 'increasing'
    :return: single boolean flag indicating whether the test has passed or not
    """

def sort_monotonically(
        df: pd.DataFrame,
        column_name: str = None,
        direction: str ='increasing'
) -> pd.DataFrame:
    """
    Sorts data frame monotonically.
    It assumes the data frame has an index of type pd.DateTimeIndex when index is datetime.
    If datetime is in a column, then the column is expected to be of type pd.Timestamp

    :param df: data frame to sort monotonically
    :param column_name: needs to be specified if and only if the date time is not in index.
        Defaults to None
    :param direction: specifies the direction in which monotonicity is being tested for.
        Defaults to 'increasing'
    :return: data frame with its index sorted
    """


# more examples below
# ...
```

UGuntupalli:
Closes #707 

# PR Description

- Adding a function that allows monotonic sorting of timestamps in data frame 

**This PR resolves #707 **

# PR Checklist

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers
- @ericmjl 
- @samukweku 

- @ericmjl

### DIFF
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -92,16 +92,19 @@ def _get_missing_timestamps(
 @pf.register_dataframe_method
 def sort_timestamps_monotonically(
     df: pd.DataFrame,
-    direction: str = "increasing"
+    direction: str = "increasing",
+    strict: bool = False
 ) -> pd.DataFrame:
     """
-    Sort dataframe to be monotonic.
+    Sort dataframe such that index is monotonic.
 
     If timestamps are monotonic,
     this function will return the dataframe unmodified.
     If timestamps are not monotonic,
     then the function will sort the dataframe.
+
     Example usage:
+
     .. code-block:: python
 
         df = (
@@ -114,11 +117,24 @@ def sort_timestamps_monotonically(
         Acceptable arguments are:
             1. increasing
             2. decreasing
-    :returns: dataframe that has monotonic timestamps.
+    :param strict: flag to enable/disable strict monotonicity.
+        If set to True,
+        will remove duplicates in the index,
+        by retaining first occurrence of value in index.
+        If set to False,
+        will not test for duplicates in the index.
+        Defaults to False.
+    :returns: Dataframe that has monotonically increasing
+        (or decreasing) timestamps.
     """
     # Check all the inputs are the correct data type
     check("df", df, [pd.DataFrame])
     check("direction", direction, [str])
+    check("strict", strict, [bool])
+
+    # Remove duplicates if requested
+    if strict:
+        df = df[~df.index.duplicated(keep='first')]
 
     # Sort timestamps
     if direction == "increasing":
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -87,3 +87,44 @@ def _get_missing_timestamps(
     missing_timestamps = expected_df.index.difference(df.index)
 
     return expected_df.loc[missing_timestamps]
+
+
+@pf.register_dataframe_method
+def sort_timestamps_monotonically(
+    df: pd.DataFrame,
+    direction: str = "increasing"
+) -> pd.DataFrame:
+    """
+    Sort dataframe to be monotonic.
+
+    If timestamps are monotonic,
+    this function will return the dataframe unmodified.
+    If timestamps are not monotonic,
+    then the function will sort the dataframe.
+    Example usage:
+    .. code-block:: python
+
+        df = (
+            pd.DataFrame(...)
+            .sort_timestamps_monotonically(direction='increasing')
+        )
+
+    :param df: Dataframe which needs to be tested for monotonicity
+    :param direction: type of monotonicity desired.
+        Acceptable arguments are:
+            1. increasing
+            2. decreasing
+    :returns: dataframe that has monotonic timestamps.
+    """
+    # Check all the inputs are the correct data type
+    check("df", df, [pd.DataFrame])
+    check("direction", direction, [str])
+
+    # Sort timestamps
+    if direction == "increasing":
+        df = df.sort_index()
+    else:
+        df = df.sort_index(ascending=False)
+
+    # Return the dataframe
+    return df
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -91,9 +91,7 @@ def _get_missing_timestamps(
 
 @pf.register_dataframe_method
 def sort_timestamps_monotonically(
-    df: pd.DataFrame,
-    direction: str = "increasing",
-    strict: bool = False
+    df: pd.DataFrame, direction: str = "increasing", strict: bool = False
 ) -> pd.DataFrame:
     """
     Sort dataframe such that index is monotonic.
@@ -134,7 +132,7 @@ def sort_timestamps_monotonically(
 
     # Remove duplicates if requested
     if strict:
-        df = df[~df.index.duplicated(keep='first')]
+        df = df[~df.index.duplicated(keep="first")]
 
     # Sort timestamps
     if direction == "increasing":
--- a/tests/timeseries/test_sort_timestamps_monotonically.py
+++ b/tests/timeseries/test_sort_timestamps_monotonically.py
@@ -1,7 +1,7 @@
 import pytest
 import pandas as pd
 from random import randint
-import janitor
+import janitor  # noqa: F401
 import janitor.timeseries  # noqa: F401
 
 
--- a/tests/timeseries/test_sort_timestamps_monotonically.py
+++ b/tests/timeseries/test_sort_timestamps_monotonically.py
@@ -26,7 +26,7 @@ def test_sort_timestamps_monotonically(timeseries_dataframe):
 
 
 @pytest.mark.timeseries
-def test_sort_timestamps_monotonically(timeseries_dataframe):
+def test_sort_timestamps_monotonically_decreasing(timeseries_dataframe):
     """Test sort_timestamps_monotonically for descending order"""
     df2 = timeseries_dataframe.sort_index(ascending=False)
     df3 = sort_timestamps_monotonically(df2, "decreasing")
@@ -41,4 +41,3 @@ def test_sort_timestamps_monotonically_strict(timeseries_dataframe):
     df4 = df.append(df.loc[df.index[random_number], :])
     df5 = sort_timestamps_monotonically(df4, "increasing", True)
     pd.testing.assert_frame_equal(df5, timeseries_dataframe)
-
--- a/tests/timeseries/test_sort_timestamps_monotonically.py
+++ b/tests/timeseries/test_sort_timestamps_monotonically.py
@@ -30,3 +30,9 @@ def test_sort_timestamps_monotonically(timeseries_dataframe):
     df2 = timeseries_dataframe.sort_index(ascending=False)
     df3 = sort_timestamps_monotonically(df2, "decreasing")
     assert df3.equals(df2)
+
+    # Test for strictness
+    random_number = randint(1, len(timeseries_dataframe))
+    df4 = df.append(df.loc[df.index[random_number], :])
+    df5 = sort_timestamps_monotonically(df4, "increasing", True)
+    assert df5.equals(timeseries_dataframe)
--- a/tests/timeseries/test_sort_timestamps_monotonically.py
+++ b/tests/timeseries/test_sort_timestamps_monotonically.py
@@ -1,8 +1,8 @@
 import pytest
 import pandas as pd
 from random import randint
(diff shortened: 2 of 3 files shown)


janitor/timeseries.py
```diff
@@ -87,3 +87,44 @@ def _get_missing_timestamps(
...
+) -> pd.DataFrame:
+    """
+    Sort dataframe to be monotonic.
+
+    If timestamps are monotonic,
+    this function will return the dataframe unmodified.
+    If timestamps are not monotonic,
+    then the function will sort the dataframe.
+    Example usage:
+    .. code-block:: python
```
ericmjl: We might need a few blank lines here for the docs to render properly.

```suggestion
    then the function will sort the dataframe.

    Example usage:

    .. code-block:: python
```

janitor/timeseries.py
```diff
@@ -87,3 +87,44 @@ def _get_missing_timestamps(
...
+            pd.DataFrame(...)
+            .sort_timestamps_monotonically(direction='increasing')
+        )
+
+    :param df: Dataframe which needs to be tested for monotonicity
+    :param direction: type of monotonicity desired.
+        Acceptable arguments are:
+            1. increasing
+            2. decreasing
+    :returns: dataframe that has monotonic timestamps.
```
ericmjl: Just a minor phrasing suggestion:

```suggestion
    :returns: Dataframe that has monotonically increasing (or decreasing) timestamps.
```

janitor/timeseries.py
```diff
@@ -87,3 +87,44 @@ def _get_missing_timestamps(
...
     return expected_df.loc[missing_timestamps]
+
+
+@pf.register_dataframe_method
+def sort_timestamps_monotonically(
+    df: pd.DataFrame,
+    direction: str = "increasing"
+) -> pd.DataFrame:
+    """
+    Sort dataframe to be monotonic.
```
ericmjl:
```suggestion
    Sort dataframe such that index is monotonic.
```

ericmjl (changes requested):
@UGuntupalli thanks for the work here! 

I am requesting some minor changes, mostly pertaining to the docs.

One thing that I'm wondering is this: is there something that could be value-added to an end-user, over a simple `df.sort_index()`? For example, could it be instead `sort_strictly_monotonically()` instead, such that there's a check for no duplicate index values baked into the function? What do you think?

If this is a simple wrapper around `df.sort_index()`, then perhaps the function should be described as "super sugary syntax" for `df.sort_index()`, not unlike some other syntactic sugar functions that we have.

janitor/timeseries.py
UGuntupalli: I am wondering if doing this would introduce a flake error, but I will give it a shot

UGuntupalli:
> @UGuntupalli thanks for the work here!
> 
> I am requesting some minor changes, mostly pertaining to the docs.
> 
> One thing that I'm wondering is this: is there something that could be value-added to an end-user, over a simple `df.sort_index()`? For example, could it be instead `sort_strictly_monotonically()` instead, such that there's a check for no duplicate index values baked into the function? What do you think?
> 
> If this is a simple wrapper around `df.sort_index()`, then perhaps the function should be described as "super sugary syntax" for `df.sort_index()`, not unlike some other syntactic sugar functions that we have.

Would the function then be doing just one thing like we wanted ? I am happy to add more functionality as recommended, but I am wondering if that violates the principle of having a function do one and only thing ?

ericmjl:
> Would the function then be doing just one thing like we wanted ? I am happy to add more functionality as recommended, but I am wondering if that violates the principle of having a function do one and only thing ?

That sounds like a familiar lesson! :smile: Yes, that lesson definitely applies here. There's a matter of scope/context, as well as level of abstraction, in deciding how big or small of a scope to make a function. No easy answers, though I can give a behind-the-scenes look at my thought process:

Firstly, sugary-syntax is cool! Just has to be documented as such in the docstrings.

Secondly, if we go with non-sugary-syntax, then the _semantics_ of the function define what "one thing" is. The semantics are what we get to define, and here I lean towards the "simplest non-trivial thing" as the guiding light. 

In the suggestion for `sort_strictly_monotonically()`, the "strictly" implies that it there can be no duplicate values in the index, so if the function _checks_ for non-strict monotonicity and errors if there's no strict monotonicity, and otherwise just sorts by index.

An alternative can be to `sort_monotonically(strict=True)`, which then gives a user-configurable option to enable/disable strict monotonicity checking. This is also a valid way to handle sorting. This opens the door to other checks that can be gradually introduced, via keyword arguments.

What are your thoughts on this?

Because PRs take time to review, I'm in favor of introducing functionality slowly, rather than at one shot. In my prior experience, introducing many things at the same time makes it very easy for the human to confuse semantics of things, which then makes the code confusing to read as well. Though it feels more inefficient to do things this way, the end-result is a saner process that allows time for ideas to crystallize.

janitor/timeseries.py
```diff
@@ -87,3 +87,60 @@ def _get_missing_timestamps(
...
+        Acceptable arguments are:
+            1. increasing
+            2. decreasing
+    :param strict: flag to enable/disable strict monotonicity.
+        If set to True,
+        will remove duplicates in the index,
+        by retaining first occurrence of value in index.
+        If set to False,
+        will not test for duplicates in the index.
+        Defaults to False.
```
ericmjl, line 124: Wonderful documentation here, @UGuntupalli! 😄

tests/timeseries/test_sort_timestamps_monotonically.py
```diff
@@ -0,0 +1,38 @@
...
+    ts_index = pd.date_range("1/1/2019", periods=1000, freq="1H")
+    v1 = [randint(1, 2000) for i in range(1000)]
+    test_df = pd.DataFrame({"v1": v1}, index=ts_index)
+    return test_df
+
+
+@pytest.mark.timeseries
+def test_sort_timestamps_monotonically(timeseries_dataframe):
+    """Test that sort_timestamps_monotonically works as expected."""
+    from sklearn.utils import shuffle
```
ericmjl: This shouldn't be necessary; pyjanitor has a `.shuffle()` class method that can be called directly.

tests/timeseries/test_sort_timestamps_monotonically.py
```diff
@@ -0,0 +1,38 @@
...
+    Returns a time series dataframe
+    """
+    ts_index = pd.date_range("1/1/2019", periods=1000, freq="1H")
+    v1 = [randint(1, 2000) for i in range(1000)]
+    test_df = pd.DataFrame({"v1": v1}, index=ts_index)
+    return test_df
+
+
+@pytest.mark.timeseries
+def test_sort_timestamps_monotonically(timeseries_dataframe):
```
ericmjl, line 24: Can this test be split out please? Yes, there's a bit more boilerplate involved, but if and when a test fails, having them split out makes it easier to identify which exact test failed. Basically, each assert would go to its own function. 

Some proposed function names:

- `def test_sort_timestamps_monotonically`
- `def test_sort_timestamps_monotonically_decreasing`
- `def test_sort_timestamps_monotonically_strict`

tests/timeseries/test_sort_timestamps_monotonically.py
```diff
@@ -0,0 +1,38 @@
...
+    """Test that sort_timestamps_monotonically works as expected."""
+    from sklearn.utils import shuffle
+
+    # Increasing direction
+    df = shuffle(timeseries_dataframe)
+    df1 = sort_timestamps_monotonically(df)
+    assert df1.equals(timeseries_dataframe)
```
ericmjl: I forget if the `assert` pattern is used in the other PR you worked on, but I think across the `pyjanitor` codebase, we have used `pd.testing.assert_frame_equal(df1, df2)` instead of `assert df1.equals(df2)`. Can you check to see what the pattern is used elsewhere, and follow that for consistency please?

tests/timeseries/test_sort_timestamps_monotonically.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Returns a time series dataframe
+    """
+    ts_index = pd.date_range("1/1/2019", periods=1000, freq="1H")
+    v1 = [randint(1, 2000) for i in range(1000)]
+    test_df = pd.DataFrame({"v1": v1}, index=ts_index)
+    return test_df
+
+
+@pytest.mark.timeseries
```
ericmjl, line 23: I have this idea that the tests here can be parametrized, but I don't want to burden you further with this on this PR, as it does involve a bit more effort learning how to write parametrized tests (unless you're game for it!). Meanwhile, just leaving a suggestion here for documentation:

```suggestion
# NOTE: The tests possibly can be merged back together later if they are parametrized properly.

@pytest.mark.timeseries
```

ericmjl (changes requested):
Some further changes requested (this is the amount that I can catch before my attention has to be turned back to my day job). Thanks for your effort here, @UGuntupalli! 

Any others willing to chime in here?

tests/timeseries/test_sort_timestamps_monotonically.py
UGuntupalli: Always happy to learn

UGuntupalli:
These test errors don't occur on my local build. Locally all tests are passing, could you check on why that would happen ?

tests/timeseries/test_sort_timestamps_monotonically.py
ericmjl: Cool stuff! I'm going to make another issue for this and then tag you on it, so that there's something we can use to remember! I'm hoping to show @samukweku the ropes around doing a release this weekend, so I'm hoping we can get your two timeseries functions into the next release first.

Meanwhile, I'm going to go ahead and commit this suggestion directly.

ericmjl:
> These test errors don't occur on my local build. Locally all tests are passing, could you check on why that would happen ?

That's very weird indeed! I also cloned your repo locally, and in the development container environment, everything passed properly. I took the liberty of committing in a suggested modification, so let's see what happens now.

samukweku (changes requested):
The errors are coming up because of the import of `janitor` and `janitor.timeseries`, which are not used in the tests. Also, maybe add one more space between the imports and `#Random data for testing ` line, to take care of the flake8 error. Beyond that, I think the code looks fine, and understandable.

UGuntupalli:
@ericmjl, 
          Anything else needed to close this out ? I am planning to work on one more issue and just realized that this was still open

samukweku:
@UGuntupalli I think there is still some deliberation on whether the function is necessary. @ericmjl @szuckerman

ericmjl:
@samukweku I think I'm good with merging the function. One of the intents of `pyjanitor` is to provide a nice library of convenience functions (e.g. higher order functions that put two or three commonly-run steps together).

The thing I'm hung up on with this PR is why the tests fail on the CI, but don't fail locally. It's baffling, and I'm not sure what's going on. In particular, looking at what it's failing on, `df.shuffle().sort_timestamps_monotonically()` apparently returns None, but that isn't the case locally. I don't fully understand that.

@UGuntupalli having one PR open shouldn't stop you from trying out another, independent thing. (Unless the next PR depends on this one?) Doing so, though, means you'd refrain from deleting your fork, and instead update the `dev` branch of your fork of pyjanitor using ericmjl/pyjanitor's `dev` branch as the source. (You can send a pull request from my `dev` to your `dev` to accomplish that.) Then, make a new branch, and finally submit a new PR from the new branch.

szuckerman:
I think the method is also fine to include since there's an aspect of `pyjanitor` that it should be open to github newcomers or people who are getting started in open source. I feel that the "default" should be to include vs exclude.

PR #726, opened 2020-08-28
[ENH] Add range_flag function

### LINKED ISSUE #708: Add the ability to test whether data is in range in a pandas data frame
# Brief Description

Following up on #703, this issue seeks to introduce the ability to test whether each column in a data frame is within a user-defined range of values. 

I would like to propose..

# Example API

```python
def flag_values_not_in_range(
        df: pd.DataFrame,
        bound: list,
        inclusive: bool = True
) -> pd.DataFrame:
    """
    :param df: data frame to test if values in columns are within a range
    :param bound: user-defined lower and upper bound of values to test in data frame columns
    :param inclusive: include boundaries in filtering.
        Defaults to True
    :return: True if value in column does not fall within range.
        False if value in column falls within range.
    """

```

UGuntupalli:
#708  

PR Description

- Added a function that flags whether data in columns are within a user defined range.

**This PR resolves #708 .**

## Quick Check

- [] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [] Build the docs locally.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -87,3 +87,54 @@ def _get_missing_timestamps(
     missing_timestamps = expected_df.index.difference(df.index)
 
     return expected_df.loc[missing_timestamps]
+
+
+@pf.register_dataframe_method
+def range_flag(
+    df: pd.DataFrame,
+    bound: list,
+    inclusive: bool = True
+):
+    """
+    Test each column's values against boundaries.
+    If values in a column,
+    are within bound,
+    this function will return True.
+    If values in a column,
+    are not within bound,
+    this function will return False.
+
+    Example usage:
+    .. code-block:: python
+
+        df = (
+            pd.DataFrame(...)
+            .flag_values_not_in_range(bound=[-3, 5])
+        )
+
+    :param df: dataframe to test for values in range
+    :param bound: left and right boundary
+    :param inclusive: flag to include boundaries
+    :returns: dataframe with booleans for each column
+    """
+    # Check the input data sets for data type
+    check("df", df, [pd.DataFrame])
+    check("bound", bound, [list])
+    check("inclusive", inclusive, [bool])
+
+    # Get names of the columns into a list
+    columns = df.columns
+
+    # Make a copy to avoid mutation of the source dataframe
+    df1 = pd.DataFrame(index=df.index)
+
+    # Loop through the columns and create new flags
+    for column in columns:
+        df1[column + '_range_flag'] = df[column].between(
+            left=bound[0],
+            right=bound[1],
+            inclusive=inclusive
+        )
+
+    # Return the result
+    return df1
--- a/tests/timeseries/test_range_flag.py
+++ b/tests/timeseries/test_range_flag.py
@@ -1,0 +1,70 @@
+import pandas as pd
+import pytest
+from random import randint
+from janitor.timeseries import range_flag
+
+
+# Random data for testing
+@pytest.fixture
+def timeseries_dataframe() -> pd.DataFrame:
+    """
+    Returns a time series dataframe
+    """
+    ts_index = pd.date_range("1/1/2019", periods=1000, freq="1H")
+    v1 = [randint(1, 2000) for i in range(1000)]
+    test_df = pd.DataFrame({"v1": v1}, index=ts_index)
+    return test_df
+
+
+@pytest.mark.timeseries
+def test_range_flag(timeseries_dataframe):
+    """ Test that range flag works as expected """
+    my_range_flag = range_flag(
+        df=timeseries_dataframe,
+        bound=[100, 750],
+        inclusive=False
+    )
+
+    # Subset using the booleans created
+    v1_subset_index = my_range_flag[
+        my_range_flag['v1_range_flag'] == 1
+    ].index
+    v1_subset = timeseries_dataframe.loc[v1_subset_index, 'v1']
+
+    # Test if bounds have been honored
+    v1_subset_minimum = v1_subset.min()
+    v1_subset_maximum = v1_subset.max()
+    assert(100 < v1_subset_minimum)
+    assert(750 > v1_subset_maximum)
+
+
+@pytest.mark.timeseries
+def test_range_flag_inclusive(timeseries_dataframe):
+    """ Test range flag with inclusive argument """
+    # Set some values equal to boundary values
+    random_low = randint(1, len(timeseries_dataframe))
+    random_high = randint(1, len(timeseries_dataframe))
+    timeseries_dataframe.loc[timeseries_dataframe.index[random_low]] = 100
+    timeseries_dataframe.loc[timeseries_dataframe.index[random_high]] = 750
+
+    my_range_flag = range_flag(
+        df=timeseries_dataframe,
+        bound=[100, 750],
+        inclusive=True
+    )
+
+    # Subset using the booleans created
+    v1_subset_index = my_range_flag[
+        my_range_flag['v1_range_flag'] == 1
+    ].index
+    v1_subset = timeseries_dataframe.loc[v1_subset_index, 'v1']
+
+    # Test if bounds have been honored
+    v1_subset_minimum = v1_subset.min()
+    v1_subset_maximum = v1_subset.max()
+    assert(100 == v1_subset_minimum)
+    assert(750 == v1_subset_maximum)
+
+
+if __name__ == '__main__':
+    pytest.main()
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,6 @@
 new version (on deck)
 =====================
+- [ENH] Added function ``range_flag`` to timeseries. @UGuntupalli
 
 
 v0.20.9


UGuntupalli:
One thing I am thinking as I make the submission is if it would be more beneficial to make a ```bound``` a dictionary of lists, so a different set of limits can be passed for each column -  Just a thought, but did not want to over complicate it

ericmjl:
@UGuntupalli thanks for the PR! As I looked through the code, I realized that this behaves as a "data validation function", rather than a "data cleaning function", and was wondering if this is better done as a `pandera` check, rather than a `pyjanitor` function, maybe? For example, if you have a dataframe for which you need to _validate_ that the values are within a given range, a possibility is as follows:

```python
import pandera as pa 

schema = pa.DataFrameSchema(
    columns={
        "column_to_check": pa.Column(pa.DateTime, checks=pa.Check.in_range(...),
    }
)
```

And then you would simply flag error using:

```python
schema.validate(df, lazy=True)
```

Immediately, an error would show up, _and_ `pandera` would also give you the exact rows that fail the check.

The use of `pandera` is something I definitely recommend for data validation, as it allows us to very easily declare up-front what we expect from our data, and catch the errors at runtime.

What are your thoughts here?

Apart from the core of the proposed function possibly being better done in a different package, I must say, your PR was quite well done, and I'm super happy to see the improvement! :smile: :heart:

janitor/timeseries.py
```diff
@@ -87,3 +87,54 @@ def _get_missing_timestamps(
...
+    :param inclusive: flag to include boundaries
+    :returns: dataframe with booleans for each column
+    """
+    # Check the input data sets for data type
+    check("df", df, [pd.DataFrame])
+    check("bound", bound, [list])
+    check("inclusive", inclusive, [bool])
+
+    # Get names of the columns into a list
+    columns = df.columns
```
samukweku, line 126: You can reduce the number of lines and avoid the for loop by using pandas transform, then assigning the result to the new dataframe : 
```python
df1 = df.transform(pd.Series.between, axis=0, left=bound[0], right=bound[-1])
```
You'll have to test the speed though; this however is a built-in pandas function and negates the need for the for loop and the temporary variables for columns.

Also, there is a [between_time](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.between_time.html#:~:text=between_time,-DataFrame.&text=Select%20values%20between%20particular%20times,not%20between%20the%20two%20times.) function as well, with left and right bounds. It could possibly be a better option, since it is dedicated to time manipulations exclusively.

samukweku (changes requested):

janitor/timeseries.py
UGuntupalli: @samukweku , 
           I like the idea of keeping it native to Pandas, so I am happy to make this modification, remove the loop and replace it with ```transform()```. The ```between_time``` is only to handle datetimes correct- in my case the data can be anything and not necessarily datetimes. 

@ericmjl,
          Just to make sure I understand your suggestion correctly: 

1. Are you suggesting implementing the function in ```pandera``` package ? 
2. Are you suggesting re-writing the function using ```pandera``` in ```pyjanitor```? 

If it is (2), I believe ```pyjanitor```development environment does not have ```pandera```included. If it is (1), I can get started trying to get familiar with the process of contributing to ```pandera``` . One issue I see with (1) is, we don't necessarily need an error to be generated, rather just generate a mask of booleans, I am not sure if that is possible with the ```check``` construct in ```pandera```. This is something I will have to check

ericmjl: @UGuntupalli I took a few days to think more carefully about this PR, and I think I was previously getting confused by the PR name. "Testing" whether a column's values are inside a pre-defined range is a different activity from "flagging" whether that column's values are inside a pre-defined range. The difference lies in what actions are taken: a test should error out if the dataframe column fails a check, while a flag will add a column that contains the flag. Tests are out-of-scope for `pyjanitor`, because there's better tooling for this (i.e. `pandera`), but flagging is not, because we already have precedent (`flag_nulls`). Imprecise semantics were what were confusing me before.

The function name here contains `flag`, which puts it under the flagging category of things. I've renamed the PR and edited the description accordingly. Let's proceed with the PR; you can also ignore my comment about `pandera`, which I typed under the assumption that the intent was to put in a data testing function. (Data tests should be contributed to `pandera` if they aren't available in there already.)

janitor/timeseries.py
```diff
@@ -87,3 +87,54 @@ def _get_missing_timestamps(
     missing_timestamps = expected_df.index.difference(df.index)
 
     return expected_df.loc[missing_timestamps]
+
+
+@pf.register_dataframe_method
+def range_flag(
```
ericmjl, line 93: The behaviour of this function seems to be to return a completely different dataframe from the original, rather than appending new columns to the original.

This deviates from the behaviour of _most_ of the functions in `pyjanitor`, and I think the behaviour has to be properly documented in the docstring.

Question for you, @UGuntupalli - is the intent here to add columns of flags to the dataframe, or to return a completely new dataframe that has the flags, i.e. with the behaviour of `pandas`-native `df.transform()`?

I'd prefer that the behaviour would match `flag_nulls`, by adding a column of booleans to the dataframe, rather than following `df.transform()`, because this would make the behaviour of `flag_` functions consistent across the library. (I'd also rename the function to `flag_in_range`.) If you want to make the function use the latter behaviour (i.e. new df with boolean flags), I'd name the function differently so that the function-name-to-behaviour consistency is maintained. My naming proposals in the latter case would be:

- `isin_range(...)` - has the same behaviour as `df["column"].isin(some_list)`. I would favour this name, as it conveys the same behaviour that other users would expect.
- `highlight_in_range(...)` - thus kickstarting a class of functions called `highlight_X`

Sorry if it feels like I'm being a bit picky with the naming, but as the library grows, new additions have to fit into the right category of things in order to ensure that the behaviour of the library is consistent.

janitor/timeseries.py
```diff
@@ -87,3 +87,54 @@ def _get_missing_timestamps(
...
+    check("inclusive", inclusive, [bool])
+
+    # Get names of the columns into a list
+    columns = df.columns
+
+    # Make a copy to avoid mutation of the source dataframe
+    df1 = pd.DataFrame(index=df.index)
+
+    # Loop through the columns and create new flags
+    for column in columns:
```
ericmjl, line 132: I'd recommend using the `df.assign` pattern here:

```python
columns_data = {
    f"{column}_range_flag": df[column].between(left_bound=bound[0], right_bound=bound[1], inclusive=inclusive]
    for column
    in df.columns
}
df = df.assign(**columns_data)
```

This seems to be the more `pandas`-native way to assign new columns to the dataframe. The older selector-style syntax modifies the original. I think it's an opportunity here to encourage other readers of the source code to use this style; this is a change I want to make across the library.

tests/timeseries/test_range_flag.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@pytest.mark.timeseries
+def test_range_flag_inclusive(timeseries_dataframe):
+    """ Test range flag with inclusive argument """
+    # Set some values equal to boundary values
+    random_low = randint(1, len(timeseries_dataframe))
+    random_high = randint(1, len(timeseries_dataframe))
+    timeseries_dataframe.loc[timeseries_dataframe.index[random_low]] = 100
+    timeseries_dataframe.loc[timeseries_dataframe.index[random_high]] = 750
```
ericmjl, line 48: I see the magic number 100 and 750 show up quite a bit. It'd be helpful for further test development later if you could assign them to a variable inside the function and use the variables instead. That way, if someone else has the bandwidth and wants to parametrize the test, they'd be able to do so.

tests/timeseries/test_range_flag.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Test if bounds have been honored
+    v1_subset_minimum = v1_subset.min()
+    v1_subset_maximum = v1_subset.max()
+    assert(100 == v1_subset_minimum)
+    assert(750 == v1_subset_maximum)
+
+
+if __name__ == '__main__':
+    pytest.main()
```
ericmjl, line 70: I think we don't use this pattern elsewhere in the test suite. Where did you read about this?

ericmjl (changes requested):

ericmjl:
Hey @UGuntupalli! Just wanted to check in, anything I can do to help move this PR along? No pressure here, I'm working under the assumption that you've got your own open source projects that you'd like to get kickstarted too :).

UGuntupalli:
@ericmjl , 
            Just slammed both personally and professionally. It might be about a week or so before I am able to pick this up again unfortunately :(. I definitely want to finish this and contribute more inspite of the other open source project we discussed, but am a little behind with things. If anybody is able to adopt the changes suggested in the PR review - then great, else I will finish it in a weeks time

samukweku:
@UGuntupalli, just leaving a note here that using pd.Interval may be faster/more efficient

ericmjl:
Going to close this PR because it is quite outdated and the codebase has moved on. Leaving for the record that I do value the contributions regardless!

PR #728, opened 2020-08-29
[ENH] Add sort_timestamps_monotonically

### LINKED ISSUE #707: Add the ability to test and sort timestamps to be monotonic in a pandas data frame

ericmjl:
# PR Description

This PR adds back the `sort_timestamps_monotonically` functionality from PR #720. In that PR, the source branch was removed, so to salvage the PR, I manually added back the functions in this PR.

To fix the failing tests on #720, I also reverted the `pd.assert_frame_equal` to `assert df.equals(other_df)`. Better this than letting the PR linger in limbo.

**This PR resolves #707 and closes #720.**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,6 @@
 new version (on deck)
 =====================
-
+- [ENH] Added function ``sort_timestamps_monotonically`` to timeseries functions @UGuntupalli
 
 v0.20.9
 =======
--- a/janitor/timeseries.py
+++ b/janitor/timeseries.py
@@ -91,17 +91,18 @@ def _get_missing_timestamps(
 
 @pf.register_dataframe_method
 def sort_timestamps_monotonically(
-    df: pd.DataFrame,
-    direction: str = "increasing"
+    df: pd.DataFrame, direction: str = "increasing", strict: bool = False
 ) -> pd.DataFrame:
     """
-    Sort dataframe to be monotonic.
+    Sort dataframe such that index is monotonic.
 
     If timestamps are monotonic,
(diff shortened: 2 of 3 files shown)


ericmjl:
I'm going to merge this PR, as it was previously reviewed.

PR #735, opened 2020-09-07
[ENH] Add `also`

### LINKED ISSUE #731: Add an `also` method akin to Kotlins `also`
# Brief Description
I would like to propose a new function `also` that takes a function with a single dataframe for input, runs the function on the chained dataframe, and returns the input dataframe.  

I would love to take a whack at implementing this if you're interested incorporating it into the library.

As I've been trying to adopt chaining as much as possible, I've routinely bumped into scenarios where I need to break the chain for one reason or another. This API would eliminate most reasons for breaking the chain

This function is inspired by Kotlin's `also` function.  If we were to mimic Kotlin's `also`, the signature would look something like

`also: (f: T -> None) -> T` 

It would behave very similar to `pipe` or `then`, but can be used on functions that do not return anything. This can be very useful for any sort of side-effect operations within a long chain, such as periodic logging or debugging.

# Example API

Below are some examples of how this API would simplify code I've written in the past. 

### print the df while processing.
```python
# instead of defining an extra function
def print_and(df):
    print(df)
    return df

df.groupby(["column1"]).agg(sum).then(print_and).then(my_transform)

# or breaking the chain up
df_tmp = df.groupby(["column1"]).agg(sum)
print(df_tmp)
df_tmp.then(my_transform)

# you could just inline the print statement
df.groupby(["column1"]).agg(sum).also(print).then(my_transform)
```

### Insert a breakpoint while debugging, but keep going when you're done
Similar to above, to add a breakpoint for debugging, you would either have to create a new `pipe` compatible function or break the chain into smaller steps.

Instead, you could chain a breakpoint like so

```python
df.groupby(["column1"]).agg(sum).also(lambda x: breakpoint()).then(my_transform)
```

### Save a midpoint of a processing chain and then keep going

Another possibility would be running IO in the middle of the chain.  Often, a midpoint in the chain is just as important as the end result.  `also` would allow me to save the results and keep going without breaking a chain.

```
df.groupby(["column1"]).agg(sum).also(lambda x: x.to_csv("filename.csv)).then(my_transform)
```

### pipe-dream: Save a midpoint as a variable

I haven't figured out how to get this to work given Python's scoping rules, maybe the new `:=` operator will make this possible. 

This could be really helpful in scenarios where you're running groupbys or other aggregations and want to join those back into an ungrouped frame.

```
df.groupby(["column1"]).agg(sum).also(lambda x: df_mid = x).then(my_transform).assign(original_var = df_mid)
```

sauln:
# PR Description

- Adds `also` method as discussed in #731 
- Adds tests for `also`
- Adds docstring for `also`

**This PR resolves #731.**

# PR Checklist

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2964,6 +2964,25 @@ def then(df: pd.DataFrame, func: Callable) -> pd.DataFrame:
 
 
 @pf.register_dataframe_method
+def also(df: pd.DataFrame, func: Callable, *args, **kwargs) -> pd.DataFrame:
+    """Add an arbitrary function to run in the ``pyjanitor`` method chain that
+    has no return.
+
+    This method does not mutate the original DataFrame.
+
+    :param df: A pandas dataframe.
+    :param func: A function you would like to run in the method chain.
+        It should take one parameter and return one parameter, each being the
+        DataFrame object. After that, do whatever you want in the middle.
+        Go crazy.
+    :param args, kwargs: Arguments for parameters.
+    :returns: The input pandas DataFrame.
+    """
+    func(df.copy(), *args, **kwargs)
+    return df
+
+
+@pf.register_dataframe_method
 @deprecated_alias(column="column_name")
 def dropnotnull(df: pd.DataFrame, column_name: Hashable) -> pd.DataFrame:
     """Drop rows that do not have null values in the given column.
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2993,7 +2993,7 @@ def also(df: pd.DataFrame, func: Callable, *args, **kwargs) -> pd.DataFrame:
     :param func: A function you would like to run in the method chain.
         It should take one DataFrame object as a parameter and have no return.
         If there is a return, it will be ignored.
-    :param args, kwargs: Optional arguments and keywork arguments for `func`.
+    :param args, kwargs: Optional arguments and keyword arguments for `func`.
     :returns: The input pandas DataFrame.
     """
     func(df.copy(), *args, **kwargs)
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2965,10 +2965,27 @@ def then(df: pd.DataFrame, func: Callable) -> pd.DataFrame:
 
 @pf.register_dataframe_method
 def also(df: pd.DataFrame, func: Callable, *args, **kwargs) -> pd.DataFrame:
-    """Add an arbitrary function to run in the ``pyjanitor`` method chain that
-    has no return.
+    """Add an arbitrary function to run in the ``pyjanitor`` method chain.
+    Returns the input dataframe, not the output of `func`.
 
     This method does not mutate the original DataFrame.
+
+    Example usage:
+
+    .. code-block:: python
+        df = (
+            pd.DataFrame(...)
+            .query(...)
+            .also(lambda df: print(f"DataFrame shape is: {df.shape}"))
+            .transform_column(...)
+            .also(lambda df: df.to_csv("midpoint.csv"))
+            .also(
+                lambda df: print(
+                    f"Column col_name has these values: {set(df['col_name'].unique())}"
+                )
+            )
+            .group_add(...)
+        )
 
     :param df: A pandas dataframe.
     :param func: A function you would like to run in the method chain.
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2995,7 +2995,7 @@ def also(df: pd.DataFrame, func: Callable, *args, **kwargs) -> pd.DataFrame:
         If there is a return, it will be ignored.
     :param args, kwargs: Optional arguments and keyword arguments for `func`.
     :returns: The input pandas DataFrame.
-    """
+    """  # noqa: E501
     func(df.copy(), *args, **kwargs)
     return df
 
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2965,8 +2965,9 @@ def then(df: pd.DataFrame, func: Callable) -> pd.DataFrame:
 
 @pf.register_dataframe_method
 def also(df: pd.DataFrame, func: Callable, *args, **kwargs) -> pd.DataFrame:
-    """Add an arbitrary function to run in the ``pyjanitor`` method chain.
-    This returns the input dataframe instead, not the output of `func`.
+    """Add an arbitrary function with no return value to run in the
+    ``pyjanitor`` method chain. This returns the input dataframe instead,
+    not the output of `func`.
 
     This method does not mutate the original DataFrame.
 
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -2966,7 +2966,7 @@ def then(df: pd.DataFrame, func: Callable) -> pd.DataFrame:
 @pf.register_dataframe_method
 def also(df: pd.DataFrame, func: Callable, *args, **kwargs) -> pd.DataFrame:
     """Add an arbitrary function to run in the ``pyjanitor`` method chain.
-    Returns the input dataframe, not the output of `func`.
+    This returns the input dataframe instead, not the output of `func`.
 
     This method does not mutate the original DataFrame.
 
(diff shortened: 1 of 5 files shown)


sauln:
Locally, everything passed except the tests with Spark. Also, when I build the docs there is are a ton of changes. Do you want the new files in `docs/reference` committed?

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
+from unittest.mock import Mock
```
ericmjl, line 2: This is a new one for me, @sauln! Funnily enough, I've never mocked up anything in a test before. 

(Not a review, btw, just a comment.)

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+import pytest
+
+
+def remove_first_two_letters_from_col_names(df):
```
ericmjl, line 6:
```suggestion
def remove_first_two_letters_from_col_names(df):
    """Helper function to mutate dataframe by changing column names."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    col_names = df.columns
+    col_names = [name[2:] for name in col_names]
+    df.columns = col_names
+    return df
+
+
+def remove_rows_3_and_4(df):
```
ericmjl, line 14:
```suggestion
def remove_rows_3_and_4(df):
    """Helper function to mutate dataframe by removing rows."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    df = df.drop(3, axis=0)
+    df = df.drop(4, axis=0)
+    return df
+
+
+def drop_inplace(df):
+    df.drop(columns=[df.columns[0]], inplace=True)
```
ericmjl:
```suggestion
def drop_inplace(df):
    """
    Helper function to mutate dataframe by dropping a column.
    
    We usually would not use `inplace=True` in a block,
    but the intent here is to test that 
    the in-place modification of a dataframe
    doesn't get passed through in the `.also()` function.
    Hence, we tell Flake8 to skip checking `PD002` on that line.
    """
    df.drop(columns=[df.columns[0]], inplace=True)  # noqa: PD002
```

Added a noqa for `PD002`, as well as the intent of doing so.

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.mark.functions
+def test_also_column_manipulation_no_change(dataframe):
```
ericmjl, line 35:
```suggestion
def test_also_column_manipulation_no_change(dataframe):
    """Test that changed dataframe inside `.also()` doesn't get returned."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    cols = tuple(dataframe.columns)
+    df = dataframe.also(remove_first_two_letters_from_col_names)
+    assert dataframe is df
+    assert cols == tuple(df.columns)
+
+
+@pytest.mark.functions
+def test_also_remove_rows_no_change(dataframe):
```
ericmjl, line 44:
```suggestion
def test_also_remove_rows_no_change(dataframe):
    """Test that changed dataframe inside `.also()` doesn't get returned."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    df = dataframe.also(remove_rows_3_and_4)
+    rows = tuple(df.index)
+    assert rows == (0, 1, 2, 3, 4, 5, 6, 7, 8)
+
+
+@pytest.mark.functions
+def test_also_runs_function(dataframe):
```
ericmjl, line 52:
```suggestion
def test_also_runs_function(dataframe):
   """Test that `.also()` executes the function."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    method = Mock(return_value=None)
+    df = dataframe.also(method)
+    assert id(df) == id(dataframe)
+    assert method.call_count == 1
+
+
+@pytest.mark.functions
+def test_also_args(dataframe):
```
ericmjl, line 61:
```suggestion
def test_also_args(dataframe):
    """Test that the args are passed through to the function."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    method = Mock(return_value=None)
+    _ = dataframe.also(method, 5)
+    assert method.call_args.args[1] == 5
+
+
+@pytest.mark.functions
+def test_also_kargs(dataframe):
```
ericmjl:
```suggestion
def test_also_kwargs(dataframe):
    """Test that the kwargs are passed through to the function."""
```

tests/functions/test_also.py
```diff
@@ -0,0 +1,63 @@
...
+    method = Mock(return_value=None)
+    _ = dataframe.also(method, n=5)
+    assert method.call_args.kwargs == {'n': 5}
+
+
+@pytest.mark.functions
+def test_also_drop_inplace(dataframe):
```
ericmjl, line 79:
```suggestion
def test_also_drop_inplace(dataframe):
    """Test that in-place modification of dataframe does not pass through."""
```

tests/functions/test_also.py
ericmjl, line 2:
```suggestion
"""Unit tests for `.also()`."""
from unittest.mock import Mock
```

janitor/functions.py
```diff
@@ -2963,6 +2963,43 @@ def then(df: pd.DataFrame, func: Callable) -> pd.DataFrame:
...
+            .group_add(...)
+        )
+
+    :param df: A pandas dataframe.
+    :param func: A function you would like to run in the method chain.
+        It should take one DataFrame object as a parameter and have no return.
+        If there is a return, it will be ignored.
+    :param args, kwargs: Optional arguments and keyword arguments for `func`.
+    :returns: The input pandas DataFrame.
+    """
```
```suggestion
    """ # noqa: E501
```

This will tell flake8 to skip the line length check on the docstring. In this particular case, I find it reasonable that line 2986 goes beyond the 79 chars that we use (I have it set to 79 because in VSCode on a 15" MBP, I can split the window left-right and still see the full source code nicely.)

ericmjl (changes requested):
Very thoughtful setup on the tests, @sauln! Thanks for the effort put in 😄. I'm only requesting some changes regarding docstrings.

The `spark` tests might fail locally, but that shouldn't worry you. As long as they don't fail on the CI, which is the ultimate arbiter of whether the tests passed or not, that's cool with me.

Speaking of the CI, it was `flake8` that was failing. Some of my suggestions below should help. 

Other than that, it's just adding docstrings to the tests. I noticed I had trouble deciphering some tests that others had submitted, so I've started asking for docstrings on tests. (Hynek Schwalak has a great [blog post](https://hynek.me/articles/document-your-tests/) on this.) Please modify my suggestions if they're phrased incorrectly!

ericmjl:
@sauln I just realized I forgot to respond to one of your questions:

> Also, when I build the docs there is are a ton of changes. Do you want the new files in docs/reference committed?

No need to worry about these files, as they are built automatically on `readthedocs` and for the backup deploy to github pages!

tests/functions/test_also.py
```diff
@@ -0,0 +1,81 @@
...
+@pytest.mark.functions
+def test_also_args(dataframe):
+    """Test that the args are passed through to the function."""
+    method = Mock(return_value=None)
+    _ = dataframe.also(method, 5)
+    assert method.call_args.args[1] == 5
+
+
+@pytest.mark.functions
+def test_also_kargs(dataframe):
```
ericmjl: Oops! Just a minor nitpick - `kwargs` instead of `kargs`.

```suggestion
def test_also_kwargs(dataframe):
```

ericmjl (approved):
Everything else looks great. Thanks @sauln!

sauln:
Awesome! Thank you @ericmjl for the thorough feedback and interest in this new feature :D I'm excited to make use of it in my own day-to-day.

sauln:
Alright, it looked like I was using some features of `unittest` only available in Python 3.8.. I've reworked the tests with mocks to reflect the 3.6 API (https://docs.python.org/3.6/library/unittest.mock.html#calls-as-tuples).  

@ericmjl  it looks like `DeepSource` doesn't like the tuple unpacking the Python docs recommend. Is this something we can silence, or would you rather I rework it to avoid the `bug risk`?

ericmjl:
@sauln hmm, I thought that the tuple unpacking inside the tests would be fine. Is there a way to rework those lines such that the bug risk is removed? If not, don't fret it, just add the `skipcq:` piece on those lines.

ericmjl:
@sauln thanks for the changes! I’m going to invite review from at least one more core dev before we merge (or wait 2 days, whichever is earlier). This prevents the situation where only one person, i.e. myself, knows what’s going on in the codebase.

sauln:
Sweet, thank you!

samukweku (approved):

PR #738, opened 2020-09-07
[INF] Adding interrogate to CI checks

ericmjl:
The intent behind this PR is to ensure that docstring coverage never drops below its current state.

New functions that are added should all have docstrings covered. This includes source code and tests.

To kickstart this process, I started adding a few docstrings into some of the tests.

Snuck in a few pieces as well:

1. Commented out Windows build. It never fails when it should, so we shouldn't run them until they can be debugged correctly.
2. Update the docs as well.


CONTRIBUTING.rst
```diff
@@ -32,7 +32,7 @@ To get started:
 2. Ensure you have Docker running on your local machine.
 3. Ensure you have VSCode running on your local machine.
 4. In Visual Studio Code,
-   click on the quick actions Status Bar item in the lower left corner.
+    click on the quick actions Status Bar item in the lower left corner.
 5. Then select "Remote Containers: Open Repository In Container".
```
samukweku: Seems Microsoft has changed `"Remote Containers: Open Repository In Container"` to `"Remote Containers: Clone Repository In Container Volume"`

samukweku (approved):

.azure-pipelines/pipeline-master.yml
```diff
@@ -45,21 +45,22 @@ jobs:
...
 
-  - job: windows
-    variables:
-      activate.command: "activate"
-      JANITOR_CI_MACHINE: 1
-    strategy:
-      matrix:
-        py37:
-          python.version: "3.7"
+  # Commenting out Windows build because it never fails when it should...
```
loganthomas, line 48: I've created an issue to capture this - #751

CONTRIBUTING.rst
```diff
@@ -108,7 +108,7 @@ To see the main pyjanitor page,
 open the ``index.html`` file.
 
 .. note:: If you get any errors related to Importing modules when running `make docs`,
```
loganthomas: Double backtick for make docs here?

CONTRIBUTING.rst
```diff
@@ -130,10 +130,10 @@ This helps us keep track of who is working on what.
 7. Create a branch for local development:
 
 New features added to ``pyjanitor`` should be done in a new branch you have based off the latest version of the ``dev`` branch.
-The protocol for ``pyjanitor`` branches for new development is that the ``master`` branch mirrors the current version of ``pyjanitor`` on PyPI,
-whereas the ``dev`` branch is for additional features for an eventual new official version of the package which might be deemed slightly less stable.
-Once more confident in the reliability/suitability for introducing a batch of changes into the official version,
-the ``dev`` branch is then merged into ``master`` and the PyPI package is subsequently updated.
+The protocol for ``pyjanitor`` branches for new development
+the ``dev`` branch is for additional features for an eventual new official version of the package
```
loganthomas: I think there is a word or a few words missing here.

Maybe "The protocol for ``pyjanitor`` branches for new development is that the ``master`` branch mirrors the current version of ``pyjanitor`` on PyPI, whereas the ``dev`` branch..."

loganthomas (changes requested):

CONTRIBUTING.rst
ericmjl: Great catch, thanks @samukweku!

CONTRIBUTING.rst
ericmjl: Yes, great catch, thanks @loganthomas!

CONTRIBUTING.rst
ericmjl: You're right - I botched the deletion of the reference to the `master` branch. I think we're going to just let the `master` branch wither away - `dev` is where all the action has been, and we have the ability to tag a commit as a release, so I think we're good to go.

ericmjl:
@loganthomas a few dependencies were missing in `requirements.txt`, which I discovered from the netlify build logs. 

I think I have to add you all onto "the team" in netlify or something like that. I'll take this to the dev discord.

loganthomas (approved):

PR #769, opened 2020-10-26
[INF] Use netlify build hook instead of automatic build previews

ericmjl:
This hopefully helps keep the build minutes usage under control. Previously, we would build on every commit. Now builds are triggered only on every _successful_ build. It's still automatic, just not indiscriminate.


ericmjl:
Netlify build previews aren't selectively trigger-able, unfortunately. All good, if we go over on minutes, I don't mind paying occasionally for the build previews, they are very useful for previewing builds.

PR #778, opened 2020-11-25
[ENH] fill_direction improve

samukweku:
# PR Description

- use existing functions to check type and presence of columns

**This PR improves #704 .**

# PR Checklist

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers

- @ericmjl


samukweku:
@ericmjl @hectormz @sallyhong and others. Need your advice on the API. Currently different values for different columns cannot be applied for the limit parameter; if there are multiple columns, the same `limit` will be applied to all the columns. I am of the opinion that the API should be changed to allow different `limit` for different columns. It looks likely to break current API. Or should I just leave as is; probably users don’t care much for `limit`?

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ new version (on deck)
 - [ENH] Add ``pivot_wider`` function, which is the inverse of the `pivot_longer`
   function. @samukweku
 - [INF] Add `openpyxl` to `environment-dev.yml`. @samukweku
+- [ENH] Reduce code by reusing existing functions for fill_direction. @samukweku
```
ericmjl, line 16: :tada:

janitor/functions.py
```diff
@@ -4332,56 +4350,51 @@ def fill_direction(
...
+            raise ValueError(
+                """
+                The direction should be a string and should be one of
+                `up`, `down`, `updown`, or `downup`.
+                """
+            )
 
+    # TODO: option to specify limit per column; current implementation
+    # is one `limit` for all the columns. Might need refactoring, or an
+    # API change.
```
ericmjl, line 4398: This is a classic "do we apply this function to one column, or multiple columns" situation! 

As the library matures, I've grown to realize that having singular and plural functions is not necessarily the best, so allowing a function to be applied to multiple columns simultaneously is probably the better way. There must be some way we can "compose" them together, I think, such that we get the desired behaviour. I'm kind of inspired by what I see in JAX, where we have `vmap` to map a function over an array axis. This makes composing functions together more natural.

Maybe a better solution to this problem is to provide a `.map_columns()` function, which allows us to map any pyjanitor function across columns. Let's talk about this in another setting. For now, I like what you've done here, @samukweku, in that you've left this API piece as a TODO to think about later on.

tests/functions/test_fill_direction.py
```diff
@@ -85,35 +85,35 @@ def df():
 
 
 def test_fill_column(df):
-    """Fill down on a single column."""
+    "Fill down on a single column."
```
ericmjl: Docstrings! They should be triple-quoted.

ericmjl (changes requested):
The only change I'm requesting, @samukweku, is about docstring triple quotes, nothing else!

ericmjl (approved):
I approve, @samukweku! Let's get this one merged in. You've done a massive job with recent PRs, please take the honour of merging this one!

Issue #789, opened 2021-01-10
Duplicate Columns in some cases with `select_columns`


samukweku:
# Brief Description

`select_columns` sometimes returns duplicate columns, especially when using the `*` symbol. I believe only unique columns should be returned.

# System Information
- Linux OS (Ubuntu 20.04)
- pyjanitor 0.20.10
- python 3.9

# Minimally Reproducible Code

 ```python
df = pd.read_csv("iris.csv")
df.head()

	sepal_length	sepal_width	petal_length	petal_width	species
0	5.1	   3.5	           1.4	           0.2	       setosa
1	4.9	   3.0	           1.4	           0.2	       setosa

df.select_columns(["petal*", "*width"]).head(2) # select columns that start with petal or end with width

petal_length | petal_width | sepal_width | petal_width
1.4 | 0.2 | 3.5 | 0.2
1.4 | 0.2 | 3.0 | 0.2

# petal_width is repeated twice.
```

ericmjl:
This is a good catch, @samukweku! I think the simple fix here is to call on `df.drop_duplicate_columns()` inside `select_columns()`. Before we go to a PR for this one, is there a case where dropping duplicate columns might not be desirable?

samukweku:
@ericmjl `drop_duplicate_columns` fixes it. However, I feel that if we make the changes and include slicing, and possibly regex options, then it would be best to adapt it to the existing for loop within `select_columns`. Also feel it should be possible to accept a string if it is just one column that will be selected.  Also avoid copying the dataframe twice 😁

Another important part which I feel keen about is I can port it into any function that requires column selection. That will significantly reduce many checks for column selections and gymnastics because we have one function that takes care of that already. It also allows reusability, and gives power/flexibility in terms of column selection. This part is significantly inspired by how dplyr allows consistent column selections in various functions. Makes it seamless.

While on the topic, does it make sense to add a level argument, if the columns are a multiindex?

ericmjl:
That sounds amazing, @samukweku! I especially like the idea of targeting `select_columns()` as a function that lives closer to the primitive level for any thing that requires column selection. Let's go with it.

Level sounds like a good idea as well. Let's go with it!

Do you have a plan? Is there a way I can chip in, for example, if there's a piece that seems a bit substantial that you'd like a second pair of hands working on it?

samukweku:
[...]

PR #790, opened 2021-01-17
[ENH] Select columns refactor

### LINKED ISSUE #784: Add slice feature to `select_columns`
# Brief Description

I love the [select_columns]() method. It is powerful, especially the glob style name selection (beautiful!). I wonder if it is possible to add slice option, or if it is just unnecessary.

# Example API
Currently, you pass the selection to a list:

```python

# current implementation
df = pd.DataFrame(...).select_columns(['a', 'b',  'c', 'col_*'], invert=True)
```
I think it would be nice if we could do this:

```python
# proposed addition
df = pd.DataFrame(...).select_columns(['a': 'c', 'col_*'], invert=True)
```
Probably similar to what you do with [np.r_](https://numpy.org/doc/stable/reference/generated/numpy.r_.html)  (which is only for indices)

samukweku:
# PR Description

- Base select function that can be reused in other janitor functions, with some more flexibility
- Removes duplicate columns, and slice feature added
-

**This PR resolves #784 and #789.**

# PR Checklist

## Quick Check

    - black formatting
    - flake8 checking
    - running the test suite
    - docs build

## Code Changes

## Documentation Changes

- [x] Build the docs locally.

# Relevant Reviewers

- @ericmjl


samukweku:
@ericmjl made the PR. I would need your help in reviewing it, and if you have time on you, and you feel it is worthwhile, maybe you could extend the function to accept integers, or slice/range/generator of integers. I dont think it would be worthwhile to mix integers with strings; infact I don't know if it is worthwhile to extend the function to integers. Just a thought.

samukweku:
@ericmjl would love your take on the errors picked by DeepSource. A bit surprised about the error message regarding classes in pytest

ericmjl:
That's definitely surprising, @samukweku! Let me take a look at it this morning. Thanks for being patient with my review speed, I finally learned how to use a baby carrier 😂 😹

janitor/utils.py
```diff
@@ -31,11 +34,14 @@ def check(varname: str, value, expected_types: list):
     """
     is_expected_type = False
     for t in expected_types:
-        if isinstance(value, t):
-            is_expected_type = True
+        if t is callable:
+            is_expected_type = t(value)
+        else:
+            is_expected_type = isinstance(value, t)
+        if is_expected_type is True:
```
ericmjl: I think this can be simplified:

```python
if is_expected_type:
    break
```

This can be coupled with an in-line type annotation first declaring `is_expected_type`:

```python
is_expected_type: bool = False
```

janitor/utils.py
```diff
@@ -31,11 +34,14 @@ def check(varname: str, value, expected_types: list):
...
             break
 
-    if not is_expected_type:
+    if is_expected_type is False:
```
ericmjl: I'm not sure where I saw the following point, but I think the idiomatic/preferred way of writing conditionals based on whether something is boolean True is to use the `if is_expected_type` or `if not is_expected_type` pattern.

janitor/utils.py
```diff
@@ -1671,3 +1677,170 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+
+def _select_columns(
+    df: pd.DataFrame,
+    columns_to_select: Union[str, slice, Pattern, callable, list],
+) -> Union[list, pd.Index]:
+    """
+    Helper function to select columns from a dataframe.
+    The aim is to make this function a base for the various
+    functions in the janitor package that require column selection,
+    while offering a lot of options/flexibility. It is inspired
```
ericmjl: Minor nitpick: SEMBR please.

janitor/utils.py
```diff
@@ -1671,3 +1677,170 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
     df = df.astype(categories_dtypes)
 
     return df
+
+
+def _select_columns(
+    df: pd.DataFrame,
+    columns_to_select: Union[str, slice, Pattern, callable, list],
+) -> Union[list, pd.Index]:
+    """
```
ericmjl, line 1694: This function is very nice and flat! :smile:

janitor/utils.py
```diff
@@ -1671,3 +1677,170 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+    for label in columns_to_select:
+        check("column label", label, [str, slice, Pattern, callable])
+
+    all_columns = []
+    for entry in columns_to_select:
+        outcome = __columns_not_a_list(df, entry)
+        all_columns.extend(outcome)
+
+    if not pd.Series(all_columns).is_unique:  # remove duplicate column names
+        return pd.unique(all_columns)
```
ericmjl: I think this check can be simplified. Instead of adding two lines to check, we can check at the `all_columns.extend(outcome)` line. Here's a proposed change:

```python
all_columns.extend([c for c in outcome if c not in all_columns])
```

janitor/utils.py
```diff
@@ -1671,3 +1677,170 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+
+    return all_columns
+
+
+def __columns_not_a_list(
```
ericmjl: Do we have a `_columns_not_a_list` (single underscore)? If so, is there a reason to do double underscores? (Having a single vs double underscore convention seems like introducing two layers of so-called private functions, which is philosophically opposed to "flat is better than nested" in the Zen of Python.)

janitor/utils.py
```diff
@@ -1671,3 +1677,170 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+    filtered_column = None
+    start_check = None
+    stop_check = None
+    step_check = None
+    if isinstance(columns_to_select, str):
+        # takes care of the shell_like glob strings
+        if "*" in columns_to_select:
+            filtered_column = fnmatch.filter(df.columns, columns_to_select)
+        else:
+            if columns_to_select in df.columns:
```
ericmjl: This could probably be made into an "elif" block instead. Keeps the function flat.

janitor/utils.py
```diff
@@ -1671,3 +1677,170 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+            raise ValueError(
+                "The callable provided must return a sequence of booleans."
+            )
+
+        # cant use `is` here, since it may be a numpy bool
+        checks = any(
+            (column == True for column in filtered_column)  # noqa: E712
+        )
+
+        if checks is False:
```
ericmjl: Could be rewritten as `if not checks`.

Maybe something to think about for code readability purposes - is `passes_checks` possibly a more idiomatic variable name? We would get:

```python
if not passes_checks:
```

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,263 @@
...
+def test_regex_presence(df):
+    """
+    Raise ValueError if `columns_to_select` is a regex
+    and none of the column names match.
+    """
+    with pytest.raises(ValueError):
+        _select_columns(df, re.compile(r"^\d+"))
+
+
+class Test_Columns_not_List_Various_Inputs:
```
ericmjl: Can I ask what your reasons were for adopting a class data structure here? Is it more concise written this way than as a suite of pytest functions? Would it be reasonable to rewrite them to test functions rather than test classes, to keep the testing style consistent? Having two styles (without reason) makes it more jarring for a newcomer to contribute than having a single consistent style.

ericmjl (changes requested):
Great work, @samukweku :smile: Thanks for the time put into refactoring `select_columns`. In particular, I like the block where you have clearly delineated behaviour based on the column selection strategy. This is much more powerful than what we had before.

Regarding DeepSource, I looked through the issues, and I trust what they've asked for. Could you make the necessary fixes, please? In particular, the nested "if x: if y" without alternate code paths means we can go less nested in the code, which is considered in line with good Python style. 

As for the object issue, the issue at hand is us being non-explicit about what attributes an object _ought_ to have, which is idiomatically defined in the `__init__`. I've seen the code pattern where a code developer adds new attributes in non `__init__` class methods and have made the same mistake myself. This leads to confusion about guarantees over what attributes exist. If the effort is not too great, I would advise rewriting those tests as test functions rather than test classes.

All-in-all, I'm happy with the PR! Looking forward to the changes; if not much else gets changed, then once these requested changes are made, and DeepSource is kept happy, then I'm happy to approve and merge :smile:.

ericmjl:
@samukweku if you merge in the latest `dev` now, most of the DeepSource issues should go away! :smile:

tests/utils/test__select_column.py
samukweku: just me and shiny things :). Was going through the pytest docs and discovered them; looked nice to have similar tests in one class. Definitely reorganize them to functions instead.

janitor/utils.py
ericmjl: Update: Now that I have seen dispatch, I _think_ we can do dispatching :smile:

tests/utils/test__select_column.py
ericmjl: Oh, I see! Haha, I was wondering where that came from. :smile:

We can use pytest marks to help organize tests together. For example, in another project, I helped break out slow tests from fast tests, so that they could run in parallel. Please feel free to add marks as you see appropriate! (But don't go too crazy with their granularity, haha.)

samukweku:
@ericmjl Getting this error when running code within the VSCode docker environment:
```
ImportError while loading conftest '/workspaces/pyjanitor/tests/conftest.py'.
tests/conftest.py:7: in <module>
    from janitor.testing_utils import date_data
janitor/__init__.py:7: in <module>
    from .functions import *  # noqa: F403, F401
janitor/functions.py:34: in <module>
    from multipledispatch import dispatch
E   ModuleNotFoundError: No module named 'multipledispatch'
```

Any suggestions on how I can fix it?

ericmjl:
Yes! Let's see, if you got VSCode to rebuild the container, that should do the trick. Multiple dispatch was missing from the conda env inside the container, from what I can see.

samukweku:
@ericmjl probably missing some other trick; I rebuilt the container. still no luck

ericmjl:
Hmmm, that's really weird. When you reload VSCode inside the docker container, can you do the following?

1. Output the result of `which python`
2. Output the result of `python -c "import multipledispatch"`
3. Output the result of `conda list | grep multipledispatch`

samukweku:
@ericmjl :


    Output the result of which python
```/opt/conda/bin/python```
    Output the result of python -c "import multipledispatch"
```
(base) root@48ef221d4b10:/workspaces/pyjanitor# python -c "import multipledispatch"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'multipledispatch'
```
    Output the result of conda list | grep multipledispatch - no results

**Comment by ericmjl** [comment]
Hmmm, as I suspected. The container's Python doesn't appear to be using the environment Python. What is the result of `conda list` itself? Is the environment name correctly showing there? If not, then there's an issue with how the container is built. I might need a bit of time to debug.

Update: I just tried creating the container on my Mac, no issues:

```bash
(base) root@55bad2ca8cb1:/workspaces/pyjanitor# conda list | grep rdkit
rdkit                     2020.09.2        py38h0d43b38_0    conda-forge

(base) root@55bad2ca8cb1:/workspaces/pyjanitor# conda list | grep pandas-flavor
pandas-flavor             0.2.0                      py_0    conda-forge
```

I guess if you've rebuilt the container correctly it should work. Maybe do a `docker system prune -a -f` after a reboot, before opening up the container again?

samukweku:
@ericmjl I did a conda install which resolved the issue; I'll rebuild the container and try out your suggestion though. However, I am using functools.singledispatch, as multipledispatch does not recognize typing.Pattern and collections.abc.Callable. I'll add examples to the `select_columns` docs for your to review as well

janitor/utils.py
```diff
@@ -1680,3 +1687,176 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+    checks = any((column == True for column in filtered_columns))  # noqa: E712
+
+    if not checks:
+        raise ValueError("No results were returned for the callable provided.")
+
+    filtered_columns = df.columns[filtered_columns]
+    return filtered_columns
+
+
+# hack to get it to recognize typing.Pattern
```
ericmjl, line 1821: That is quite the hack. In the interest of maintainability, would you be up for explaining the `re.compile` piece? I'm not well-versed enough in regex to know what's happening - what is being searched for? How come the "type of a re.compile" works here? 

If there's a reference you used that explains this piece, can you stick it in the docstring too?

janitor/utils.py
```diff
@@ -1680,3 +1687,176 @@ def _computations_as_categorical(df: pd.DataFrame, **kwargs) -> pd.DataFrame:
...
+        filtered_columns = fnmatch.filter(df.columns, columns_to_select)
+    elif columns_to_select in df.columns:
+        filtered_columns = [columns_to_select]
+    if not filtered_columns:
+        raise NameError(f"No match was returned for '{columns_to_select}'")
+    return filtered_columns
+
+
+@_select_columns.register(slice)
+def _(columns_to_select, df):
```
ericmjl: Using `_` as the function name collides with the Python convention of using `_` as a variable to ignore. Could you change it to something like `def col_sel_dispatch`, or a more informative name?

I did a bit of googling myself, it looks like you might have read [the blog post on pythonlibrary.org](https://www.blog.pythonlibrary.org/2016/02/23/python-3-function-overloading-with-singledispatch/), is that right? :joy_cat:

A possibly saner convention with functions is described in [Michael Chow's blog post](https://mchow.com/posts/2020-02-24-single-dispatch-data-science/), and is what I'm thinking of above.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+def test_regex_presence(df):
+    """
+    Raise ValueError if `columns_to_select` is a regex
+    and none of the column names match.
+    """
+    with pytest.raises(ValueError):
+        _select_columns(re.compile(r"^\d+"), df)
+
+
+def test_strings(df1):
```
ericmjl, line 138: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+    assert _select_columns("id", df1) == ["id"]
+    assert _select_columns("*type*", df1) == [
+        "type",
+        "type1",
+        "type2",
+        "type3",
+    ]
+
+
+def test_slice(df1):
```
ericmjl, line 149: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+        df1.loc[:, slice(None, "code2")].columns,
+    )
+    assert_index_equal(_select_columns(slice(None, None), df1), df1.columns)
+    assert_index_equal(
+        _select_columns(slice(None, None, 2), df1),
+        df1.loc[:, slice(None, None, 2)].columns,
+    )
+
+
+def test_callable_data_type(df1):
```
ericmjl, line 170: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+        _select_columns(pd.api.types.is_datetime64_dtype, df1),
+        df1.select_dtypes(np.datetime64).columns,
+    )
+    assert_index_equal(
+        _select_columns(pd.api.types.is_object_dtype, df1),
+        df1.select_dtypes("object").columns,
+    )
+
+
+def test_callable_string_methods(df1):
```
ericmjl, line 201: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+    assert_index_equal(
+        _select_columns(
+            lambda x: x.name.startswith("code") or x.name.endswith("1"), df1
+        ),
+        df1.filter(regex=r"^code.*|.*1$").columns,
+    )
+
+
+def test_callable_computations(df1):
```
ericmjl, line 232: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+    assert_index_equal(
+        _select_columns(lambda x: x.isna().any(), df1),
+        df1.columns[df1.isna().any().array],
+    )
+
+
+def test_regex(df1):
```
ericmjl, line 243: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

tests/utils/test__select_column.py
```diff
@@ -0,0 +1,248 @@
...
+    assert _select_columns(re.compile(r"\d$"), df1) == list(
+        df1.filter(regex=r"\d$").columns
+    )
+    assert _select_columns(patterns(r"\d$"), df1) == list(
+        df1.filter(regex=r"\d$").columns
+    )
+
+
+def test_list_various(df1):
```
ericmjl, line 253: Missing docstrings for intent of the test function. Going to copy/paste below, pardon me please.

janitor/functions.py
```diff
@@ -3061,66 +3061,170 @@ def currency_column_to_numeric(
...
+    )
+    if isinstance(search_column_names, list):
+        for label in search_column_names:
+            check("column label", label, [str, slice, Pattern, callable])
 
-    if len(missing_wildcards) > 0:
-        raise NameError(
-            f"Search string(s) {missing_wildcards} not found in DataFrame"
-        )
+    full_column_list = _select_columns(search_column_names, df)
```
ericmjl, line 3223: So much cleaner here :smile:.

ericmjl (approved):
Pre-approving @samukweku! Mostly asking for two things in this final review:

1. Docstrings for the intent of the tests.
2. A slightly saner set of names for the dispatched functions.

Because these are cosmetic, I'm going to pre-approve, and as soon as the tests pass, please hit the big green button!

janitor/utils.py
samukweku: Actually noticed it in the python [docs](https://docs.python.org/3/library/functools.html#functools.singledispatch) :grinning:

janitor/utils.py
samukweku: type(re.compile(r"\d")) returns the type for a regex, which is re.Pattern. functools.singledispatch has no problem with that; if however, I typed in re.Pattern, it associates it instead with the typing module and fails.

ericmjl:
Please hit the big green button, @samukweku!

PR #791, opened 2021-01-24
[ENH] Refactor label_encode to use multiple dispatch.

ericmjl:
By using multiple dispatch, we can have a single function name that behaves correctly based on the data types of the input functions. This beats out checking types inside a function in order to determine that function's behaviour. Apparently, [PEP443](https://www.python.org/dev/peps/pep-0443/) says checking types inside a function to determine behaviour is a bit of an anti-pattern. I've led countless people down the wrong path :joy_cat:.

@samukweku using single/multiple dispatching might be a better way to approach type-specific behaviour because we can get an automatically-raised `NotImplementedError` for things not implemented. I should have realized this programming behaviour earlier on.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -21,6 +21,7 @@ dependencies:
 - lxml
 - make
 - missingno
+- multipledispatch
 - mypy
 - natsort
 - nbsphinx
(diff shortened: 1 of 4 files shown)


samukweku (approved):
Now I have a better understanding of `dispatch`!

PR #792, opened 2021-01-25
[ENH] Switched fill_empty to dispatch pattern

ericmjl:
As per title. 

Everything else should work properly, I think.


samukweku (approved):

ericmjl:
I'm going to try adding coverage to the tests, so there's no excuse :)

PR #823, opened 2021-03-27
[ENH] General Fixes

### LINKED ISSUE #812: Enhancements/Fixes/Suggestions for pivot_longer/pivot_wider
-  [X] use the `pivot` function instead of `stack/unstack`, for `pivot_wider`, since the minimum version is pandas 1.1 (the `pivot` function in versions prior to pandas 1.1 did not support lists in the `index`/`columns` parameters)
- ~~not sure if it is necessary to have an `aggfunc` parameter in `pivot_wider`. Maybe deprecate?~~
- ~~use `stack` where possible for `pivot_longer` - it should offer a simpler code and more maintainable~~
- ~~stick to single index columns for `pivot_longer` - more maintainable (current code looks spaghetti and brittle); any multiIndex column transformations can be done by the user with `stack` or `melt`. if accepted, this means deprecating the `col_level` parameter. also remove it from the `_select_columns` function.~~
- [X] Use the `ignore_index` capability from pandas `melt` directly, instead of the custom function created (since pandas minimum version in pyjanitor is now 1.1)
- [X] For `select_columns` return tuples as tuples (allows column names that are tuples to be selected)
- [X] remove `fill_value` from `complete`; user can always achieve this with better results with `fillna`
- [X] fix errors with groupby in `complete`

samukweku:
# PR Description

- Drop `fillna` from `complete` function. Users are better served with `fillna` instead.
- Use `pivot` for `pivot_wider`, since minimum pandas version is 1.1
- Fix errors with the `by` parameter in the `complete` function.
- `pivot_longer` uses more functions directly from the Pandas library directly (`ignore_index`)

**This PR resolves #812.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -3218,6 +3218,7 @@ def select_columns(
               .select_columns(['a', 'b', 'col_*'],
               invert=True))
 
+
     :param df: A pandas DataFrame.
     :param search_column_names: Valid inputs include:
 
@@ -3241,13 +3242,16 @@ def select_columns(
         raise ValueError(
             """
             MultiIndex columns
-            not supported for `select_columns`.
+            are not supported for `select_columns`.
             """
         )
 
     # applicable for any
-    # list-like object (ndarray, Series, pd.Index, tuple, ...)
-    if is_list_like(search_column_names):
+    # list-like object (ndarray, Series, pd.Index, ...)
+    # excluding tuples, which are returned as is
+    if is_list_like(search_column_names) and (
+        not isinstance(search_column_names, tuple)
+    ):
         search_column_names = list(search_column_names)
 
     full_column_list = _select_columns(search_column_names, df)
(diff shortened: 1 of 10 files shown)


janitor/functions.py
```diff
@@ -5781,10 +5762,10 @@ def pivot_longer(
 @pf.register_dataframe_method
 def pivot_wider(
     df: pd.DataFrame,
-    index: Optional[Union[List, str]] = None,
-    names_from: Optional[Union[List, str]] = None,
-    values_from: Optional[Union[List, str]] = None,
-    names_sort: Optional[bool] = False,
+    index: Optional[Union[Sequence[str], str]] = None,
+    names_from: Optional[Union[Sequence[str], str]] = None,
+    values_from: Optional[Union[Sequence[str], str]] = None,
```
ericmjl, line 5767: Usually I've seen `Iterable[str]`, rather than `Sequence[str]`. May I ask what's the difference?

ericmjl (approved):

ericmjl:
This is wonderful, @samukweku! Thanks for the work that went into the PR. I'm pre-approving. Let's see if we can get any more reviews in.

janitor/functions.py
samukweku: @ericmjl I think I noticed it in pandas pivot do strings; tried to follow their annotation 😁

ericmjl: Okie dokes! If the `pandas` dev team does it, I'm sure there's a good reason.

ericmjl:
@samukweku it's been a bit more than a week now, so I think we should merge!

samukweku:
Just noticed a bug in pivot longer; I will push a fix for it today

PR #830, opened 2021-04-20
Pin py at 1.10.0

ericmjl:
h/t @jk3587 for pointing this out!


jk3587:
The only difference for this diff is that we bumped py to 1.10.0 and there was an extra whitespace in front of each comment. Coverage going down is likely just noise.

PR #867, opened 2021-08-14
[DOC] update gh-pages with docs build from dev branch

loganthomas:
# PR Description

Please describe the changes proposed in the pull request:

- Not sure if this is the right approach, but I noticed that the docs haven't been updated recenlty. I think this is due to the fact that we are using a `gh-pages` branch and haven't merged updates to it recently.
- I executed a `make docs` from the `dev` branch and uploaded the html files to the `gh-pages` branch within this PR.

**See conversation with @samukweku in #865**

# PR Checklist
Please ensure that you have done the following:

1. [X] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [X] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

3. [X] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

# Automatic checks

There will be automatic checks run on the PR. These include:

- Building a preview of the docs on Netlify
- Automatically linting the code
- Making sure the code is documented
- Making sure that all tests are passed
- Making sure that code coverage doesn't go down.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #870, opened 2021-08-15
[INF] Add docs build GitHub action

### LINKED ISSUE #868: [INF] Add GH actions workflow to automatically build dev docs
Our docs are built off of the dev branch. The static site docs that are generated by sphinx should then be force pushed to the gh-pages branch.

I have a pattern for doing these in other projects that we might be able to follow:

- nxviz: https://github.com/ericmjl/nxviz/blob/master/.github/workflows/docs.yaml
- Network Analysis Made Simple: https://github.com/ericmjl/Network-Analysis-Made-Simple/blob/master/.github/workflows/deploy.yaml#L66
- dl-workshop: https://github.com/ericmjl/dl-workshop/blob/master/.github/workflows/build-website.yaml#L56

I believe the key here is to use the peaceiris website deploy. 

Referencing #867; once this is implemented, we shouldn't need to do PRs like this going forward.

ericmjl:
This allows us to build docs on `dev` branch automatically on every push to `dev`. 

Enabling this PR should also nudge us to do more frequent releases especially when we have an API change.

 Tagging @loganthomas, it should save us the need to do what you did in #867.

Resolves #868.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - dev
+      - ghactions-docs
 
 jobs:
   build-environment:
@@ -74,6 +75,7 @@ jobs:
         run: |
           source /tmp/pyjanitor-dev_env/bin/activate
           python -m ipykernel install --user --name pyjanitor-dev
+          pip install -e .
           make docs
 
       - name: Deploy website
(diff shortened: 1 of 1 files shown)


samukweku (approved):

ericmjl:
Okie dokes, I'm going to merge in this PR, as we'll need to update the docs for other PRs as well.

PR #892, opened 2021-08-25
[INF] simplify a bit linting, use pre-commit as CI linting checker

Zeroto521:
# PR Description

There is two similar linting CIs (pre-commit and code-checks) but also have tiny differences.
We should use one of them as the standard, not both. Otherwise, we would get confused.

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


.pre-commit-config.yaml
```diff
@@ -28,3 +28,8 @@ repos:
   hooks:
   - id: darglint
     args: [-v 2]  # this config makes the error messages a bit less cryptic.
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.9.2
+  hooks:
+    - id: flake8
+      args: [--exclude, nbconvert_config.py]
```
Zeroto521, line 30: code-checks.yml has flake8, but pre-commit doesn't.

.pre-commit-config.yaml
```diff
@@ -12,7 +12,7 @@ repos:
   rev: 21.7b0
   hooks:
   - id: black
-    args: [-l 79]
+    args: [-c, pyproject.toml]
```
Zeroto521: code-checks.yml's config is complicated, but pre-commit's config is simple.

Zeroto521:
Because the linting standard is different, so https://github.com/pyjanitor-devs/pyjanitor/pull/892/checks is red.

Now we should determine to use which linting is the standard.
code-checks.yml or pre-commit?

ericmjl:
@Zeroto521 thanks for handling this! In the interest of minimizing the diff, I think we should defer to the codechecks.yml file for the settings, but use pre-commit as our actual code checker. Would you be open to making that happen please?

.pre-commit-config.yaml
```diff
@@ -12,12 +12,7 @@ repos:
   rev: 21.7b0
   hooks:
   - id: black
-    args: [-l 79]
-- repo: https://github.com/kynan/nbstripout
-  rev: 0.5.0
-  hooks:
-  - id: nbstripout
-    files: ".ipynb"
```
Zeroto521, line 20: nbstripout should be removed.
It changed a lot, even would make ipynb outputs become empty.

Zeroto521 (commented):
Now use the old code-checks.yml as the base standard and pre-commit as the lining checker.

But these hooks weren't removed. They changed a little. Just remove some whitespace when lint.
```yaml
hooks:
  - id: trailing-whitespace
  - id: end-of-file-fixer
  - id: check-yaml
  - id: check-added-large-files
```

.pre-commit-config.yaml
ericmjl: Got it, thanks @Zeroto521.

We can remove it for now. In a future setting, though, I think we'd want to do the notebooks such that they are built in the CI system and deployed.

ericmjl:
Let's get one more review in before merging. Tagging @loganthomas and @hectormz, whom I think might take an interest in the PR.

ericmjl (approved):

.pre-commit-config.yaml
Zeroto521: I see conf.py already adds nbsphinx, but no a `.rst` file index these notebooks.

ericmjl: Right. Regarding the notebooks (and docs more generally), I have been debating, btw, to switch the docs over to mkdocs + mkdocs-material. It is much easier to maintain in the long-run than Sphinx; I know I have been hesitant to make any changes to the docs as it stands right now because of a lot of magic that happens behind the scenes that I'm just not well-versed in, whereas MkDocs and mkdocs-material are much easier to maintain combined. (It feels... flatter.)

ericmjl:
@Zeroto521 I'm going to merge as I'd like to have these changes present in #897. Thanks for your contributions here, much appreciated!

PR #909, opened 2021-09-05
[DOC] MkDocs Finance

nvamsikrishna05:
This PR is for docstring changes to finance submodule

- [x] `inflate_currency`
- [x] `convert_stock`
- [x] `get_symbol`

This doesn't have 8 functions as mentioned in the main PR. Since couple of Functions are commented out due to unavailability of exchange rates API and few are internal functions.

Please review - @ericmjl  @loganthomas  @samukweku 

I have observed that `convert_stock` and `get_symbol` aren't functioning. Can anyone confirm if I'm doing something wrong or something changed in the API ? The direct access to the url Internally used isn't working in my quick test.

Also `convert_stock` and `get_symbol` both do the same functionality and one of them can be made internal function also especially `get_symbol` which does the api call and response parsing.

P.S. The tests for those two methods are already marked as xfail in pytest.


janitor/finance.py
```diff
@@ -630,51 +630,52 @@ def inflate_currency(
     the currency's country.
 
     The provided country can be any economy name or code from the World Bank
-    list of economies:
-    https://databank.worldbank.org/data/download/site-content/CLASS.xls.
+    [List of economies](
+    https://databank.worldbank.org/data/download/site-content/CLASS.xls).
```
ericmjl: @nvamsikrishna05 going to suggest a slightly different way of handling:

```suggestion
    Here is the [list of economies][econ]

    [econ]: https://databank.worldbank.org/data/download/site-content/CLASS.xls.
```

ericmjl (approved):

PR #925, opened 2021-09-12
[DOCS] Mkdocs function - 3

nvamsikrishna05:
This PR contains the mkdocs docstring fixes for the following functions under the `function` module -

- [x] `conditional_join`
- [x] `pivot_wider`

### DIFF
--- a/janitor/functions.py
+++ b/janitor/functions.py
@@ -6306,6 +6306,7 @@ def pivot_wider(
 
     Reshaping to wide form :
 
+    ```python
              name variable  value
         0   Alice      wk1      5
         1   Alice      wk2      9
@@ -6332,9 +6333,11 @@ def pivot_wider(
         0    Alice     5     9    20    22
         1    Bob       7    11    17    33
         2    Carla     6    13    39    40
+    ```
 
     Pivoting on multiple columns is possible :
 
+    ```python
             name    n  pct
         0     1  10.0  0.1
         1     2  20.0  0.2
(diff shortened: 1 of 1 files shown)


janitor/functions.py
```diff
@@ -6820,8 +6844,7 @@ def conditional_join(
     :raises ValueError: if `left_on` and `right_on` in condition are not
         both numeric, or string, or datetime.
 
-
-    .. # noqa: DAR402
+     # noqa: DAR402
```
samukweku: I think you can get rid of the noqa here

janitor/functions.py
```diff
@@ -6488,7 +6498,7 @@ def pivot_wider(
     :raises ValueError: if values in `index` or `names_from` or `values_from`
         do not exist in the dataframe.
     :raises ValueError: if the combination of `index` and `names_from` is not
-        unique and `aggfunc` is `None``.
+        unique and `aggfunc` is `None`.
 
 
     .. # noqa: DAR402
```
samukweku: You can get rid of the noqa here

samukweku (changes requested):

janitor/functions.py
```suggestion
```

janitor/functions.py
```suggestion
```

ericmjl (approved):

ericmjl:
@nvamsikrishna05 we're failing on darglint errors, but I'll fix this in the `mkdocs-functions` branch so we don't hold this PR up.

PR #936, opened 2021-10-02
[DOC] MkDocs Utils Part 5

loganthomas:
This PR is for docstring changes to the utils submodule. Since utils has 74 functions, each PR will contain ~5 functions for review.

- [x]  `_create_indexer_for_complete()`
- [x] `_complete_indexer_expand_grid()`
- [x]  `_complete_column()`
- [x] `_sub_complete_column(str)`
- [x] `_sub_complete_column(list)`


janitor/utils.py
```diff
@@ -869,8 +870,8 @@ def _create_indexer_for_complete(
 
 def _complete_indexer_expand_grid(indexer):
     """
-    Generate indices to expose explicitly missing values,
-    using the `expand_grid` function.
+    Generate indices to expose explicitly missing values, using the
+    `expand_grid` function.
```
ericmjl: We should reinstate the original as the original docstring follows SEMBR. https://sembr.org

janitor/utils.py
```diff
@@ -900,14 +901,12 @@ def _complete_indexer_expand_grid(indexer):
 @functools.singledispatch
 def _complete_column(column, index):
     """
-    This function processes the `columns` argument,
-    to create a pandas Index or a list.
+    This function processes the `columns` argument, to create a pandas
+    Index or a list.
```
ericmjl: Original should be reinstated because it follows SEMBR.

janitor/utils.py
```diff
@@ -917,15 +916,12 @@ def _complete_column(column, index):
...
 def _sub_complete_column(column, index):  # noqa: F811
     """
-    This function processes the `columns` argument,
-    to create a pandas Index.
-
-    Args:
-        column : str
-        index: pandas Index
+    This function processes the `columns` argument, to create a
+    pandas Index.
```
ericmjl: Let's make sure this follows SEMBR.

janitor/utils.py
```diff
@@ -938,17 +934,13 @@ def _sub_complete_column(column, index):  # noqa: F811
 @_complete_column.register(list)  # noqa: F811
 def _sub_complete_column(column, index):  # noqa: F811
     """
-    This function processes the `columns` argument,
-    to create a pandas Index.
+    This function processes the `columns` argument to create a
+    pandas Index.
```
ericmjl: SEMBR.

ericmjl (changes requested):
Let's make sure docstrings follow SEMBR.

ericmjl:
@loganthomas for the set of MkDocs utils PRs, could you ensure conflicts are resolved?

I've merged `dev` into `mkdocs`, which brought in a bunch of conflicts that I had to resolve, which are identical to the conflicts you'll encounter, I think. I believe that the solution here is to simply accept every change in `mkdocs`, which will correspondingly bring in every change from `dev`.

PR #970, opened 2022-01-02
[ENH] Deprecate bin_numeric params and allow kwargs

### LINKED ISSUE #969: [DOC] bin_numeric parameters
# Brief Description of Fix

Currently, the docs for `bin_numeric` describes the parameters to be more restrictive than they actually are. According to the pyjanitor docs, `num_bins` is expected to be an Optional[int]:

image

However,
pyjanitor's `num_bins` is passed directly into the `bins` parameter of `pd.cut`, as in:
```python
# pyjanitor.bin_numeric src code snippet
df[str(to_column_name)] = pd.cut(df[str(from_column_name)], bins=num_bins)
```
But `bins` parameter of `pd.cut` also accepts list of values / intervalIndex, cf. [pandas docs](https://pandas.pydata.org/pandas-docs/version/1.3.4/reference/api/pandas.cut.html). Which means that at the moment, this is possible:

```python
df = pd.DataFrame(
    np.random.default_rng(42).integers(300, 800, size=(400, 2)),
    columns=['a', 'b'],
).reset_index()

(
    df
    .bin_numeric(from_column_name='a', to_column_name='binned_a', num_bins=[300, 450, 550, 650, 800])
    .head()
)
```
image

I see 2 possibilities:
1. Rename `num_bins` to `bins`, deprecate the latter parameter name, and update pyjanitors' docs accordingly to inform users that bins need not necessarily be integer.
2. Keep `num_bins`; in which case, let's check and enforce that the user is passing in an integer (?)

I'm not fond of option 2; I use `pd.cut` with bins specified as a list of cut-values alot in my daily work, so I would prefer to keep this functionality open.

---

In the same vein, how about also allowing **kwargs to be passed in to `bin_numeric` (and pass it directly to `pd.cut` within) so it is more flexible? I don't see a good reason why we are only exposing the `bins`&`labels` parameters to the end user.

To be clear, what I'm proposing is:
```python
def bin_numeric(
    df: pd.DataFrame,
    from_column_name: Hashable,
    to_column_name: Hashable,
    bins: Union[int,...] = 5,  # I'm actually not sure how to annotate this properly at the moment...
    labels: Optional[str] = None,
    **kwargs,
) -> pd.DataFrame:
    if not labels:
        df[str(to_column_name)] = pd.cut(
            df[str(from_column_name)], bins=bins, **kwargs,
        )
    [...snip...]
```

# Relevant Context

- [Link to documentation page](https://pyjanitor-devs.github.io/pyjanitor/api/functions/#janitor.functions.bin_numeric.bin_numeric)
- [Link to exact file to be edited](https://github.com/pyjanitor-devs/pyjanitor/blob/dev/janitor/functions/bin_numeric.py)

thatlittleboy:
# PR Description
This PR pertains to the [`bin_numeric`](https://pyjanitor-devs.github.io/pyjanitor/api/functions/#janitor.functions.bin_numeric.bin_numeric) function.

- Deprecate `num_bins` in favour of `bins` parameter.
- Remove `labels` parameter, and open `**kwargs` to the end user. As a result, we defer the `len(labels) == num_bins` check to pandas, don't do this check within pyjanitor.
- Add an additional test case to ensure `retbins` is not included in `kwargs`. Raise error if it is passed in. (Because retbins changes the output signature of `pd.cut`; we have to ensure only the pandas dataframe is returned.

**This PR resolves #969 .**

# PR Checklist

1. [X] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

3. [X] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.

# Automatic checks

# Relevant Reviewers

- @ericmjl


janitor/functions/bin_numeric.py
```diff
@@ -1,60 +1,66 @@
...
-    else:
-        if not len(labels) == num_bins:
-            raise ValueError("Number of labels must match number of bins.")
+    if "retbins" in kwargs:
+        raise ValueError("`retbins` is not an acceptable keyword argument.")
 
-        df[str(to_column_name)] = pd.cut(
-            df[str(from_column_name)], bins=num_bins, labels=labels
-        )
+    df[str(to_column_name)] = pd.cut(
```
samukweku: - let's make it such that it does not mutate the original dataframe; `assign` can be helpful here , in which case you use a dictionary within the assign function.
- do we need the `str` function? i'd say it is better to check if the column name is a string with the `check` function in utils, and check if the `from_column_name` exists in the df, with the `check_column` function, also in utils.

samukweku (changes requested):

janitor/functions/bin_numeric.py
thatlittleboy: 1. Sure. 2. I suppose not, I just assumed it was cast to str because the original author had intended to accept generic hashables for the input. I'll add in some checks to enforce the arguments as strings and change the docs accordingly.

samukweku: Ahhh I didn’t think of that. If that’s the case maybe do a check for hashable instead? Just not sure how to integrate that into assign

samukweku (approved):

janitor/functions/bin_numeric.py
thatlittleboy: My view is to keep it consistent across the library? If other similar functions (e.g. `add_column`) require the "new column name" parameter to be a string (+ if we haven't gotten complaints about not supporting generic hashables), we can stick with string here as well for the time being. :)

Also, according to `add_column` [function docstring](https://pyjanitor-devs.github.io/pyjanitor/api/functions/#janitor.functions.add_columns.add_column) for its `column_name` parameter: 

> Name of the new column. Should be a string, in order for the column name to be compatible with the Feather binary format (this is a useful thing to have).

ericmjl (approved):
Wonderful work, @thatlittleboy! Could you make sure the latest `dev` branch is merged into your PR branch? Once we're done, and we're all green, I'd like to invite anyone on @pyjanitor-devs/core-devs to do the honors of hitting that big green button! :tada:

ericmjl:
@samukweku, @nvamsikrishna05, or @loganthomas, whenever either of you have a moment, please feel free to do the honors here! :tada:

PR #977, opened 2022-01-07
[ENH] Add I/O function read_commandline()

Nalmeder:
# PR Description

First order of business: No automated tests, sadly. Not because I didn't feel like it or didn't want to- but rather pytest was giving me trouble and I wanted to get a bit of feedback before wasting too much time. It took me longer than I'd like to admit to throw in the towel on that one. 

Testing WOULD be done with the included  ```manual_read_command(string:str)``` in case any body wants to take up the mantle with that real quick.

Second: I did manual testing using a publicly available .csv spreadsheet of data, which I have named sep-quarter.csv. For those unfamiliar: ```argv``` is used to represent the commandline arguments. ```argv[0]``` is always the filename (thus ```python``` or ```python3``` are not included in the argument list) and everything from ```argv[1]``` until end-of-line is an additional command.

I have tested the following use cases:

``` python3 <filename>.py cat sep-quarter.csv ``` creates a pandas df from the .csv file, which in my testing is a big dataframe. About 19000 rows.

 ``` python3 <filename>.py 'cat sep-quarter.csv | grep '.SEA1AA' ' ```  creates a pandas df only from the infromation in the .csv file that contains .SEA1AA, which in my testing results in 41 rows. Much smaller, and only containing information relevant to what was queried, in this case, anything with .SEA1AA in one of its columns.

 NOTE: the entirety of argv[1] and beyond needs to be in a string for this, it doesn't work when it's not in quotes. The command I mentioned above this one however does work, with or without quotes. Unfortunately, when working with the shell there's always the possibility of a conflict: I think the ``` | grep ... ``` line tells the shell that the grep is a different command, so the shell will execute ```python 3 <filename>.py cat sep-quarter.csv ``` and then ``` grep '.SEA1AA' ``` separately, which naturally yields an incorrect result for the df.

```python3 <filename.py>  'unzip -cq sep-quarter.csv.zip' ``` (in quotes for the same reason as above): Will unzip the .csv file before creating a df out of the contents within. This creates an identical df to the first command I mentioned.

I have included the .csv files in the branch, both zipped and unzipped. Those can be moved deleted at will, they're just there for testing.

**This PR lays the foundation for resolving #945 **

# Relevant Reviewers

- @ericmjl

### CODE REVIEW

File: janitor/functions/read_commandline.py [hunk 1]
```diff
@@ -0,0 +1,77 @@
+import pandas_flavor as pf
+import pandas as pd
+from subprocess import Popen, PIPE
+import sys
+
+
+@pf.register_dataframe_method
```
ericmjl: I think this line isn't necessary, because the intent of `read_commandline` is to be a function that operates outside of a dataframe.

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+def read_commandline() -> pd.DataFrame:
+    """
+    Read command from commandline.
```
ericmjl:
```suggestion
    Read a CSV file based on a command-line command.
```

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
+import pandas_flavor as pf
+import pandas as pd
+from subprocess import Popen, PIPE
+import sys
+
+
+@pf.register_dataframe_method
+def read_commandline() -> pd.DataFrame:
```
ericmjl:
```suggestion
def read_commandline(cmd) -> pd.DataFrame:
```

The `cmd` should be the shell command that is desired.

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    Read command from commandline.
+
+    argv[0] should be program name
+    argv[1:] will be the a list of commands available to the shell
+    such as:
+    cat sep-quarter.csv | grep .SEA1AA
```
ericmjl:
```suggestion
    For example, you may wish to run the following command on `sep-quarter.csv`
    before reading it into a pandas DataFrame:
    
    ```bash
    cat sep-quarter.csv | grep .SEA1AA
    ```
```

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+
+    calling
+    df = read_commandline()
+    will create a new dataframe from the commandline arguments given.
+
+    :returns: a dataframe which has been created based on the arguments
```
ericmjl:
```suggestion
    :param cmd: Shell command to preprocess a file on disk.
    :returns: A pandas DataFrame parsed from the stdout of the underlying shell.
```

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+    Read command from commandline.
+
+    argv[0] should be program name
+    argv[1:] will be the a list of commands available to the shell
+    such as:
+    cat sep-quarter.csv | grep .SEA1AA
+
+    calling
+    df = read_commandline()
+    will create a new dataframe from the commandline arguments given.
```
ericmjl:
```suggestion
    In this case, you can use the following Python code to load the dataframe:
    
    ```python
    import janitor as jn
    df = jn.read_commandline("cat sep-quarter.csv | grep .SEA1AA")
    ```
```

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+
+    :returns: a dataframe which has been created based on the arguments
+        given in the commandline
+    """
+    command = __preprocess_command()
```
ericmjl:
```suggestion
    cmd = __preprocess_command(cmd)
```

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    command = ""
+    for word in sys.argv[1:]:  # must be a single string to be used in Popen
+        command += word + " "
+    command = str(command)
+    print(command)
+    return command
+
+
+def query_df(command: str) -> pd.DataFrame:
```
ericmjl: Friendly reminder doctoring required! :smile:

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    Transforms argv[1:] into a single string
+    which is more easily parsed by PIPE and Popen above
+    """
+    command = ""
+    for word in sys.argv[1:]:  # must be a single string to be used in Popen
+        command += word + " "
+    command = str(command)
+    print(command)
+    return command
```
ericmjl: I think you can avoid this function altogether.

janitor/functions/read_commandline.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def query_df(command: str) -> pd.DataFrame:
+    with Popen(command, shell=True, stdout=PIPE) as process:
```
ericmjl:
```suggestion
    command = command.split(" ")
    with Popen(command, shell=True, stdout=PIPE) as process:
```

The added line above will ensure that the command is split into a string.

ericmjl (changes requested):
@BaritoneBeard this is a good start. Thank you for taking a stab at the PR!

Architecturally, I think there's some room for improvement here. Some ideas you could chew on:

1. `read_commandline` should accept a string command that we simply shell out using `Popen`.
2. Treat the `read_commandline` function as if it were being executed from a Jupyter notebook, rather than a thing executed at the shell.

I added additional comments that should help illuminate where I think the PR should head.

For automated testing, I'd suggest the following.

1. Using Python code, write a very small dummy dataframe to disk at a temp directory, e.g. `/tmp/data.csv`. You can use one of the data frames that are being added to the docstrings' minimum workable examples (I'm thinking of @thatlittleboy's work thus far).
2. Have `read_commandline` read in the result of `cat /tmp/data.csv`. This simple command should work.
3. Then, do something perhaps a bit more complicated, such as grep-ing 2 or more lines from the table, and then reading that as a dataframe.

Looking forward to seeing the changes! Let me know if there's anything unclear, happy to clarify.

Down below, I riffed off what you had provided and made direct suggestions in there. If you accept them all and then read the code afterward, you might find it clearer to understand everything in context than if you tried to understand each line change independent of one another.

samukweku (commented):
thanks @BaritoneBeard for this; I feel it is a bit different from what I think the shell command should do - this is based on my observation from the datatable `fread` library that I am emulating. The idea is to pass a shell command string, and let python's subprocess worry about the rest. the output will then be read into pandas as a string. Here is a basic template: 

```py
import subprocess
from io import StringIO
def read_command_line(cmd: str):
    try:
        outcome = subprocess.run(cmd, shell = True, capture_output = True, text = True)
        outcome = outcome.stdout
   except:
        raise
   return pd.read_csv(StringIO(outcome))
```

no need for the preprocessing, or any other prework per se (user is responsible, and an appropriate error message should pop if it doesnt work out). It also, I hope, allows us to be able to maintain it (I am not a linux master :grin:). 

Here is an example usage - [example](https://samukweku.github.io/data-wrangling-blog/python/pandas/compressed%20data/zip/archived%20data/2020/07/21/Extract-DataFrame-from-Compressed-Data-into-Pandas.html#Read-Compressed-Data-through-the-Command-Line).

we can work together on the tests; dont think it'd be too hard (took me a while to get the hang of testing though ... pyjanitor helped me get to where I am :grin:)

Also, I think this should go into the `janitor.io` submodule

Hope my comment makes some sense; let me know what you think.

Nalmeder:
Heard, thank you both.

I'll accept some or all changes suggested by @ericmjl and get to work refactoring to look more in line with what @samukweku has laid out. A word of 'warning' is that I tend to work slowly on stuff I'm not being paid for (because I'm tired from working on stuff I do get paid for), I'll probably take a few days for this.

ericmjl:
Not a problem! We’re all volunteers here on this project, so it’s
totally reasonable :). 

Can I check, did you enable allowing maintainers to push to the PR
branch? If so, I might help move some things along, though I’m also busy
with another PR to the repo.

Nalmeder:
According to the little check box, moderators are allowed access and edit, so they should be allowed to push, correct?

ericmjl:
I think that's right! Thank you for confirming, @BaritoneBeard!

Nalmeder:
@samukweku  I think I'm missing something. Because I don't use pandas (I'm here for that sweet, sweet open source contribution experience, so I only know the bare basics in order to test that my stuff works) I'm not really sure of what exactly it is that's missing. The template you supplied passes the manual tests: it can make a df out of the commands I tested earlier in the pr -
```
#%%
df = read_commandline('cat sep-quarter.csv | grep BDCQ.SEA1AA')

df
```
makes a df of 41 rows as before when the cell is run.

if this is all you needed the function to do, then it looks like you didn't even need me, as the template did all the work. 

On Testing: Pytest is giving me trouble despite trying my usual methods of making it work (pytest seems to be very fickle, it breaks every time I fork a repo). I've written plenty of tests before, so I can try to write some blind tests. Alternatively, my other test suite Behave would probably work for this and I can write some tests in there and translate those working tests into pytest ones, but again those translated tests would be blind. Should I give some blind testing a go?

ericmjl:
> if this is all you needed the function to do, then it looks like you didn't even need me, as the template did all the work.

@BaritoneBeard no, not at all! Your contribution is highly valued, even if the final PR ends up using something very close to what we've provided. This is because it takes care, time, and effort to not only implement the function but also to document it and test it sufficiently for others to use :smile:. Plus, it's also a great learning opportunity for newcomers who desire to broaden their software skills.

As for why the CI is failing, in this case, it's because the test coverage on the newly added code is below the target that we wish to hit. We measure test line coverage percentage and require it to be above a certain number as a first gatekeeping measure to ensure that there is high test coverage on the codebase. (That doesn't ensure the quality of tests, which is another matter, but on thoughtfully maintained repos, quality and percentage are correlated.) 

To get the PR over the line, I think we need at least one software test. Here are a few ways we can proceed.

Firstly, can you move the newly added function into the `janitor.io` submodule? This is because it's a function that handles data I/O, rather than a data preprocessing function, akin to `read_csvs` in there.

Secondly, can you then add a test for `read_commandline`, perhaps mimicking [this file](https://github.com/pyjanitor-devs/pyjanitor/blob/dev/tests/io/test_read_csvs.py)? 

Thirdly, can you remove the files `seq-quarter.csv` and `sep-quarter.zip` from the PR? We can create the test data on the fly rather than rely on pre-created data by using some of the provided test fixtures that we have. The pattern you can use for writing the test is as follows:

```python
def test_read_commandline(dataframe):
    # set up the test
    dataframe.to_csv("/tmp/dataframe.csv")
    df = janitor.read_commandline("cat /tmp/dataframe.csv")
    # make assertion in the test
    assert pd.assert_frame_equal(dataframe, df)
    # clean up after the test
    Path("/tmp/dataframe.csv").unlink()
```

That test uses a magical variable, `dataframe`, that is provided as a "test fixture" for all of our tests. There are other magically available dataframes that you can use, they are discoverable [here](https://github.com/pyjanitor-devs/pyjanitor/blob/dev/tests/conftest.py), and basically only need to be placed in the test function signature for the test body to be aware of them.

Nalmeder:
@ericmjl 
First, thanks for the reaffirming words, the reason I came back to try my hand at another PR here was because you all were so welcoming the first time, so it's good to see nothing's changed on that front.

> Firstly, can you move the newly added function into the janitor.io submodule? 

Right, forgot to do that, consider it done. 

> Secondly, can you then add a test for read_commandline, 

No problem, ~~as I said though, my pytest is on the fritz and I have no way of making sure the tests work or pass their assertions. I can *write* the tests but that's about it. Do reviewers get emails every time I push? I know the workflow runs all tests each push so worst case I can push tests if it's not an inconvenience to anybody else but myself. Before that though I'll obviously see if I can get pytest working with a pair of fresh eyes tonight.~~

Nevermind all of that, about 6 seconds after I hit send I found a solution. Every time. The test will be included in the next push.

>Thirdly, can you remove the files seq-quarter.csv and sep-quarter.zip from the PR? 

Absolutely. I planned to remove them before the final push anyway. Thanks for highlighting how to better use the tests available.

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+    # cmd = cmd.split(" ")
+    try:
+        outcome = subprocess.run(
+            cmd, shell=True, capture_output=True, text=True
+        )
+        outcome = outcome.stdout
+    except pd.EmptyDataError:
+        msg = (
+            "Empty Data Error: Be sure your parameter"
+            " is both a valid shell command and a string"
```
ericmjl: I think the error message could be improved here! What do you think of the following instead?

"It appears your DataFrame was loaded incorrectly. Please check the command string at the terminal to ensure that it returns a CSV-like output."

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+    cat sep-quarter.csv | grep .SEA1AA
+    ```
+
+    In this case, you can use the following Python code to load the dataframe:
+
+    ```python
+    import janitor as jn
+    df = jn.read_commandline("cat sep-quarter.csv | grep .SEA1AA")
+    ```
+
```
ericmjl, line 98: In reviewing the docstring, I think it is important to communicate to end-users the following:

"This function assumes that your command line command will return an output that is parsable using pandas.read_csv and StringIO. We default to using pd.read_csv underneath the hood. Keyword arguments are passed through to read_csv."

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
         )
     else:
         return dfs_dict
+
+
+def read_commandline(cmd: str) -> pd.DataFrame:
```
ericmjl:
```suggestion
def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
```

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+
+    In this case, you can use the following Python code to load the dataframe:
+
+    ```python
+    import janitor as jn
+    df = jn.read_commandline("cat sep-quarter.csv | grep .SEA1AA")
+    ```
+
+    :param cmd: Shell command to preprocess a file on disk.
+    :returns: A pandas DataFrame parsed from the stdout of the underlying
```
ericmjl, line 102:
```suggestion
    :param kwargs: Keyword arguments that are passed through to `pd.read_csv()`.
    :returns: A pandas DataFrame parsed from the stdout of the underlying
```

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+        )
+        outcome = outcome.stdout
+    except pd.EmptyDataError:
+        msg = (
+            "Empty Data Error: Be sure your parameter"
+            " is both a valid shell command and a string"
+        )
+        raise pd.EmptyDataError(msg)
+
+    return pd.read_csv(StringIO(outcome))
```
ericmjl:
```suggestion
    return pd.read_csv(StringIO(outcome), **kwargs)
```

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
...
+    # create a new dataframe from the temporary .csv using
+    #   the cat command from the bash commandline
+    df = janitor.io.read_commandline("cat /tmp/dataframe.csv")
+
+    # The user may be expected to process some of the df themself
+    df.drop(df.columns[[0]], axis=1, inplace=True)
+
+    # Make assertion dataframe is equal to itself
+    assert not pd.testing.assert_frame_equal(dataframe, dataframe)
+    assert dataframe.equals(dataframe)
```
ericmjl: I think this assertion is unnecessary and we can remove it. When testing, we only need to assert the reliability of our own code, and don't need to assert the reliability of underlying code, as that is a responsibility we can delegate to the underlying library maintainers. If we uncover a problem in the underlying library, we can raise an issue instead.

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
...
+    test data is equal to itself. Note the
+    requirement of 'assert not' for
+    assert_frame_equal.
+
+Tests 3 and 4 assert that the dataframe made
+    from the read_commandline function is
+    identical to the test dataframe from
+    which the .csv file was created.
+
+"""
```
ericmjl: This should go into the test function as a docstring, rather than as a string outside of the function. 

Additionally, I think it can be shortened. It seems like only tests 3 and 4 are necessary.

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
+import os
```
ericmjl, line 1: We should have a module-level docstring here.

```suggestion
"""Tests for read_commandline."""
import os
```

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
...
+def test_read_commandline(dataframe):
+    # create a temporary .csv file from test data
+    dataframe.to_csv("/tmp/dataframe.csv")
+
+    # create a new dataframe from the temporary .csv using
+    #   the cat command from the bash commandline
+    df = janitor.io.read_commandline("cat /tmp/dataframe.csv")
+
+    # The user may be expected to process some of the df themself
+    df.drop(df.columns[[0]], axis=1, inplace=True)
```
ericmjl: With `**kwargs` pass-through, we can remove this data-processing line.

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+
+def test_read_commandline(dataframe):
+    # create a temporary .csv file from test data
+    dataframe.to_csv("/tmp/dataframe.csv")
+
+    # create a new dataframe from the temporary .csv using
+    #   the cat command from the bash commandline
+    df = janitor.io.read_commandline("cat /tmp/dataframe.csv")
```
ericmjl, line 23: With `**kwargs` pass-through, we can do this:

```suggestion
    df = janitor.io.read_commandline("cat /tmp/dataframe.csv", index_col=0)
```

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
...
+    # The user may be expected to process some of the df themself
+    df.drop(df.columns[[0]], axis=1, inplace=True)
+
+    # Make assertion dataframe is equal to itself
+    assert not pd.testing.assert_frame_equal(dataframe, dataframe)
+    assert dataframe.equals(dataframe)
+
+    # Make assertion that new dataframe created with read_commandline
+    #   is equal to the test dataframe
+    assert not pd.testing.assert_frame_equal(dataframe, df)
```
ericmjl: Something isn't sitting right here for me - shouldn't it be the case that `pd.testing.assert_frame_equal(dataframe, df)` should evaluate to `True`? Why do we assert that it is `False`?

ericmjl (changes requested):
We're almost there, @BaritoneBeard! I've had a chance to more carefully review the tests, now that the function implementation is much cleaner, and left my comments and questions below.

The summary of my suggestions are:

1. Pass-through `**kwargs` to `pd.read_csv()`.
2. Simplify the test a bit.
3. Add a test for error handling.

On number 3, a pattern you can use looks like this:

```python
def test_read_commandline_bad_cmd(dataframe):
    """<Fill in docstring here>"""
    dataframe.to_csv("/tmp/dataframe.csv")
    with pytest.raises(pd.EmptyDataError):
        df = janitor.io.read_commandline("blah")
```

That aforementioned test will check that a `pd.EmptyDataError` is raised when a bad command is passed in. Adding in that test will also bump up the test coverage to 100% of the added code, which CodeCov will be happy about :smile:.

Great work thus far, thanks for the time and effort put into this!

janitor/io.py
```diff
@@ -1,11 +1,13 @@
 import os
+import subprocess
 from glob import glob
 from typing import Iterable, Union
 
 import pandas as pd
 
 from .errors import JanitorError
 from .utils import deprecated_alias
+from io import StringIO
```
thatlittleboy: The standard practice is to structure the imports with Python standard library modules (of which `io` is part of) at the top.

Just as a quick tip, you can just run `isort` on this file to get it fixed automatically. Else, you can just manually move this line to the line under the "glob" import.

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+
+    ```bash
+    cat sep-quarter.csv | grep .SEA1AA
+    ```
+
+    In this case, you can use the following Python code to load the dataframe:
+
+    ```python
+    import janitor as jn
+    df = jn.read_commandline("cat sep-quarter.csv | grep .SEA1AA")
```
thatlittleboy: Gentle reminder to also update these lines to mimic actual usage of the function. In particular, this `read_commandline` function is not available at the global module level.

So the correct way to use this function at the moment is `janitor.io.read_commandline("...")`. You've got it right in the tests, just need to update it here as well. :)

(On a related note, I'm also wondering why the `io` & a few other modules' functions are left out in the `janitor/__init__.py` file. 😅)

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+        nothing happens in the shell , or if cmd param is not a string,
+        thus creating an invalid shell command.
+    """
+    # cmd = cmd.split(" ")
+    try:
+        outcome = subprocess.run(
+            cmd, shell=True, capture_output=True, text=True
+        )
+        outcome = outcome.stdout
+    except pd.EmptyDataError:
```
thatlittleboy: Could you explain when such an exception will occur? I doubt it will, right? Since there is nothing related to `pandas` in the try block. I'm guessing this is leftover from previous code. Or possibly you meant to have the `pd.read_csv` inside the try block. (btw, there is no `pd.EmptyDataError`, it's `pd.errors.EmptyDataError`.)

Furthermore, I'm thinking whether it is a better user experience if we can also propagate the stderr back to the user IF return_code is non-zero, so the user explicitly knows that the error is because of the shell command being malformed etc., instead of a pandas read error.

---

Also, I'm not convinced that we should even be catching empty data error; that is functionality that is passed to pandas, we can simply let pandas throw the error in a transparent manner, IMO.

For example:
```python
>>> janitor.io.read_commandline("")
```
image

I think this stacktrace is clear enough, we shouldn't need to catch and re-raise the exception?

janitor/io.py
```diff
@@ -69,3 +71,45 @@ def read_csvs(
...
+    df = jn.read_commandline("cat sep-quarter.csv | grep .SEA1AA")
+    ```
+
+    :param cmd: Shell command to preprocess a file on disk.
+    :returns: A pandas DataFrame parsed from the stdout of the underlying
+        shell.
+    :raises EmptyDataError: If there is no data to parse, this often happens
+        because the cmd param is either an invalid bash command, thus
+        nothing happens in the shell , or if cmd param is not a string,
+        thus creating an invalid shell command.
```
thatlittleboy: "if cmd param is not a string" -> I think we can use the `check` function from `utils` to check the type of `cmd` and enforce this to be a string type.
Since,`read_commandline(6)` will throw a TypeError coming from the `subprocess.run` function call (aside: I think this is worth adding a test for as well).

This is a case where I think it is important to "mask" the downstream error from the end-user. If the user does not pass in a string, we should raise a TypeError of our own to prompt the user to pass in a string.

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,48 @@
...
+    from the read_commandline function is
+    identical to the test dataframe from
+    which the .csv file was created.
+
+"""
+
+
+def test_read_commandline(dataframe):
+    # create a temporary .csv file from test data
+    dataframe.to_csv("/tmp/dataframe.csv")
```
thatlittleboy: maybe consider adding an `index=False` here? I'm unsure whether this is the reason for you dropping `columns[0]` below or not.

thatlittleboy (changes requested):
Thanks for the effort thus far @BaritoneBeard !
I'll just jump in here with a quick review based on what I've seen, thought it's better to give feedback earlier than later. :)

tests/io/test_read_commandline.py
Nalmeder: I agree, I was going to leave out ```assert_frame_equal``` altogether but wanted to ask you or anybody else what they thought was going on here. In the code, ```pd.testing.assert_frame_equal(dataframe, dataframe) ``` evaluates to false, which shouldn't happen since that line would be testing a dataframe against itself, therefore it should be identical and thus evaluate to true. It does not. thoughts there?

```.equals()``` works as expected, though thats not a pandas function.

File: janitor/io.py (hunk 24 above)
Comment (Nalmeder) [reply janitor/io.py]: That was my thought too, I think Pandas throws an error before it even finishes parsing through the try block, which would indeed do nothing more than re-raising an exception that was already raised if that code is left in.

File: janitor/io.py (hunk 25 above)
Comment (ericmjl) [reply janitor/io.py]: Great point, @thatlittleboy! I agree.

**Comment by Nalmeder** [comment]
@ericmjl  All green here, let me know if anything else jumps out.

@thatlittleboy I looked around a little bit for docs on python utils ```check``` function but couldn't find any, I know it exists but I guess the word "check" is a little too ubiquitous to reliably get search or ctrl-F results. I implemented raising TypeError without it though. If you feel the PR would benefit from using ```check``` over/in addition to what's provided then would you mind pointing me in the direction of some docs? Thanks!

File: janitor/io.py [hunk 27]
```diff
@@ -63,9 +65,68 @@ def read_csvs(
...
+    :param kwargs: Keyword arguments that are passed through to pd.read_csv().
+    :returns: A pandas DataFrame parsed from the stdout of the underlying
+        shell.
+    :raises TypeError: If the input is anything except a string.
+    :raises EmptyDataError: If there is no data to parse, this often happens
+        because the cmd param is an invalid bash command,
+        thus nothing happens in the shell.
+    """
+    # cmd = cmd.split(" ")
+
```
samukweku, line 105: the subprocess will return an error if the return code is non zero, since capture_output is set to True. so maybe we should use that instead (agreeing with @thatlittleboy ):

```py
from .utils import check # at the top
check('cmd', cmd, str)
outcome = subprocess.run(
            cmd, shell=True, capture_output=True, text=True
        )
```
removes the need for the try clause for the shell section.

janitor/io.py
```diff
@@ -63,9 +65,68 @@ def read_csvs(
...
+    except TypeError:
+        msg = (
+            "It appears the command was not a string."
+            " Please check that the input is a string."
+        )
+        raise TypeError(msg)
+
+    try:
+        df = pd.read_csv(StringIO(outcome), **kwargs)
+    except pd.errors.EmptyDataError:
```
samukweku: not so sure about this - is it possible that there'd be other errors apart from EmptyDataError? should we just leave this and raise whatever error pandas returns? so instead of the try except, just `df = pd.read_csv(StringIO(outcome), **kwargs)` and if there are any errors, pandas will loudly scream it? (loud is good :grin:)

samukweku: the **kwargs idea is awesome !

samukweku:
@BaritoneBeard the `check` function is the first in the utils section [here](https://github.com/pyjanitor-devs/pyjanitor/blob/dev/janitor/utils.py)

janitor/io.py
Nalmeder: I don't usually like excepting just anything that comes up, but in this case where the user may input just about any shell command I think you make a good point, I'll push the changes.

Nalmeder:
@samukweku Your suggestions have been implemented.

(By the way I was looking for check in python's utils instead of janitors, I guess I need a nap. Thanks for the assist).

janitor/io.py
```diff
@@ -63,9 +65,45 @@ def read_csvs(
...
+    :param cmd: Shell command to preprocess a file on disk.
+    :param kwargs: Keyword arguments that are passed through to pd.read_csv().
+    :returns: A pandas DataFrame parsed from the stdout of the underlying
+        shell.
+    """
+
+    check("cmd", cmd, [str])
+    outcome = subprocess.run(cmd, shell=True, capture_output=True, text=True)
+    outcome = outcome.stdout
+    df = pd.read_csv(StringIO(outcome), **kwargs)
```
samukweku: Petty me. I don't think there is a need to assign to DF here , we can just return `pd.read_csv...` without assigning to a variable

samukweku (approved):

tests/io/test_read_commandline.py
ericmjl: Hmmm, now I'm remembering. 

`pd.testing.assert_frame_equal(dataframe, dataframe)` doesn't return anything. It is intended to be used as a standalone statement. Example usage would be:

```python
pd.testing.assert_frame_equal(dataframe, df)
```

That line would be semantically identical to `assert dataframe.equals(df)`, I think. In any case, though, using the pandas built-in testing tooling is probably more stylistically idiomatic than using `assert df.equals(other_df)`.

janitor/io.py
ericmjl: Hmmm, that's a good point, @thatlittleboy, we probably can make the `io` submodule's contents exposed in the top-level.

janitor/io.py
```diff
@@ -63,9 +65,44 @@ def read_csvs(
...
+
+    ```python
+    import janitor as jn
+    df = jn.io.read_commandline("cat data.csv | grep .SEA1AA")
+
+    This function assumes that your command line command will return
+    an output that is parsable using pandas.read_csv and StringIO.
+    We default to using pd.read_csv underneath the hood.
+    Keyword arguments are passed through to read_csv.
+    ```
```
ericmjl, line 97: Formatting change:

```suggestion
    ```python
    import janitor as jn
    df = jn.io.read_commandline("cat data.csv | grep .SEA1AA")
    ```

    This function assumes that your command line command will return
    an output that is parsable using pandas.read_csv and StringIO.
    We default to using pd.read_csv underneath the hood.
    Keyword arguments are passed through to read_csv.

```

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,53 @@
...
+
+import janitor.io
+
+
+def test_read_commandline(dataframe):
+    """
+    Test asserts that the dataframe made
+        from the read_commandline function is
+        identical to the test dataframe from
+        which the .csv file was created.
```
ericmjl, line 15:
```suggestion
    Check that the dataframe returned from the read_commandline function is
    identical to the test dataframe from which the .csv file was created.
```

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,51 @@
...
+    # create a temporary .csv file from test data
+    dataframe.to_csv("/tmp/dataframe.csv", index=0)
+
+    # create a new dataframe from the temporary .csv using
+    #   the cat command from the bash commandline
+    df = janitor.io.read_commandline("cat /tmp/dataframe.csv")
+
+    # Make assertion that new dataframe created with read_commandline
+    #   is equal to the test dataframe
+    assert df.equals(dataframe)
```
ericmjl, line 27:
```suggestion
    pd.testing.assert_frame_equal(df, dataframe)
```

tests/io/test_read_commandline.py
```diff
@@ -0,0 +1,51 @@
...
+def test_read_commandline_bad_cmd(dataframe):
+    """
+    Test 1 raises a TypeError if read_commandline
+        is given an input that is not a string.
+
+    Test 2 raises an EmptyDataError if
+        read_commandline is given a string
+        which is not a valid bash command.
+        This results in the shell not doing anything
+        and thus no dataframe is created.
```
ericmjl:
```suggestion
    Test 1 raises a TypeError if read_commandline
    is given an input that is not a string.

    Test 2 raises an EmptyDataError if
    `read_commandline` is given a string
    which is not a valid bash command.
    This results in the shell not doing anything
    and thus no dataframe is created.
```

ericmjl (approved):

janitor/io.py
thatlittleboy: Okay sure, let's do this as a separate PR then. Will open an issue so we don't forget.

janitor/io.py
```diff
@@ -63,9 +65,45 @@ def read_csvs(
...
+
+    :param cmd: Shell command to preprocess a file on disk.
+    :param kwargs: Keyword arguments that are passed through to pd.read_csv().
+    :returns: A pandas DataFrame parsed from the stdout of the underlying
+        shell.
+    """
+
+    check("cmd", cmd, [str])
+    outcome = subprocess.run(cmd, shell=True, capture_output=True, text=True)
+    outcome = outcome.stdout
```
thatlittleboy: Just wanted to bring this to attention again for discussion: I would prefer if we could not proceed with `pd.read_csv` if the return_code is non-zero.

Basically, there are at least 2 "kinds" of errors that can happen here, either the user enters a malformed shell cmd or the shell cmd is valid but the data itself is problematic (whether it is empty, or missing a comma, etc.) for `pd.read_csv` to parse.
Example of 1: `jn.io.read_commandline("catt data.csv")`
Example of 2: `jn.io.read_commandline("touch data.csv && cat data.csv")`

I am of the opinion that these 2 types of errors should not be categorized together. In particular, the former will (currently) lead to a pandas EmptyDataError, which is not actually the crux of the problem. I would very much prefer to catch the non-zero code, and raise a `JanitorError` containing the shell stderr.

thatlittleboy (changes requested):
I just have 1 final gripe about the handling of errors, but please, I don't want to be the blocker of this PR.
@BaritoneBeard has done great work here. If the others are agreeable with the current code state, I'm happy to be outvoted and proceed with the merging.

janitor/io.py
samukweku: @thatlittleboy if the data has a non-zero code, the std error is outputted (since `capture_output` is set to True). for the other case where the data itself is problematic, I'd like to think that `pd.read_csv` should output the error. 

I probably misunderstood your comment; maybe an example with the current code could be helpful.

thatlittleboy: Sure, apologies for being unclear. Pertaining to error type1, at the moment:

```shell
In [13]: janitor.io.read_commandline("echoo 'a,b\n1,2'")
---------------------------------------------------------------------------
EmptyDataError                            Traceback (most recent call last)
<ipython-input-13-12d5880211fb> in <module>
----> 1 janitor.io.read_commandline("echoo 'a,b\n1,2'")

/workspaces/pyjanitor/janitor/io.py in read_commandline(cmd, **kwargs)
    107     outcome = subprocess.run(cmd, shell=True, capture_output=True, text=True)
    108     outcome = outcome.stdout
--> 109     return pd.read_csv(StringIO(outcome), **kwargs)

/opt/conda/lib/python3.7/site-packages/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    309                     stacklevel=stacklevel,
    310                 )
--> 311             return func(*args, **kwargs)
    312 
    313         return wrapper

/opt/conda/lib/python3.7/site-packages/pandas/io/parsers/readers.py in read_csv(filepath_or_buffer, sep, delimiter, header, names, index_col, usecols, squeeze, prefix, mangle_dupe_cols, dtype, engine, converters, true_values, false_values, skipinitialspace, skiprows, skipfooter, nrows, na_values, keep_default_na, na_filter, verbose, skip_blank_lines, parse_dates, infer_datetime_format, keep_date_col, date_parser, dayfirst, cache_dates, iterator, chunksize, compression, thousands, decimal, lineterminator, quotechar, quoting, doublequote, escapechar, comment, encoding, encoding_errors, dialect, error_bad_lines, warn_bad_lines, on_bad_lines, delim_whitespace, low_memory, memory_map, float_precision, storage_options)
    584     kwds.update(kwds_defaults)
    585 
--> 586     return _read(filepath_or_buffer, kwds)
    587 
    588 

/opt/conda/lib/python3.7/site-packages/pandas/io/parsers/readers.py in _read(filepath_or_buffer, kwds)
    480 
    481     # Create the parser.
--> 482     parser = TextFileReader(filepath_or_buffer, **kwds)
    483 
    484     if chunksize or iterator:

/opt/conda/lib/python3.7/site-packages/pandas/io/parsers/readers.py in __init__(self, f, engine, **kwds)
    809             self.options["has_index_names"] = kwds["has_index_names"]
    810 
--> 811         self._engine = self._make_engine(self.engine)
    812 
    813     def close(self):

/opt/conda/lib/python3.7/site-packages/pandas/io/parsers/readers.py in _make_engine(self, engine)
   1038             )
   1039         # error: Too many arguments for "ParserBase"
-> 1040         return mapping[engine](self.f, **self.options)  # type: ignore[call-arg]
   1041 
   1042     def _failover_to_python(self):

/opt/conda/lib/python3.7/site-packages/pandas/io/parsers/c_parser_wrapper.py in __init__(self, src, **kwds)
     67         kwds["dtype"] = ensure_dtype_objs(kwds.get("dtype", None))
     68         try:
---> 69             self._reader = parsers.TextReader(self.handles.handle, **kwds)
     70         except Exception:
     71             self.handles.close()

/opt/conda/lib/python3.7/site-packages/pandas/_libs/parsers.pyx in pandas._libs.parsers.TextReader.__cinit__()

EmptyDataError: No columns to parse from file
```

so from the perspective of the end user, the problem is empty data, but the actual error is hidden away in the stderr

```shell
In [15]: subprocess.run("echoo 'a,b\n1,2'", shell=True, capture_output=True, text=True)
Out[15]: CompletedProcess(args="echoo 'a,b\n1,2'", returncode=127, stdout='', stderr='/bin/sh: 1: echoo: not found\n')
```

Unless I'm missing something, while stderr is captured with `capture_output`, it's not actually visible to the end user?

samukweku: totally clear now; @BaritoneBeard my apologies for misleading you on that. so @BaritoneBeard @thatlittleboy something like this: 
```py
 outcome = subprocess.run(cmd, shell=True, capture_output=True, text=True)
if outcome.returncode != 0:
    raise ValueError(outcome.stderr) # or JanitorError
else:
     outcome = outcome.stdout
```

?

ericmjl: Ahhh, now I get it. @thatlittleboy, what you've suggested sounds like the saner thing to do. Thanks @samukweku for asking to clarify!

@BaritoneBeard, could you make this change, and do one final thing -- add this contribution to the CHANGELOG.md file please?

Nalmeder:
@ericmjl Updated the changelog.

@thatlittleboy Added the error catching you and @samukweku discussed. I also added a test for it because that just seemed like the obvious thing to do. I opted to go with a Janitor Error arbitrarily, it just felt right since the error is being found within janitor.io rather than pandas. 

I had some issues and had to merge/revert here and there, so @samukweku pet peeve made it's way back in undiscovered until now(```df = pd...  return df ``` instead of just ``` return pd...```), but once I fix that again, is there anything else that jumps out at anybody?

Update: Looks like some tests failed, none of which were mine, I don't think. Is there something I should do about these?

ericmjl:
@BaritoneBeard they are all spark errors. In #986, I asked @thatlittleboy to mark them all as xfail; they are very flaky. Once that's merged in, we'll have you merge in the latest `dev` commits and the tests will re-run; I'm confident they'll pass.

janitor/io.py
```diff
@@ -63,9 +65,49 @@ def read_csvs(
...
+
+    :param cmd: Shell command to preprocess a file on disk.
+    :param kwargs: Keyword arguments that are passed through to pd.read_csv().
+    :raises JanitorError: If commandline command is malformed or invalid.
+    :returns: A pandas DataFrame parsed from the stdout of the underlying
+        shell.
+    """
+
+    check("cmd", cmd, [str])
+    outcome = subprocess.run(cmd, shell=True, capture_output=True, text=True)
```
samukweku, line 107: @BaritoneBeard it seems you have this line repeated twice.

thatlittleboy (approved):

ericmjl:
Hitting da merge button!

Issue #972, opened 2022-01-02
[DOC] Adding minimal working examples to docstrings; a checklist


thatlittleboy:
## Background

This thread is borne out of the discussion from #968 , in an effort to make documentation more beginner-friendly & more understandable.
One of the subtasks mentioned in that thread was to go through the function docstrings and include a *minimal* working example to each of the public functions in pyjanitor.

Criteria reiterated here for the benefit of discussion:

>     It should fit with our existing choice to go with mkdocs, mkdocstrings, and mknotebooks.
>     The examples should be minimal and executable and complete execution within 5 seconds per function.
>     The examples should display in rich HTML on our docs page.
>     We should have an automatic way of identifying whether a function has an example provided or not so that every function has an example.

Sample of what MWE should look like is shown [here](https://github.com/pyjanitor-devs/pyjanitor/issues/968#issuecomment-1003672331).

---

I'm thinking we can create a task list so that 1. we can encourage more users to join in the effort, and 2. make sure we don't do duplicate work. A lot of the groundwork can be covered by selectively copying one or two examples over from the software test suite.

Then we can label this issue as a Help Wanted / Low-Hanging Fruit and get people to mention in this thread if they're intending to work on the files?

### Task list

- [X] functions/add_columns.py
- [x] functions/also.py
- [x] functions/bin_numeric.py
- [x] functions/case_when.py
- [x] functions/change_type.py
- [x] functions/clean_names.py
- [x] functions/coalesce.py
- [x] functions/collapse_levels.py
- [x] functions/complete.py
- [x] functions/concatenate_columns.py
- [x] functions/conditional_join.py
- [x] functions/convert_date.py
- [x] functions/count_cumulative_unique.py
- [x] functions/currency_column_to_numeric.py
- [x] functions/deconcatenate_column.py
- [x] functions/drop_constant_columns.py
- [x] functions/drop_duplicate_columns.py
- [x] functions/dropnotnull.py
- [x] functions/encode_categorical.py
- [x] functions/expand_column.py
- [x] functions/expand_grid.py
- [x] functions/factorize_columns.py
- [x] functions/fill.py
- [x] functions/filter.py
- [x] functions/find_replace.py
- [x] functions/flag_nulls.py
- [x] functions/get_dupes.py
- [x] functions/groupby_agg.py
- [x] functions/groupby_topk.py
- [x] functions/impute.py
- [x] functions/jitter.py
- [x] functions/join_apply.py
[...]

PR #978, opened 2022-01-07
[DOC] Add minimum working examples to clean_names and case_when

ericmjl:
This PR adds minimum working examples to two functions, `clean_names` and `case_when`.

Related to: #972.

### DIFF
--- a/janitor/functions/case_when.py
+++ b/janitor/functions/case_when.py
@@ -17,12 +17,12 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
         >>> import pandas as pd
         >>> import janitor
         >>> df = pd.DataFrame(
-        ...         {
-        ...             "a": [0, 0, 1, 2, "hi"],
-        ...             "b": [0, 3, 4, 5, "bye"],
-        ...             "c": [6, 7, 8, 9, "wait"],
-        ...         }
-        ...     )
+        ...     {
+        ...         "a": [0, 0, 1, 2, "hi"],
+        ...         "b": [0, 3, 4, 5, "bye"],
+        ...         "c": [6, 7, 8, 9, "wait"],
+        ...     }
+        ... )
         >>> df
             a    b     c
         0   0    0     6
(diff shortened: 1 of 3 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-978--pyjanitor.netlify.app

janitor/functions/case_when.py
```diff
@@ -10,10 +10,40 @@
...
+
+        >>> import pandas as pd
+        >>> import janitor
+        >>> df = pd.DataFrame(
+        ...         {
+        ...             "a": [0, 0, 1, 2, "hi"],
+        ...             "b": [0, 3, 4, 5, "bye"],
+        ...             "c": [6, 7, 8, 9, "wait"],
+        ...         }
+        ...     )
```
thatlittleboy: The indentation for the braces and the final parenthesis are off.

janitor/functions/case_when.py
```diff
@@ -10,10 +10,40 @@
...
+        0   0    0     6     x
+        1   0    3     7     0
+        2   1    4     8     8
+        3   2    5     9     9
+        4  hi  bye  wait    hi
+
+    Similar to SQL and dplyr's case_when
     with inspiration from `pydatatable` if_else function.
 
     If your scenario requires direct replacement of values,
```
thatlittleboy, line 49: Hmm, Github doesn't allow reviews to be marked on unchanged code as of now 😅

Anyway, just wanted to point out:
- there's a typo of "callabe" at approximately L104 (params docstring for *args).
- and it bugs me a little that in the docstring of `case_when`, `df` is described as "Pandas dataframe" while the `returns` is described as "pandas DataFrame", shall we standardize the casing across the documentation? I've been using the latter in my previous PR.

thatlittleboy (approved):
I've just pointed a couple formatting issues to fix. The overall logic and "minimalism" of the examples LGTM :)

janitor/functions/case_when.py
ericmjl: Good catch! I will push a fix shortly.

janitor/functions/case_when.py
ericmjl: Good catch on the typo! Thank you for that.

Also, I agree with you about standardizing the casing. We'll go with the latter.

Issue #826, opened 2021-04-07
Minimal Pyjanitor instalation


GuiMarthe:
Hey folks, I recently looked at the package dependency for pyjanitor and it seems too large for a production environment.
I saw in the `.requirements` directory that there are a few sets of dependencies for different use cases, but I don't see anywhere how to actually limit the scope at installation time. 

Is the following possible?
`pip install pyjanitor[base]`

Just in case, this is the packages dependency list generated by `pipdeptree`.

```
pyjanitor==0.20.10
  black==20.8b1
    appdirs==1.4.4
    click==7.1.2
    mypy-extensions==0.4.3
    pathspec==0.8.1
    regex==2021.4.4
    toml==0.10.2
    typed-ast==1.4.2
    typing-extensions==3.7.4.3
  hypothesis==6.8.9
    attrs==20.3.0
    sortedcontainers==2.3.0
  interrogate==1.3.2
    attrs==20.3.0
    click==7.1.2
    colorama==0.4.4
    py==1.10.0
    tabulate==0.8.9
    toml==0.10.2
  ipykernel==5.5.3
    ipython==7.21.0
      backcall==0.2.0
      decorator==5.0.5
      jedi==0.18.0
        parso==0.8.2
      pexpect==4.8.0
        ptyprocess==0.7.0
      pickleshare==0.7.5
      prompt-toolkit==3.0.18
        wcwidth==0.2.5
      Pygments==2.8.1
      setuptools==54.2.0
      traitlets==5.0.5
        ipython-genutils==0.2.0
    jupyter-client==6.1.13
      jupyter-core==4.7.1
        traitlets==5.0.5
          ipython-genutils==0.2.0
      nest-asyncio==1.5.1
      python-dateutil==2.8.0
        six==1.15.0
      pyzmq==22.0.3
      tornado==6.1
      traitlets==5.0.5
        ipython-genutils==0.2.0
    tornado==6.1
    traitlets==5.0.5
      ipython-genutils==0.2.0
  isort==5.8.0
  jupyter-client==6.1.13
    jupyter-core==4.7.1
      traitlets==5.0.5
        ipython-genutils==0.2.0
    nest-asyncio==1.5.1
    python-dateutil==2.8.0
      six==1.15.0
    pyzmq==22.0.3
    tornado==6.1
    traitlets==5.0.5
      ipython-genutils==0.2.0
  lxml==4.6.3
  natsort==7.1.1
  nbsphinx==0.8.2
[...]

PR #990, opened 2022-01-16
[ENH] Add tidyxl's `xlsx_table` to the `io` module.

### LINKED ISSUE #929: Implement tidyxl for Excel files
# Brief Description

I would like to propose an implementation of [tidyxl](https://cran.r-project.org/web/packages/tidyxl/vignettes/tidyxl.html) within Pyjanitor for non tabular data in Excel files.

The idea is to read in the Excel file, if it cannot be handled easily with Pandas because of the non tabular nature of the data; [tidyxl](https://cran.r-project.org/web/packages/tidyxl/vignettes/tidyxl.html) in R represents each cell in its own row, where it describes the cell’s address, value and other properties. The data can then be wrangled with Pandas/Pyjanitor in a vectorized fashion.

There is a [cookbook](https://nacnudus.github.io/spreadsheet-munging-strategies/case-studies.html) that shows the advantages of this approach. I feel it might be a way to ease wrangling Excel data in python, especially for non tabular data.

### LINKED ISSUE #992: Read table(s) from Excel File
[# Brief Description

I would like to propose an API to read [tables](https://support.microsoft.com/en-us/office/overview-of-excel-tables-7ab0bb7d-3a9e-4b56-a3c9-6c94334e492c#:~:text=To%20quickly%20create%20a%20table,row%2C%20and%20then%20click%20OK.) from an Excel (`.xlsx`) file.

Currently, reading this in Pandas, requires knowing the region where the table is, and in some cases, you might have to drop nulls, or do some other cleanup. With this proposed API, it should be easy to pull the table, since it is already in a clean format

# Example API

```python
# transform only one column, while creating a new column name for it.
jn.xlsx_table(filename, sheetname, table_name)
```

samukweku:
# PR Description

- Create DataFrame or dictionary of DataFrames from table(s) in xlsx sheet.

**This PR resolves #992  .**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


janitor/io.py
```diff
@@ -110,3 +111,66 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    If the `table` argument is provided, a pandas DataFrame is returned;
+    if the `table` argument is None, or a list/tuple of names,
+    a dictionary of DataFrames is returned, where the keys of the dictionary
+    are the table names.
+
+    :param path: path to the Excel File.
+    :param sheetname: Name of the sheet
+        from which the tables are to be extracted.
+    :param table: name of a table, or list of tables in the sheet.
+    :raises ValueError: if there are not tables in the sheet.
```
ericmjl: I think the `:raises ValueError:` section is conventionally placed at the end of the docstring. Could you check our convention used in the library and conform this docstring to the rest?

janitor/io.py
```diff
@@ -110,3 +111,66 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    table: Union[str, list, tuple] = None,
+    header: bool = True,
+) -> pd.DataFrame:
+    """
+    Returns a DataFrame of values in a table in the Excel file.
+    If the `table` argument is provided, a pandas DataFrame is returned;
+    if the `table` argument is None, or a list/tuple of names,
+    a dictionary of DataFrames is returned, where the keys of the dictionary
+    are the table names.
+
```
ericmjl, line 132: Could you add in an example usage (using the triple back-ticks (```) , rather than the triple right-chevron `>>>`? 

We're avoiding the latter here because the I/O module can't be easily doc-tested, I think.

ericmjl (changes requested):
Great start, @samukweku! I put in some early review comments here. There are tests that are missing too, but I trust you're already working on them. Looking forward to reviewing!

ericmjl:
I'm inclined to ignore the decrease in project test coverage; they're irrelevant to the PR. 100% of the patch diff being hit is good enough.

ericmjl:
Btw, @samukweku, I think going forth you can use a branch off the `pyjanitor` repo, rather than a personal fork. Or otherwise, you can delete the old forks that you have and re-fork. (This is what I do with projects I'm not a maintainer on.) It helps eliminate the need to maintain a fork and keep the `dev` branch up-to-date.

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+                      1           2       Advanced
+                      2           3      Freestyle
+                      3           4    Competition
+                      4           5  Long Distance,
+        'dSupplier':   SupplierID             Supplier        City State                         E-mail
+                     0         GB       Gel Boomerangs     Oakland    CA          gel@gel-boomerang.com
+                     1         CO  Colorado Boomerangs    Gunnison    CO  Pollock@coloradoboomerang.com
+                     2         CC        Channel Craft    Richland    WA                    Dino@CC.com
+                     3         DB        Darnell Booms  Burlington    VT            Darnell@Darnell.com}
+    ```
```
ericmjl: I think the extra intent inside the triple back-ticks is unnecessary.

midpoint.csv
```diff
@@ -0,0 +1,3 @@
+,a_new,b
```
ericmjl: Curious to hear, what is the purpose of this file? If it is not necessary, how did it arise, and can we delete it?

tests/io/test_xlsx_table.py
```diff
@@ -0,0 +1,109 @@
+import os
+import pandas as pd
+import pytest
+
+from janitor import io
+from pandas.testing import assert_frame_equal
+
+
+TEST_DATA_DIR = "tests/test_data"
```
ericmjl, line 9: Can we make use of Pathlib to provide a _resolved_ path?

E.g. `(here() / tests/test_data).resolve()`?

Doing so will make explicit to someone reading the code how the `test_data` dir is resolved, at least relative to the root dir (by use of `here()`), while the `.resolve()` will help with debugging the path (in case the need ever arises).

ericmjl (changes requested):
I love the tests provided here, @samukweku, they're really comprehensive!

I have only a few nit-picky requests for changes. They are explicit requests for changes, rather than mere suggestions, because I think changing them will help with maintenance clarity. Hope you don't mind them!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 -   [ENH] Fix `concatenate_columns` not working on category inputs @zbarry
 -   [INF] Simplify CI system @ericmjl
 -   [ENH] Added "read_commandline" function to janitor.io @BaritoneBeard
+-   [ENH] Add 'xlsx_table`, for reading tables from an Excel sheet. @samukweku
```
thatlittleboy: Typo here with the quote/backticks

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    are the table names.
+
+    Example:
+
+    ```python
+
+        filename = "excel_table.xlsx"
+
+        # single table
+        jn.xlsx_table(filename, table = 'dCategory')
```
thatlittleboy: A little confused by the given examples, is the `sheetname` argument missing?

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+        'dSupplier':   SupplierID             Supplier        City State                         E-mail
+                     0         GB       Gel Boomerangs     Oakland    CA          gel@gel-boomerang.com
+                     1         CO  Colorado Boomerangs    Gunnison    CO  Pollock@coloradoboomerang.com
+                     2         CC        Channel Craft    Richland    WA                    Dino@CC.com
+                     3         DB        Darnell Booms  Burlington    VT            Darnell@Darnell.com}
+    ```
+
+    :param path: path to the Excel File.
+    :param sheetname: Name or position of the sheet
+        from which the tables are to be extracted.
```
thatlittleboy: Question: (haven't used openpyxl before) But from what I can gather, the Workbook object only allows indexing (`__getitem__`) by sheetname, not position? https://openpyxl.readthedocs.io/en/stable/_modules/openpyxl/workbook/workbook.html#Workbook

(When I read the word "position", I'm assuming I can do `sheetname=0`, which I suppose is not allowed.)

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+                      3           4    Competition
+                      4           5  Long Distance,
+        'dSupplier':   SupplierID             Supplier        City State                         E-mail
+                     0         GB       Gel Boomerangs     Oakland    CA          gel@gel-boomerang.com
+                     1         CO  Colorado Boomerangs    Gunnison    CO  Pollock@coloradoboomerang.com
+                     2         CC        Channel Craft    Richland    WA                    Dino@CC.com
+                     3         DB        Darnell Booms  Burlington    VT            Darnell@Darnell.com}
+    ```
+
+    :param path: path to the Excel File.
```
thatlittleboy: Let's standardize capitalization in these parameter descriptions? Similarly for the description of the `table` parameter.

```suggestion
    :param path: Path to the Excel File.
```

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    :param sheetname: Name or position of the sheet
+        from which the tables are to be extracted.
+    :param table: name of a table, or list of tables in the sheet.
```
thatlittleboy: Might be good to reiterate what `table=None` means here. Even more so since it's the default argument. :)
If I'm reading the code right, it means to return all of the tables in the worksheet?

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    :returns: A pandas DataFrame, or a dictionary of DataFrames,
+        if there are multiple arguments for the `table` parameter.
```
thatlittleboy: Not strictly correct, I suppose. Since `None` also returns a dictionary (and to confirm: even if there is only 1 table in the sheet, right?)

EDIT: ah.. doesn't seem like it, I see a `if len(frames) == 1: frames.popitem()` at the end.. Perhaps the docstrings can be clarified a little to reflect this better?

I presume the intended effect is to: return a pandas DataFrame if there is only 1 table in the sheet, or if table is specified as a string. Otherwise, return a dictionary, where the keys are the Excel table names, while values are the Excel tables read in as pandas DataFrames.

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    :raises ValueError: if there are no tables in the sheet.
+
+    """  # noqa : E501
+
+    wb = load_workbook(filename=path, read_only=False, keep_links=False)
```
thatlittleboy, line 185: Perhaps I'm missing something, but would just like to understand a little why a decision was made to set `read_only` to False here?

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+        table = [table]
+    if table is not None:
+        check("table", table, [list, tuple])
+        for entry in table:
+            if entry not in contents.keys():
+                raise ValueError(
+                    f"""
+                    {entry} is not a table
+                    in the {sheetname} sheet.
+                    """
```
thatlittleboy: The newlines in this triplequoted string will actually manifest in the error message itself, which is probably not desirable. Something like this might be better if your error message is too long for your liking, just FYI:

```python
    raise ValueError(
        f"{entry} is not a table "
        f"in the {sheetname} sheet."
    )
```
Note the space after the word "table " so you don't get "...not a tablein the...".

But I just tried this ^^ in the codebase, seems like this error message of yours is short enough that `black` autoformats it to be in 1 line. So let's just write it in 1 line here?

janitor/io.py
```diff
@@ -110,3 +111,103 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
+    contents = ws.tables
+    if not contents:
+        raise ValueError(f"There is no table in `{sheetname}` sheet.")
+
+    if isinstance(table, str):
+        table = [table]
+    if table is not None:
+        check("table", table, [list, tuple])
+        for entry in table:
+            if entry not in contents.keys():
```
thatlittleboy: Could you try removing `.keys()` here and see if it still runs well?

If `contents` is a dictionary (and the openpyxl docs say it is a "dictionary-like" object), then the "is in" check on dictionaries is automatically checked against the keys.
The `.keys()` call is redundant.

thatlittleboy (changes requested):

CHANGELOG.md
ericmjl: Wow, that's next-level detail, @thatlittleboy! Great catch!

janitor/io.py
```diff
@@ -2,6 +2,7 @@
 import subprocess
 from glob import glob
 from io import StringIO
+from openpyxl import load_workbook
```
thatlittleboy: Two questions regarding requirements:
1. Can we make this import optional?
2. I suppose we need to add this to our package requirements? And I suppose this package needs to be optional as well. To prevent pyjanitor install from being too bloated.

thatlittleboy (changes requested):
Sorry, immediately thought of another point I would like to make in the review right as I pressed "Submit" 😅

janitor/io.py
thatlittleboy: 1 more thing: the return type annotation of the function was written above as `-> pd.DataFrame`, after reviewing the full MR, I suppose it should be something like `-> Union[pd.DataFrame, dict]`?

midpoint.csv
samukweku: hmmm, no idea what i was doing to get this :grin:, will delete

tests/io/test_xlsx_table.py
samukweku: what is `here()`? seems like an R thing. Do we have something similar in pyjanitor?

janitor/io.py
samukweku: accessing tables is not possible when `read_only` is True ( True is for memory efficiency, but limited in scope)

janitor/io.py
samukweku: which brings us to #826 ; any ideas on how to go about this?

samukweku: I'll probably have a look at this and see where the rabbit hole leads

janitor/io.py
thatlittleboy: I see, gotcha 👌🏼

tests/io/test_xlsx_table.py
ericmjl: Oh my, my bad - `here()` comes from `pyprojroot`, a project by @chendaniely. It's become so ingrained in my headspace that I forgot that it's actually an external dependency. I think we can ignore the request to use `here()`, though I think the `.resolve()` will help debug in the future.

janitor/io.py
ericmjl: Good link there, @samukweku! I haven't had the time to think about the installation experience, but it looks like improving that would be incredibly helpful. We can ignore the issue for now, though. I would suggest that to go simple, we add `openpyxl` as an optional dependency and use the same pattern in the `chemistry` and `biology` submodules to conditionally import or raise warning/error.

thatlittleboy: For 1., I was thinking of shifting this openpyxl import into the `xlsx_table` function itself. Then do a try import + raising import error message similar to how `biology.py` does it.

For 2. I'm not sure if `openpyxl` fits into any existing EXTRA_REQUIRES categories ([cf. here](https://github.com/pyjanitor-devs/pyjanitor/blob/dev/setup.py#L59)), maybe a new category like `io` or `excel` would be necessary. I'm leaning more towards `io`to leave the option open in the future for more io functions that require other packages.

ericmjl (approved):
I'm approving. I looked on CodeCov and saw that the missed patch is on the conditional import. I don't think it's easy to test that in the current testing environment, so we should let it go.

I'm going to approve! @thatlittleboy since you also did a big review on this one, feel free to approve and merge!

thatlittleboy:
@samukweku There's a merge conflict in `CHANGELOG.md` because of the previous PR going through. Could you address this, then I'll approve and merge? Thanks!

thatlittleboy (approved):

PR #996, opened 2022-01-22
[DOC] Add MWE for drop_* functions

ericmjl:
As per title.

Tagging @samukweku and @thatlittleboy for review.

### DIFF
--- a/janitor/functions/drop_constant_columns.py
+++ b/janitor/functions/drop_constant_columns.py
@@ -1,45 +1,35 @@
+"""Implementation of drop_constant_columns."""
 import pandas_flavor as pf
 import pandas as pd
 
 
 @pf.register_dataframe_method
-def drop_constant_columns(
-    df: pd.DataFrame,
-) -> pd.DataFrame:
+def drop_constant_columns(df: pd.DataFrame) -> pd.DataFrame:
     """
     Finds and drops the constant columns from a Pandas DataFrame.
 
-    This method does not mutate the original DataFrame.
+    Example:
 
-    Functional usage syntax:
-
-    ```python
-    import pandas as pd
-    import janitor as jn
-
-    data_dict = {
-    "a": [1, 1, 1] * 3,
-    "Bell__Chart": [1, 2, 3] * 3,
-    "decorated-elephant": [1, 1, 1] * 3,
-    "animals": ["rabbit", "leopard", "lion"] * 3,
-    "cities": ["Cambridge", "Shanghai", "Basel"] * 3
-    }
-
-    df = pd.DataFrame(data_dict)
-
-    df = jn.functions.drop_constant_columns(df)
-    ```
-
-    Method chaining usage example:
-
-    ```python
-    import pandas as pd
-    import janitor
-
-    df = pd.DataFrame(...)
-
-    df = df.drop_constant_columns()
-    ```
+        >>> import pandas as pd
+        >>> import janitor
+        >>> data_dict = {
+        ...     "a": [1, 1, 1] ,
+        ...     "Bell__Chart": [1, 2, 3] ,
+        ...     "decorated-elephant": [1, 1, 1] ,
+        ...     "animals": ["rabbit", "leopard", "lion"] ,
+        ...     "cities": ["Cambridge", "Shanghai", "Basel"]
+        ... }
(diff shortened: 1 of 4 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-996--pyjanitor.netlify.app

janitor/functions/drop_constant_columns.py
```diff
@@ -1,45 +1,35 @@
...
+
+        >>> import pandas as pd
+        >>> import janitor
+        >>> data_dict = {
+        ...     "a": [1, 1, 1] ,
+        ...     "Bell__Chart": [1, 2, 3] ,
+        ...     "decorated-elephant": [1, 1, 1] ,
+        ...     "animals": ["rabbit", "leopard", "lion"] ,
+        ...     "cities": ["Cambridge", "Shanghai", "Basel"]
+        ... }
```
thatlittleboy: 1. There are unnecessary spaces before each comma in these few lines.
2. And would like to suggest to simplify the column names a little (not so many symbols, preferably just a,b,c or col1,2,3 etc.), since the point of the MWE here is not the column names themselves.

janitor/functions/drop_duplicate_columns.py
```diff
@@ -9,40 +10,43 @@ def drop_duplicate_columns(
 ) -> pd.DataFrame:
     """Remove a duplicated column specified by column_name, its index.
 
-    This method does not mutate the original DataFrame.
-
-    Column order 0 is to remove the first column,
-           order 1 is to remove the second column, and etc
+    Specifying `nth_index=0` will to remove the first column,
```
thatlittleboy:
```suggestion
    Specifying `nth_index=0` will remove the first column,
```

slight typo here

janitor/functions/drop_duplicate_columns.py
```diff
@@ -51,7 +55,7 @@ def drop_duplicate_columns(
     removed_col_idx = col_indexes[nth_index]
     # get the column indexes without column that is being removed
     filtered_cols = [
-        c_i for c_i, c_v in enumerate(cols) if c_i != removed_col_idx
+        c_i for c_i, c_v in enumerate(df.columns) if c_i != removed_col_idx
```
thatlittleboy: This implementation looks a little peculiar to me. Though nothing wrong logically, but since `c_v` is not used, it appears to me that enumerate is not necessary here.

Something like
```python
filtered_cols = [idx for idx in range(df.shape[1]) if idx != removed_col_idx]
```
seems like it would work as well?

thatlittleboy (approved):
Just a couple of things to clean up, otherwise LGTM :)

janitor/functions/drop_constant_columns.py
ericmjl: That's a great nitpick, @thatlittleboy! I like the idea, and will change the column names.

janitor/functions/drop_duplicate_columns.py
ericmjl: Good point. I think the implementation could be simplified further too, by using `df.drop` instead. Let me try that out.

ericmjl: Never mind, I was wrong :smile:. Using `df.drop` makes things a bit more difficult!

ericmjl: @thatlittleboy on second thought, I think iterating over `df.columns` is a bit more explicit than doing `df.shape[1]`. I've replaced the loop variable `c_v` with an underscore `_` instead, which together with `enumerate(df.columns)` should help make clear that we're only interested in the index of each column name.

thatlittleboy: Yep, understand. That works too!

ericmjl:
Oh bummer, once again Codecov is being finnicky. Haha. I looked at the diff and since 100% of diff is hit, I think we can safely merge. Thanks @thatlittleboy for the review comments! I'll wait for @samukweku to review first, or if you're in a time crunch, sam, then just let us know. Once that's done, @thatlittleboy, please do the honors of hitting Squash and Merge!

samukweku (approved):

thatlittleboy:
Merged. Mentioning #972 for posterity.

PR #997, opened 2022-01-25
[DOC] Add MWE for 4 functions

thatlittleboy:
# PR Description

Add MWE to and clean up the following functions

- `functions/move.py`
- `functions/remove_columns.py`
- `functions/remove_empty.py`
- `functions/limit_column_characters.py`

Also note that for `move.py`, I moved the `df.copy()` call lower down the function until after the argument validation checks, as it is unnecessary if the checks fail.

**This PR is related to #972.**

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions/move.py
+++ b/janitor/functions/move.py
@@ -1,3 +1,4 @@
+"""Implementation of move."""
 import pandas_flavor as pf
 import pandas as pd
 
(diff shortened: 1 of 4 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-997--pyjanitor.netlify.app

janitor/functions/move.py
```diff
@@ -13,46 +14,63 @@ def move(
...
+        >>> import pandas as pd
+        >>> import janitor
+        >>> df = pd.DataFrame({"a": [2, 4, 6, 8], "b": list("wxyz")})
+        >>> df
+           a  b
+        0  2  w
+        1  4  x
+        2  6  y
+        3  8  z
+        >>> df.move(source=0, target=3, position="before", axis=0)
```
ericmjl, line 37: @thatlittleboy I think we might be able to make the API a bit nicer. What do you think of the following?

```python
# args order is [axis, source, position, target, cmd], all default to None.
# `cmd` is a placeholder; we don't have to use this name, can be called something better instead.
df.move("row", 0, "after", 3)
df.move("column", "b", "after", "c")

# here we pass in only cmd, which gets split into [axis, source, position, target]
# and used in the underlying implementation.
df.move(cmd="row 0 after 3")
df.move(cmd="column b after c")
```

This would make the `move` function a bit more declarative and English-like. What do you think?

ericmjl: This change doesn't have to be made in this PR, and we can raise a new issue for it instead.

samukweku (approved):

ericmjl (approved):

janitor/functions/move.py
thatlittleboy: @ericmjl I like it (on initial instincts) 👍🏼 ! It reads really nicely, particularly with the function name, "move row 0 after 3". Agree, let's raise an issue and deliberate it separately from this PR.

ericmjl:
Going to merge!

PR #1001, opened 2022-01-27
[ENH] Refactor/Clean conditional_join

samukweku:

- A bit of refactoring, with a view to providing a numba option in the future ( for performance gains)
- Bit more comments so i can understand the code more (and hopefully others)
- A bit more strict - equality only joins are not supported. (pd.merge does this excellently already)

**This PR improves conditional_join.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1001--pyjanitor.netlify.app

janitor/functions/conditional_join.py
```diff
@@ -1,21 +1,20 @@
...
+    is_extension_array_dtype,
 )
+from pandas.core.reshape.merge import _MergeOperation
 import pandas_flavor as pf
 import pandas as pd
 from typing import Union
 import operator
 from janitor.utils import check, check_column
 import numpy as np
 from enum import Enum
```
thatlittleboy: Shall we `isort` these imports? It's a bit disorderly at the moment.

janitor/functions/conditional_join.py
```diff
@@ -65,38 +64,58 @@ def conditional_join(
...
+        8         3         1         3
+        9         3         2         4
+        10        3         3         5
+        11        3         3         6
+        12        4         2         4
+        13        4         3         5
+        14        4         3         6
 
 
     :param df: A Pandas DataFrame.
```
thatlittleboy: Some comments pertaining to the docstring of the params:

1. `df`
```suggestion
    :param df: A pandas DataFrame.
```

2. `conditions`: Might it be worth mentioning that "if multiple conditions are passed, they will be combined together using a `logical_and`"? (if you have some better way of phrasing this, pls do!)

janitor/functions/conditional_join.py
```diff
@@ -154,75 +173,32 @@ class _JoinTypes(Enum):
...
+    Used in `conditional_join`.
+    """
+    sequence_of_operators = {op.value for op in _JoinOperator}
+    if op not in sequence_of_operators:
+        raise ValueError(
+            f"""
+             The conditional join operator
+             should be one of {", ".join(sequence_of_operators)}
+             """
         )
```
thatlittleboy: Similar comment to the the previous PR: triple-quoted strings in Python retain the newlines in the output, which is not what we want in the error message. To break up the message string into visual lines in the source code but not in the printed output, we can make use of implicit string concatenation.

Also, I'm thinking perhaps there's no need to manually join them up, the repr for python sets should already format the output nicely.

Something like:

```python
raise ValueError(
    "The conditional join operator should be one "
    "of {}".format(sequence_of_operators)
)
```

janitor/functions/conditional_join.py
```diff
@@ -333,21 +313,22 @@ def _conditional_join_type_check(
         raise ValueError(
             """
             conditional_join only supports
-            string, category, integer,
-            float or date dtypes.
+            string, category, numeric, or date dtypes.
             """
         )
```
thatlittleboy: Same comment here about multiline error messages. There's a couple of these scattered across the file, I'm unable to point them out due to them not being part of the PR diff, please address them nevertheless, thank you!

janitor/functions/conditional_join.py
thatlittleboy: For the second point, something similar was done for the `join_types` as well in the error msg, I suggest simplifying that too.

janitor/functions/conditional_join.py
```diff
@@ -65,38 +64,58 @@ def conditional_join(
...
+        5         1         0         1
+        6         1         0         1
+        7         1         1         3
+        8         3         1         3
+        9         3         2         4
+        10        3         3         5
+        11        3         3         6
+        12        4         2         4
+        13        4         3         5
+        14        4         3         6
```
ericmjl: Just a suggestion to remove the unnecessary whitespace.

```suggestion
        >>> import pandas as pd
        >>> import janitor
        >>> df1 = pd.DataFrame({"value_1": [2, 5, 7, 1, 3, 4]})
        >>> df2 = pd.DataFrame({"value_2A": [0, 3, 7, 12, 0, 2, 3, 1],
        ...                     "value_2B": [1, 5, 9, 15, 1, 4, 6, 3],
        ...                    })
        >>> df1
           value_1
        0        2
        1        5
        2        7
        3        1
        4        3
        5        4
        >>> df2
           value_2A  value_2B
        0         0         1
        1         3         5
        2         7         9
        3        12        15
        4         0         1
        5         2         4
        6         3         6
        7         1         3
        >>> df1.conditional_join(
        ...     df2,
        ...     ("value_1", "value_2A", ">="),
        ...     ("value_1", "value_2B", "<=")
        ... )
            value_1  value_2A  value_2B
        0         2         1         3
        1         2         2         4
        2         5         3         5
        3         5         3         6
        4         7         7         9
        5         1         0         1
        6         1         0         1
        7         1         1         3
        8         3         1         3
        9         3         2         4
        10        3         3         5
        11        3         3         6
        12        4         2         4
        13        4         3         5
        14        4         3         6
```

janitor/functions/conditional_join.py
```diff
@@ -300,6 +276,11 @@ def _conditional_join_preliminary_checks(
         check_column(right, right_on)
         _check_operator(op)
 
+    if all(
+        (op == _JoinOperator.STRICTLY_EQUAL.value for *_, op in conditions)
+    ):
+        raise ValueError("""Equality only joins are not supported.""")
```
ericmjl: Looks like codecov is complaining that this line wasn't covered by the tests.

Can we add a test for this case, please?

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +695,421 @@ def _not_equal_indices(
...
+    left_c = extract_array(df[left_on], extract_numpy=True)[left_index]
+    right_c = extract_array(right[right_on], extract_numpy=True)[right_index]
+
+    mask = op(left_c, right_c)
+
+    if ext_arr:
+        mask = mask.to_numpy(dtype=bool, na_value=False)
+
+    if not mask.any():
+        return None
```
ericmjl: Looks like this one is also not covered by the tests.

ericmjl (changes requested):
Just a few requests to add coverage to the tests!

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +695,421 @@ def _not_equal_indices(
...
+    or _not_equal_indices).
+    """
+    strict = False
+
+    if op in {
+        _JoinOperator.GREATER_THAN.value,
+        _JoinOperator.LESS_THAN.value,
+        _JoinOperator.NOT_EQUAL.value,
+    }:
+        strict = True
```
thatlittleboy, line 693: Could we do this assignment for `strict` as an if-else instead? Intention is clearer that way IMO.

janitor/functions/conditional_join.py
```diff
@@ -375,554 +356,141 @@ def _conditional_join_type_check(
...
 
-    if mask.all():
-        return df_index, right_index
+        result = _generic_func_cond_join(
+            df[left_on], right[right_on], op, multiple_conditions
+        )
 
-    if is_extension_array_dtype(mask):
-        mask = mask.to_numpy(dtype=bool, na_value=False)
+        # return result
```
thatlittleboy: Can we remove this `# return result` line? I'm guessing this line was leftover from debugging.

And also one more couple lines below from here.

tests/functions/test_conditional_join.py
```diff
@@ -1457,69 +1457,98 @@ def test_multiple_non_equii(df, right):
...
+    columns = ["B", "A", "E", "Floats", "Integers", "Dates"]
+    expected = (
+        df.assign(t=1)
+        .merge(
+            right.assign(t=1),
+            on="t",
+            how="inner",
+            sort=False,
+        )
+        .query(f"{l_ge} <= {r_ge} and {l_le} < {r_le} and {l_eq} < {r_eq}")
```
thatlittleboy, line 1521: Hmm.. does this test run correctly?

I think the individual conditions need to be surrounded by parenthesis, i.e., 

```python
.query(f"({l_ge} <= {r_ge}) and ({l_le} < {r_le}) and ({l_eq} < {r_eq})")
```

I see a couple more "query-multiple-expressions" scattered around, consider fixing those if the parentheses are necessary.

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +695,421 @@ def _not_equal_indices(
...
+    le_lt = None
+    ge_gt = None
+    for condition in conditions:
+        *_, op = condition
+        if op in less_than_join_types:
+            le_lt = condition
+        elif op in greater_than_join_types:
+            ge_gt = condition
+        if le_lt and ge_gt:
+            break
```
thatlittleboy, line 880: Could you help me understand why we're breaking here? Suppose if conditions is written like:
```python
('a', 'b', '<'),
('b', 'c', '>'),
('c', 'd', '<'),
```
Won't the third condition be included in `rest`?
Or would that be expected behaviour?

(Actually I think this is stemming from me not quite understanding the logic for `rest` and its place in `_generate_indices`. Would be grateful if you could summarize the logic here haha)

thatlittleboy (changes requested):
It's a pretty big PR, I'll admit I'm abit lost in the logic implemented for conditional join  :x 
I've made some general comments, but can't say that I've checked through the accuracy of the refactoring in full.

Apologies! (Will try to understand the logic in the meantime nevertheless)

janitor/functions/conditional_join.py
ericmjl: I think the pattern that @samukweku has here is one that I showed him before. I think my rationale here was that for short code blocks, the pattern shown here communicates "here's my default, change only under these conditions", which is a bit like having default values for `kwargs`. Hope you don't mind, @thatlittleboy, but in this case, I think leaving as-is might be better for consistency across the codebase.

tests/functions/test_conditional_join.py
samukweku: the query function allows this form, a similar form is in the pandas docs, as an example

janitor/functions/conditional_join.py
samukweku: added an explanation to the code; hopefully it clarifies things

janitor/functions/conditional_join.py
ericmjl: @samukweku just curious, I think I agree with @thatlittleboy about the multi-line string. If you need to conform to line limits, you could try:

```python
raise ValueError(
    "conditional_join only supports "
    "string, category, numeric, or date dtypes."
)
```

janitor/functions/conditional_join.py
```diff
@@ -332,21 +315,15 @@ def _conditional_join_type_check(
...
     if (
         is_categorical_dtype(left_column)
         and op != _JoinOperator.STRICTLY_EQUAL.value
     ):
         raise ValueError(
-            """
-            For categorical columns,
-            only the `==` operator is supported.
-            """
+            "For categorical columns, " "only the `==` operator is supported."
```
ericmjl:
```suggestion
            "For categorical columns, only the `==` operator is supported."
```

janitor/functions/conditional_join.py
```diff
@@ -332,21 +315,15 @@ def _conditional_join_type_check(
         and op != _JoinOperator.STRICTLY_EQUAL.value
     ):
         raise ValueError(
-            """
-            For string columns,
-            only the `==` operator is supported.
-            """
+            "For string columns, " "only the `==` operator is supported."
```
ericmjl:
```suggestion
            "For string columns, only the `==` operator is supported."
```

janitor/functions/conditional_join.py
```diff
@@ -291,10 +278,8 @@ def _conditional_join_preliminary_checks(
 def _conditional_join_type_check(
     left_column: pd.Series, right_column: pd.Series, op: str
 ) -> None:
-    """
-    Raise error if column type is not
-    any of numeric or datetime or string.
-    """
+    "Raise error if column type is not"
```
ericmjl:
```suggestion
    "Raise error if column type is not "
```

ericmjl (approved):
Final suggested changes, @samukweku! Because they're so small, I'm pre-approving the PR. @thatlittleboy please do the honors!

janitor/functions/conditional_join.py
```diff
@@ -245,10 +219,7 @@ def _conditional_join_preliminary_checks(
 
     if isinstance(df.columns, pd.MultiIndex):
         raise ValueError(
-            """
-            MultiIndex columns are not
-            supported for conditional_join.
-            """
+            "MultiIndex columns are not " "supported for conditional_join."
```
thatlittleboy:
```suggestion
            "MultiIndex columns are not supported for conditional_join."
```

janitor/functions/conditional_join.py
```diff
@@ -259,37 +230,25 @@ def _conditional_join_preliminary_checks(
     if isinstance(right, pd.Series):
         if not right.name:
             raise ValueError(
-                """
-                Unnamed Series are not supported
-                for conditional_join.
-                """
+                "Unnamed Series are not supported " "for conditional_join."
```
thatlittleboy:
```suggestion
                "Unnamed Series are not supported for conditional_join."
```

janitor/functions/conditional_join.py
```diff
@@ -259,37 +230,25 @@ def _conditional_join_preliminary_checks(
...
             )
         right = right.to_frame()
 
     if isinstance(right.columns, pd.MultiIndex):
         raise ValueError(
-            """
-            MultiIndex columns are not supported
-            for conditional joins.
-            """
+            "MultiIndex columns are not supported " "for conditional joins."
```
thatlittleboy:
```suggestion
            "MultiIndex columns are not supported for conditional joins."
```

janitor/functions/conditional_join.py
```diff
@@ -314,15 +278,12 @@ def _conditional_join_preliminary_checks(
 def _conditional_join_type_check(
     left_column: pd.Series, right_column: pd.Series, op: str
 ) -> None:
-    """
-    Raise error if column type is not
-    any of numeric or datetime or string.
-    """
+    "Raise error if column type is not"
+    "any of numeric or datetime or string."
```
thatlittleboy:
```suggestion
    """
    Raise error if column type is not any of numeric or datetime or string.
    """
```
Let's use multiline string here instead

janitor/functions/conditional_join.py
```diff
@@ -985,10 +511,9 @@ def _greater_than_indices(
     where `left_c` is greater than
     (but not equal to) `right_c` are returned.
 
-    Returns a tuple of (left_c, right_c).
-    """
-    left_c = df[left_on]
-    right_c = right[right_on]
+    if len_conditions, a tuple of integer indexes for left_c and right_c
+    is returned; else a tuple of the index for left_c, right_c, as well
+    as the positions of left_c in right_c is returned."""
```
thatlittleboy: I'm seeing from the code below:

```python
    if multiple_conditions:
        return left_index, right_index, search_indices
    # <...omitted...>
    return left_c, right_c
```
So can I clarify you meant to say if there is only 1 condition, return int indexes of left_c and right_c. If there are multiple conditions then return the indexes + positions? Since there is no `len_conditions` anywhere.

janitor/functions/conditional_join.py
```diff
@@ -1000,6 +525,9 @@ def _greater_than_indices(
     any_nulls = pd.isna(left_c.array)
     if any_nulls.any():
         left_c = left_c[~any_nulls]
+    if left_c.empty | right_c.empty:
+        return None
+    any_nulls = None
```
thatlittleboy, line 532: A touch curious why this `any_nulls = None` line is here. To save on memory?

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +661,493 @@ def _not_equal_indices(
...
+def _generate_indices(
+    left_index: np.ndarray, right_index: np.ndarray, conditions: list
+):
+    """
+    Run a for loop to get the final indices.
+    This iteratively goes through each condition,
+    builds a boolean array,
+    and gets indices for rows that meet the condition requirements.
+    `conditions` is a list of tuples, where a tuple is of the form:
+    `(Series from df, Series from right, operator).
```
thatlittleboy:
```suggestion
    `(Series from df, Series from right, operator)`.
```

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +661,493 @@ def _not_equal_indices(
...
+
+    Returns a tuple of (df_index, right_index)
+    """
+
+    # there is an opportunity for optimization for range joins
+    # which is usually `lower_value < value < upper_value`
+    # or `lower_value < a` and `b < upper_value`
+    # intervalindex is not used here, as there are scenarios
+    # where there will be overlapping intervals;
+    # intervalindex does not offer an efficeint way to get
```
thatlittleboy:
```suggestion
    # intervalindex does not offer an efficient way to get
```

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +661,493 @@ def _not_equal_indices(
...
+    # range joins are also the more common types of non-equi joins
+    # the other joins do not have an optimisation opportunity
+    # as far as I know, so a blowup of all the rows
+    # is unavoidable.
+    # future PR would use numba to improve performance
+    # still doesnt help that an optimisation path is not available
+    # that I am aware of
+
+    # first step is to get two conditions, if possible
+    # wehre one has a less than operator
```
thatlittleboy:
```suggestion
    # where one has a less than operator
```

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +661,493 @@ def _not_equal_indices(
...
+            # get the index positions where left_c is </<= right_c
+            # that minimum position combined with the equivalent position
+            # from search_indices becomes our search space
+            # for the equivalent left_c index
+            if keep_rows.any():
+                pos[counter[keep_rows]] = uniqs_index[ind]
+                counter = counter[~keep_rows]
+                left_c = left_c[~keep_rows]
+            if not counter.size > 0:
+                break
```
thatlittleboy: Just an observation that this is largely duplicated code, and wondering if we can refactor this down a little.

One thing is w.r.t. `ext_arr` boolean variable. Qn: If `extract_array` with numpy=True is already called for both `left_c` and `right_c`, would this variable ever be True?
What I understand from the pandas docs is that this call always extracts the ndarray underneath, so I'm missing when `left_c` will ever be a (pandas) extension array.

janitor/functions/conditional_join.py
```diff
@@ -1151,9 +661,493 @@ def _not_equal_indices(
...
+    # which are all in the original `df` and `right`
+    # doing this allows some speed gains
+    # while still ensuring correctness
+    left_c = extract_array(df[left_on], extract_numpy=True)[left_index]
+    right_c = extract_array(right[right_on], extract_numpy=True)[right_index]
+
+    mask = op(left_c, right_c)
+
+    if ext_arr:
+        mask = mask.to_numpy(dtype=bool, na_value=False)
```
thatlittleboy, line 1051: Same question here as above. (Pls ignore if I'm mistaken about the previous review)

thatlittleboy (changes requested):
Hey, sorry for the delay. The comments added from before were very helpful. Just a few more clarifying questions + formatting fixes on my end.

janitor/functions/conditional_join.py
samukweku: if the column is an extension array (Int64, Float64, categoricals), the numpy array wont be extracted (numpy does not recognize pd.NA for instance). `extract_array` will only return numpy arrays if the data is not an extension array.

samukweku: i'd probably be more strict and just avoid extension arrays, similar to what pd.merge_asof does; i'm just not sure it is the right approach in terms of restrictions. i am open to suggestions

janitor/functions/conditional_join.py
samukweku: yea, to save memory. might be handy for large arrays.

janitor/functions/conditional_join.py
thatlittleboy: I see. Thanks, didn't know about that behaviour of `extract_array`.

And when you compare two extension arrays (`op(left_c, right_c)`), is there a situation where you get an extension array back? -- The reason I'm asking is because that `to_numpy()` call is the only thing different between `ext_arr=True/False`. --
From my very limited testing, it seems if I compare 2 Categorical extensiondtype arrays, I get a numpy boolean array back.

```python
import operator
import pandas as pd
from pandas.core.construction import extract_array
from pandas.core.dtypes.common import is_extension_array_dtype


s1 = pd.Categorical(pd.Series([1,2,3.,pd.NA]), ordered=True)
s2 = pd.Categorical(pd.Series([2,3,1.,pd.NA]), ordered=True)
assert is_extension_array_dtype(s1)
assert is_extension_array_dtype(s2)

s1 = extract_array(s1, extract_numpy=True)
s2 = extract_array(s2, extract_numpy=True)

res = operator.le(s1, s2)
display(res)  # array([ True,  True, False, False])
is_extension_array_dtype(res)  # False
```

Edit: Hmm, ok there are Boolean extension arrays as well which pandas may use as the output of the comparison between 2 extension arrays.. (?)

thatlittleboy: How about just doing the if-else check inside, then we can eliminate the code duplication at least:
```python
for ind in range(uniqs_index.size):
    keep_rows = op(left_c, right_c[ind])
    if keep_rows.any():
        # [[ move the is extension array check here ]]
        if ext_arr:
            keep_rows = keep_rows.to_numpy(dtype=bool, na_value=False)
        pos[counter[keep_rows]] = uniqs_index[ind]
        counter = counter[~keep_rows]
        left_c = left_c[~keep_rows]
    if not counter.size > 0:
        break
```

Would this break anything?

Edit: Furthermore, I think we can refactor this further into a guard clause pattern instead, as in:
```python
    if not counter.size > 0:
        break
    if not keep_rows.any():
        continue
    # rest of the code, dedented by 1 level
```
?

thatlittleboy:
@samukweku Could you also update the CHANGELOG while you're at it as well? Thanks!

janitor/functions/conditional_join.py
samukweku: so in pandas 1.3.5 (which is the version we r using in the config - i am working in the docker env) , a pandas BooleanArray is returned, not an ndarray. And thanks for the observations, apparently categories are coerced to ndarrays, when they are compared :

```py
In [20]: import operator
    ...: import pandas as pd
    ...: from pandas.core.construction import extract_array
    ...: from pandas.core.dtypes.common import is_extension_array_dtype
    ...: 
    ...: 
    ...: s1 = pd.Series([1, 2, 3.0, pd.NA], dtype="Int64")
    ...: s2 = pd.Series([2, 3, 1.0, pd.NA], dtype="Int64")
    ...: assert is_extension_array_dtype(s1)
    ...: assert is_extension_array_dtype(s2)
    ...: 
    ...: s1 = extract_array(s1, extract_numpy=True)
    ...: s2 = extract_array(s2, extract_numpy=True)
    ...: 
    ...: res = operator.le(s1, 2)
    ...: display(res)  # array([ True,  True, False, False])
    ...: is_extension_array_dtype(res)  # False
<BooleanArray>
[True, True, False, <NA>]
Length: 4, dtype: boolean
Out[20]: True
```

I think I moved it out because I didnt want to pay the price of checking for the extension array each time (imagine so many iterations and for each one a check for extension array dtype), compared to doing the check only once (hence the duplicated code). still open to a more performant option though.


do you mind explaining the logic for the guard clause (for the existing code, if keep_rows.any() is False, nothing is executed right?)

pls, i appreciate this feedback and the hole poking (i feel bluntness is better for growth), so pls bring them on. this is a sure fire way for me to improve. so thank you!

thatlittleboy: 1. Ah yes, indeed. I kinda figured the BooleanArray (extension array) would pop up somewhere after I wrote the initial comment. Should've played around with more examples 😅. I think it's fine to keep support for extension arrays, we should embrace them as pandas v1+ become more prevalent.
2. Okay, I see your point about the if-check happening O(n) times -- I had just felt that an additional if-check using a precomputed boolean variable `ext_arr` shouldn't have that much of an effect on the runtime since the loop contains a more expensive computation. I got these results while I was experimenting (though it's probably not the right way to test, hah. 🤷🏼‍♂️) 
```python
import numpy as np
rng = np.random.default_rng(seed=42)
df = pd.DataFrame(rng.integers(0, 100, size=(50_000, 2)), columns=['a', 'b'])
s1, s2 = df['a'], df['b']
ext_arr = True

%%timeit -n3 -r5
for ind in range(len(s2)):
    keep_rows = s1> s2[ind]  # "expensive" computation
    if ext_arr: pass  # comment this out (?)

# with if-check:    6.89 s ± 380 ms per loop (mean ± std. dev. of 5 runs, 3 loops each)
# without if-check: 6.54 s ± 228 ms per loop (mean ± std. dev. of 5 runs, 3 loops each)
```
Regardless, I think it's largely just a matter of trade-off. Could we get one more opinion on this (@ericmjl ? 😄 ) and then I'll let this matter rest regardless, haha. Probably the more performant option will come from dropping down to numba so.. I'll still lean towards having cleaner code and DRY (don't-repeat-yourself) instead of having code duplication.

3. With respect to guard clauses, e.g. pattern described in general [here](https://medium.com/lemon-code/guard-clauses-3bc0cd96a2d3). You're right in saying the code in the if-block doesn't get executed if `~keep_rows.any()`, so performance- and functionality-wise, they are equivalent. It's more for ensuring code readability, as the "main" logic (when `keep_rows.any() == True`) is left un-nested.
```python
for ind in range(uniqs_index.size):
    if counter.size == 0:
        # this is actually checking the counter of the previous iteration (or, of the initial counter
        # before the loop started). It is 99% equivalent to having the check at the end of the
        # loop (the way it is now). If this check is placed first, we (a) exit the loop if the original
        # counter is alr empty; but (b) lose a "break" statement if the counter happens to be
        # empty at the end of the loop. But this is inconsequential since the loop has already
        # ended.
        break
    keep_rows = op(left_c, right_c[ind])
    keep_rows = keep_rows.to_numpy(dtype=bool, na_value=False)
    if not keep_rows.any():
        # this is the main guard clause logic. It is just the inverse of the existing logic, so the
        # code after this block is NOT executed if keep_rows.any() is False. Functionally, it
        # makes no difference whether you do it this way or leave it the way it is now, but...
        continue
    # the benefit is this code block (which is where the loop should end up most of the time,
    # and contains the rest of the "heavy" logic) is unnested. This gives you more "room"
    # to work with before reaching the PEP8 char limit; 
    pos[counter[keep_rows]] = uniqs_index[ind]
    counter = counter[~keep_rows]
    left_c = left_c[~keep_rows]
```
But this was kind of an aside comment / suggestion. I think it's also fine not to use guard clauses specifically in this case as the "main block" is quite minimal.

ericmjl: I guess I have to adjudicate? :joy_cat:

Regarding guard clauses: The pattern is a good one, definitely for readability and maintainability. Thanks, @thatlittleboy for finding the definition and explaining it. I now have another word in my vocabulary to name the good programming pattern. @samukweku, would you be open to modifying your implementation? It'd definitely be an upgrade for readability.

Regarding the point I was actually asked to adjudicate :joy_cat:, I actually am a bit lost. (ahhhh!) Is it the case that the guard clause pattern induces the extra check? Without the details in my mental RAM, I'd default to the following principles to decide:

1. Reduce duplication,
2. Increase readability,
3. Document extensively and locally,
4. Flat is better than nested,
5. If there's little practical difference in runtime, then I guess we're okay :smile:.

thatlittleboy: 😂 The simple tl;dr is: 1) We have a codeblock that has largely duplicated logic (the highlighted code block at the very beginning of this thread). 2) I was making the proposal to shift the `if ext_arr` logic *inside* the for-loop, so we can avoid the duplication of code (a comparison between the if- and else- block shows you the only difference is the conversion of the extension array into a numpy array).

So either we: 1) leave it as it was -- mostly duplicated logic, but do if-check only once outside the loop, or 2) shift the if check inside resulting in O(n) number of if-checks (possibly slower, but I doubt so in any meaningful fashion), but we achieve DRY-er code.

thatlittleboy (approved):
Approving the PR! I'll merge by the end of the day if there are no further comments 😊

PR #1009, opened 2022-02-05
Ensure that dataframe is not mutated on sort_column_value_order.

### LINKED ISSUE #1008: [BUG] Unexpected mutation of original df by `sort_column_value_order`
# Brief Description
The `sort_column_value_order` function adds a new column `cond_order` to the original dataframe (see [here](https://github.com/pyjanitor-devs/pyjanitor/blob/9d3653f959f11150bcc9aca87476efc72affc60a/janitor/functions/sort_column_value_order.py#L55)) for implementation purposes.

But this is mutating the original dataframe in a non-expected way -- the end user shouldn't be seeing this new column since `cond_order` is only used in the internal implementation of the sorting function.

# Minimally Reproducible Code

```python
df = pd.DataFrame({
    "a1": [4, 2, 9, 12],
    "a2": ["x", "y", "z", "y"],
})
df.sort_column_value_order("a2", {"x": 3, "y": 2, "z": 1});
df

# original df now has an additional column...
   a1 a2  cond_order
0   4  x           3
1   2  y           2
2   9  z           1
3  12  y           2
```

ericmjl:
Closes #1008.


ericmjl:
🚀 Deployed on https://deploy-preview-1009--pyjanitor.netlify.app

janitor/functions/sort_column_value_order.py
```diff
@@ -50,6 +51,7 @@ def sort_column_value_order(
         Dataframe, or if column_value_order dictionary is empty.
     :return: This function returns a Pandas DataFrame
```
thatlittleboy: Just a minor suggestion to make the description of the output df a little clearer / accurate.
```suggestion
    :return: A sorted pandas DataFrame.
```

janitor/functions/sort_column_value_order.py
```diff
@@ -50,6 +51,7 @@ def sort_column_value_order(
...
     """
+    df = df.copy()
```
thatlittleboy: A couple of improvements I think we can incorporate here:
- Since we are already doing a copy here, we can avoid the use of `new_df` later on, and just use `df` all the way.
- Actually I think this is one example where guard clauses really shine; -> I suggest we do all the relevant type & value checks (using our `check` and `check_column` util functions) and raise all the errors we can first, *before* doing the `df.copy()` and the rest of the sorting logic in an unnested block. This way, we avoid copying df if the input is invalid.
- And the `del new_df['cond_order']` line can be refactored out of the if-else block, to reduce code duplication.

thatlittleboy (changes requested):
I think this is one example where we can put the recently-mentioned guard clause pattern to good use! :)

janitor/functions/sort_column_value_order.py
ericmjl: Great point, thanks @thatlittleboy!

ericmjl:
Great catches, @thatlittleboy, thanks for the review. If you think the PR is good, please feel free to do the honors!

janitor/functions/sort_column_value_order.py
```diff
@@ -48,19 +51,13 @@ def sort_column_value_order(
     :param columns: This is a list of additional columns that we can sort by
     :raises ValueError: raises error if chosen Column Name is not in
         Dataframe, or if column_value_order dictionary is empty.
-    :return: This function returns a Pandas DataFrame
+    :return: A sorted pandas DataFrame.
     """
+    check_column(df, column, present=True)
     if len(column_value_order) > 0:
```
thatlittleboy: The way to implement guard clause for this particular if-else check will be something like this:

```python
check("column_value_order", column_value_order, [dict])  # I would probably just throw this in just in case
if not column_value_order:  # once we ensure this var is a dict, a "not var" is the pythonic way to check it is empty
    raise ValueError("column_value_order dictionary cannot be empty")
# then the remainder of the logic here
```

janitor/functions/sort_column_value_order.py
```diff
@@ -48,19 +51,13 @@ def sort_column_value_order(
...
-            if columns is None:
-                new_df = df.sort_values("cond_order")
-                del new_df["cond_order"]
-            else:
-                new_df = df.sort_values(columns + ["cond_order"])
-                del new_df["cond_order"]
-            return new_df
-        else:
-            raise ValueError("Column Name not in DataFrame")
+        df = df.assign(cond_order=df[column].replace(column_value_order))
```
thatlittleboy: Good call on using assign here :)

janitor/functions/sort_column_value_order.py
```diff
@@ -1,44 +1,47 @@
...
 import pandas_flavor as pf
 import pandas as pd
+from janitor.utils import check_column
+from .remove_columns import remove_columns  # noqa: F401
 
 
 @pf.register_dataframe_method
 def sort_column_value_order(
-    df: pd.DataFrame, column: str, column_value_order: dict, columns=None
+    df: pd.DataFrame, column: str, column_value_order: dict, columns=[]
```
thatlittleboy: Ah, no, this is probably a bad idea. (see [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)).

In general, the safer pattern is to initialize with `columns=None` and `if columns is None: columns = []` in the function body. This ensures that `columns`, if not provided by the user during the function call, is always the empty list every function call.

In this case, I think we can probably leave it as it was originally (separate if-else blocks), or have something like:
```python
columns = [] if columns is None else columns
```
and do what you did. I'll leave it to you to decide which you think is cleaner :)

thatlittleboy (changes requested):
There is just one more common-gotcha with regards to mutable default arguments that we have to fix (I think will cause serious bugs if we don't address this).

Hope you understand!

janitor/functions/sort_column_value_order.py
ericmjl: Good call! I'm going to use the latter.

ericmjl: Great catch! Made the change.

janitor/functions/sort_column_value_order.py
ericmjl: Great catch, thanks @thatlittleboy!

thatlittleboy (approved):
LGTM now, thanks for the updates! (There seems to be a codecov warning though, I suppose we should address it first before merging, yeah?)

ericmjl:
Just bumped up the code diff coverage!

PR #1014, opened 2022-02-09
[BUG] Extend `fill_empty`'s `column_names` type range

### LINKED ISSUE #998: [BUG] Extend `fill_empty`'s `column_names` type range
# Brief Description

https://github.com/pyjanitor-devs/pyjanitor/blob/3fab49e8c89f1a5e4ca7a6e4fdbbe8e2f7b89c66/janitor/functions/fill.py#L148-L152

Quickly fix this, could add `pd.Index`.

And a little bit more thinking, using `Iterable` is better?
Because `check_column`'s `column_names` also support `Iterable`.
And `pd.Index`'s instance is also `Iterable`.
So it would be `@dispatch(pd.DataFrame, Iterable)`

The `Iterable` is not `typing.Iterable` is `collections.abc.Iterable`.
Or else using `typing.Iterable` would get another error.

# Minimally Reproducible Code

```python
>>> import pandas as pd
>>> import janitor  # noqa

>>> df = pd.DataFrame({"attr":[None, None]})
>>> df.fill_empty(df.columns, 0)
  File "C:\Software\miniforge3\envs\work\lib\site-packages\multipledispatch\dispatcher.py", line 269, in __call__
    func = self._cache[types]
KeyError: (<class 'pandas.core.frame.DataFrame'>, <class 'pandas.core.indexes.base.Index'>)

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "C:\Software\miniforge3\envs\work\lib\site-packages\pandas_flavor\register.py", line 29, in __call__
    return method(self._obj, *args, **kwargs)
  File "C:\Software\miniforge3\envs\work\lib\site-packages\janitor\utils.py", line 231, in wrapper
    return func(*args, **kwargs)
  File "C:\Software\miniforge3\envs\work\lib\site-packages\janitor\functions\fill.py", line 199, in fill_empty
    return _fill_empty(df, column_names, value=value)
  File "C:\Software\miniforge3\envs\work\lib\site-packages\multipledispatch\dispatcher.py", line 273, in __call__
    raise NotImplementedError(
NotImplementedError: Could not find signature for _fill_empty: <DataFrame, Index>
```

Zeroto521:
# PR Description

- Let `column_names` support more type range, from `list` & `tuple` to `Iterable`.

**This PR resolves #998.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/tests/functions/test_fill_empty.py
+++ b/tests/functions/test_fill_empty.py
@@ -1,4 +1,5 @@
 import pytest
+import pandas as pd
 
 
 @pytest.mark.functions
@@ -11,3 +12,22 @@ def test_fill_empty(null_df):
 def test_fill_empty_column_string(null_df):
     df = null_df.fill_empty(column_names="2", value=3)
     assert set(df.loc[:, "2"]) == set([3])
+
+
+@pytest.mark.functions
+@pytest.mark.parametrize(
+    "column_names",
+    [
+        (0, 1, "2", "3"),  # tuple
+        [0, 1, "2", "3"],  # list
+        {0, 1, "2", "3"},  # set
+        ({0: 0, 1: 1, "2": "2", "3": "3"}).keys(),  # dict key
+        ({0: 0, 1: 1, "2": "2", "3": "3"}).values(),  # key value
+        pd.Index([0, 1, "2", "3"]),  # Index
+    ],
+)
+def test_column_names_iterable_type(null_df, column_names):
+    result = null_df.fill_empty(column_names=column_names, value=3)
+    excepted = null_df.fillna(3)
+
+    assert result.equals(excepted)
--- a/tests/functions/test_fill_empty.py
+++ b/tests/functions/test_fill_empty.py
@@ -22,7 +22,7 @@ def test_fill_empty_column_string(null_df):
         [0, 1, "2", "3"],  # list
         {0, 1, "2", "3"},  # set
         ({0: 0, 1: 1, "2": "2", "3": "3"}).keys(),  # dict key
-        ({0: 0, 1: 1, "2": "2", "3": "3"}).values(),  # key value
+        ({0: 0, 1: 1, "2": "2", "3": "3"}).values(),  # dict value
         pd.Index([0, 1, "2", "3"]),  # Index
     ],
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 -   [ENH] Add `xlsx_table`, for reading tables from an Excel sheet. @samukweku
 -   [ENH] minor improvements for conditional_join; equality only joins are no longer supported; there has to be at least one non-equi join present. @samukweku
 -   [BUG] `sort_column_value_order` no longer mutates original dataframe.
+-   [BUG] Extend `fill_empty`'s `column_names` type range. Issue #998. @Zeroto521
 
 ## [v0.22.0] - 2021-11-21
 
--- a/janitor/functions/fill.py
+++ b/janitor/functions/fill.py
@@ -145,6 +145,7 @@ def fill_empty(
     check_column(df, column_names)
     return _fill_empty(df, column_names, value=value)
 
+
 @dispatch(pd.DataFrame, abcIterable)
 def _fill_empty(df, column_names, value=None):
     """Fill empty function for the case that column_names is list or tuple."""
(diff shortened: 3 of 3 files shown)


tests/functions/test_fill_empty.py
```diff
@@ -11,3 +12,22 @@ def test_fill_empty(null_df):
...
+@pytest.mark.parametrize(
+    "column_names",
+    [
+        (0, 1, "2", "3"),  # tuple
+        [0, 1, "2", "3"],  # list
+        {0, 1, "2", "3"},  # set
+        ({0: 0, 1: 1, "2": "2", "3": "3"}).keys(),  # dict key
+        ({0: 0, 1: 1, "2": "2", "3": "3"}).values(),  # key value
+        pd.Index([0, 1, "2", "3"]),  # Index
+    ],
```
Zeroto521, line 27: `map(lambda x: x if x < 2 else str(x), range(4)),  # generator`

The generator is also `Iterable`, but this wouldn't work.

samukweku (approved):

tests/functions/test_fill_empty.py
ericmjl: That's a valuable thing to note. @Zeroto521 does the function raise an error when a generator is passed in? If so, then we may want to have a test for that case where we explicitly ask `pytest` to expect an error when passing in a generator.

thatlittleboy (approved):

ericmjl:
@Zeroto521 I'd love to merge, but before we do so, would love to hear your thoughts on the above discussion point!

tests/functions/test_fill_empty.py
Zeroto521: https://github.com/pyjanitor-devs/pyjanitor/blob/d3146bbf0dddcb0cd979691bd171be69e12c0af5/janitor/functions/fill.py#L144-L145

`map(lambda x: x if x < 2 else str(x), range(4))` is `map` type and it can only use once.
The `check_column` consumes `column_names` data.

Zeroto521: One solution.

Let `check_column` return a new `list` type `column_names`.

```python
def check_column( 
    df: pd.DataFrame,
    column_names: Union[Iterable, str, int],
    present: bool = True,
) -> List[Union[str, int]]: 
    if isinstance(column_names, (str, int)) or not isinstance(column_names, Iterable): 
        column_names = [column_names] 
    else:
        column_names = list(column_names)

    for column_name in column_names: 
        if present and column_name not in df.columns:  # skipcq: PYL-R1720 
            raise ValueError( 
                f"{column_name} not present in dataframe columns!" 
            ) 
        elif not present and column_name in df.columns: 
            raise ValueError( 
                f"{column_name} already present in dataframe columns!" 
            )

    return column_names
```

Zeroto521: Another solution.

Copy `column_names`

```python
from copy import deepcopy

 check_column(df, deepcopy(column_names)) 
 return _fill_empty(df, column_names, value=value)
```

ericmjl: @thatlittleboy @samukweku what do you all think? I think I'd be in favour of the `deepcopy` implementation over re-casting to list; something feels wonky to do change of types, but I might be over-sensitive here.

thatlittleboy: Deep-copying a generator doesn't seem to work (error is thrown about not being able to pickle a generator).
The canonical way is to convert the generator into a list to persist it, so I think I'll be in favour of that. But I suggest we leave this to a separate PR (changing the "api" of `check_columns` to take care of generators properly).

We may add in the additional test for this function in that PR as well once we've modified `check_columns`.

ericmjl: Thanks @thatlittleboy! Good catch. Passing in generators seems like a problem that has low probability of occurring in real life usage. I think it's okay for us to follow @thatlittleboy's guidance and leave this issue to be a separate PR. 

@Zeroto521, could you raise a new issue for us on this matter? Meanwhile, I am going to merge.

ericmjl:
Thank you, @Zeroto521 for this contribution!

tests/functions/test_fill_empty.py
Zeroto521: There is a new issue #1015 to trace this.

PR #1017, opened 2022-02-12
[TST] Fix clean names tests

thatlittleboy:
Tests are failing for `clean_names` on the latest commit due to the Data on wikipedia being modified. (someone removed the "Current account balance (in % of GDP)" column from the wiki table).

## Changes
- Use a fixed dataframe for the `test_char` test instead of relying on the live table from wikipedia (for stability of our tests).
- Removed the `test_space` test (which also relied on the Russia GDP wikipedia page); this test can already be covered by the `test_char` test function.
- Removed the use of `hypothesis` in the `test_clean_names` test suite; It is redundant since we only care about the column names, not the df data.

### DIFF
--- a/tests/functions/test_clean_names.py
+++ b/tests/functions/test_clean_names.py
@@ -1,15 +1,13 @@
 import pandas as pd
 import pytest
-from hypothesis import given
 
 from janitor.errors import JanitorError
-from janitor.testing_utils.strategies import df_strategy
 
 
 @pytest.mark.functions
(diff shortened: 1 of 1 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1017--pyjanitor.netlify.app

ericmjl (approved):
Wonderful catch, thank you, @thatlittleboy!

PR #1018, opened 2022-02-12
[ENH] Fix round_to_fraction default arguments

### LINKED ISSUE #1005: [BUG] Unhelpful default arguments in `round_to_fraction`
# Brief description

## [`round_to_fraction`](https://github.com/pyjanitor-devs/pyjanitor/blob/9d3653f959f11150bcc9aca87476efc72affc60a/janitor/functions/round_to_fraction.py#L11) function
- The `column_name` parameter is currently given a non-useful default argument (`None`) while the function logic dictates that it should be an existing column within the dataframe. I propose to: 1) make this parameter non-optional, and 2) add an additional check to enforce that the given column name is present in the dataframe.
- Similarly, the `denominator` parameter is currently given a non-useful default `None`, while it should clearly be a number. Propose to either make it non-optional since I'm unsure if there is a sensible default value to place here.

### Minimally Reproducible Code

```python
import numpy as np
import pandas as pd
import janitor
df = pd.DataFrame({
    "a1": [1.203, 2.499, np.nan],
    "a2": ["x", "y", "z"],
})

df.round_to_fraction()  # KeyError: None
df.round_to_fraction("a1")  # TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
```

The first error should be replaced with a more informative errmsg as per the improvements suggested for `column_name` above.

The second error should be caught earlier (at the moment, the type check of the `denominator` parameter is only done if `denominator` is Truthy, which is not really accurate.)

thatlittleboy:
**This PR resolves #1005**.

Also add in additional check (and tests) for when denominator is passed in a value <= 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 -   [ENH] minor improvements for conditional_join; equality only joins are no longer supported; there has to be at least one non-equi join present. @samukweku
 -   [BUG] `sort_column_value_order` no longer mutates original dataframe.
 -   [BUG] Extend `fill_empty`'s `column_names` type range. Issue #998. @Zeroto521
+-   [BUG] Removed/updated error-inducing default arguments in `row_to_names` (#1004) and `round_to_fraction` (#1005). @thatlittleboy
 
 ## [v0.22.0] - 2021-11-21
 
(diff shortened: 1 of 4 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1018--pyjanitor.netlify.app

thatlittleboy:
ps. The tests will fail on the first pass due to #1017 . We can rerun the tests here again after #1017 is merged.

samukweku (approved):

ericmjl (approved):

ericmjl:
Wonderful work, thank you so much, @thatlittleboy!

Issue #1016, opened 2022-02-11
[BUG] Support `Hashable` type column


Zeroto521:
# Brief Description

The type of DataFrame's columns is support `Hashable`.
But Janitor only handles `str` type columns.

# Related Source Codes List

- Fix bugs
    - [ ] [janitor.utils.check_column](https://github.com/pyjanitor-devs/pyjanitor/blob/0f98b8874a9bea1652b0f934fd086f64131a675b/janitor/utils.py#L298-L300)
    - [ ] [janitor.fill.fill_empty](https://github.com/pyjanitor-devs/pyjanitor/blob/0f98b8874a9bea1652b0f934fd086f64131a675b/janitor/functions/fill.py#L156-L157)
- Update type annotations
    - [ ] [janitor.utils.check_column](https://github.com/pyjanitor-devs/pyjanitor/blob/0f98b8874a9bea1652b0f934fd086f64131a675b/janitor/utils.py#L298-L300)
    - [ ] [janitor.fill.fill_empty](https://github.com/pyjanitor-devs/pyjanitor/blob/0f98b8874a9bea1652b0f934fd086f64131a675b/janitor/functions/fill.py#L113-L115)
    - [ ] [janitor.functions.encode_categorical](https://github.com/pyjanitor-devs/pyjanitor/blob/0f98b8874a9bea1652b0f934fd086f64131a675b/janitor/functions/encode_categorical.py#L12-L16)
    - [ ] and more

# Minimally Reproducible Code

```python
>>> import pandas as pd
>>> import janitor  # noqa

>>> df = pd.DataFrame({0:[None, None]})
>>> df.columns
Int64Index([0], dtype='int64')

>>> df.fill_empty(0, 0)
Traceback (most recent call last):
  File "C:\Software\miniforge3\envs\work\lib\site-packages\multipledispatch\dispatcher.py", line 269, in __call__
    func = self._cache[types]
KeyError: (<class 'pandas.core.frame.DataFrame'>, <class 'int'>)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Software\miniforge3\envs\work\lib\site-packages\pandas_flavor\register.py", line 29, in __call__
    return method(self._obj, *args, **kwargs)
  File "C:\Software\miniforge3\envs\work\lib\site-packages\janitor\utils.py", line 231, in wrapper
    return func(*args, **kwargs)
[...]

PR #1028, opened 2022-02-23
[DOC] MWE for `label_encode`, `factorize_columns`, `encode_categorical`

thatlittleboy:
# PR Description

- Add MWE for the 3 functions mentioned in the title;
- Add note of (deprecation) warning to the docs of `label_encode`.
- Make `factorize_columns` non-mutating.

**This PR is related to #972 .**

<s>(Let's review this after #1029 has been merged. I'll deal with any merge conflicts then.)</s>
 (`encode_categorical` MWE has been incorporated in #1029)

### DIFF
--- a/janitor/functions/factorize_columns.py
+++ b/janitor/functions/factorize_columns.py
@@ -24,7 +24,7 @@ def factorize_columns(
     It takes in an optional suffix and keyword arguments also.
     An empty string as suffix will override the existing column.
 
-    This method mutates the original DataFrame.
+    This method does not mutate the original DataFrame.
 
     Example:
 
@@ -60,5 +60,5 @@ def factorize_columns(
 
     :returns: A pandas DataFrame.
     """
-    df = _factorize(df, column_names, suffix, **kwargs)
+    df = _factorize(df.copy(), column_names, suffix, **kwargs)
     return df
--- a/janitor/functions/factorize_columns.py
+++ b/janitor/functions/factorize_columns.py
@@ -1,3 +1,4 @@
+"""Implementation of the `factorize_columns` function"""
 from typing import Hashable, Iterable, Union
 import pandas_flavor as pf
 import pandas as pd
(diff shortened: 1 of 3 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1028--pyjanitor.netlify.app

janitor/functions/factorize_columns.py
```diff
@@ -13,50 +14,49 @@ def factorize_columns(
...
     This method will create a new column with the string `_enc` appended
     after the original column's name.
     This can be overriden with the suffix parameter.
 
-    Internally this method uses pandas `factorize` method.
+    Internally, this method uses pandas `factorize` method.
     It takes in an optional suffix and keyword arguments also.
     An empty string as suffix will override the existing column.
 
     This method mutates the original DataFrame.
```
samukweku: Should we make it non mutating? Does it matter? Is it outside d scope of this PR?

samukweku (approved):

janitor/functions/factorize_columns.py
thatlittleboy: Yea, as a matter of principle, I am of the opinion that we should make it non-mutating.
Though, that seems like a topic to be discussed in general. I seem to recall reading a discussion about this somewhere in a previous thread; I would opt to adjust this outside of this PR, and leave it for when we do revisit that topic again.

Or if we have already come to a conclusion about this before (and I missed it), then let's this change it in a future sprint (along with all the other functions that need changing). 😃

ericmjl: Agreed, non-mutating is a very good idea. 

I think the way to not mutate the original column is to use a `.assign(col_name=some_value)` instead of `df[col_name] = some_value`.

@thatlittleboy if you'd like to sneak in the change here, that'd be great! Otherwise, we can just raise another issue. :smile:.

ericmjl (approved):

janitor/functions/factorize_columns.py
samukweku: Just as an aside to be wary of `assign`, as it only supports strings. #1016

ericmjl: Oh! That’s right, good catch. Then copying the data frame seems to be the way out, I guess.

samukweku (approved):

ericmjl:
@samukweku please do the honors when the checks are done! :tada:

PR #1040, opened 2022-03-13
[DOC/ENH] MWE and fixes for `truncate_datetime`

thatlittleboy:
# PR Description

- add MWE.
- make `truncate_datetime_dataframe` not mutate original df.
- fix SettingWithCopyWarning; and typo bug with the "MINUTE:" key in `ENUM` dict of
    `_truncate_datetime` function.
- ditch the previous error-catching mechanism. Optimize to early exit if no datetime columns found in df; Only apply `_truncate_datetime` on datetime-like columns, and when NaT is encountered, it is ignored.
- small optimization to exit immediately if wrong datepart is passed in, instead of carrying on iterating
    through all elements.
- not iterating over rows, and instead using `.applymap` on relevant columns only yields about 60x improvement. (5+ seconds -> 80ms on a df with 10k timestamps).
- fix warning about use of `pd.np` in tests for `to_datetime`.
- make tests more complete (test all dateparts instead of just MONTH, and test specifically when there is `pd.NaT`)
- administrative: collate the tests into just 1 file for this `truncate_datetime` function.

**This PR is related to #972**


ericmjl:
🚀 Deployed on https://deploy-preview-1040--pyjanitor.netlify.app

samukweku (approved):

janitor/functions/truncate_datetime.py
```diff
@@ -15,46 +17,79 @@ def truncate_datetime_dataframe(
...
+    datepart = datepart.upper()
+    if datepart not in ACCEPTABLE_DATEPARTS:
+        raise ValueError(
+            "Received an invalid `datepart` precision. "
+            f"Please enter any one of {ACCEPTABLE_DATEPARTS}."
+        )
+
+    df = df.copy()
+
+    for col in df.columns:
```
samukweku: Apologies, I know I approved ... Just couldn't hake this off, what is the purpose of iterating through the index? Isn't the same function applied to every row?

thatlittleboy: No worries,
Yeah, it is. Are you implying we can use `.apply` here?

(I originally started this PR only with the intention to add a MWE, but then realized there were bugs and warnings being thrown, so just added in some additional fixes without really intending to optimize the code any further)

samukweku: Yea, apply or if there is a faster option.

thatlittleboy: Ok, I've modified it to use applymap so it's much faster now (5s -> 80ms on a df with 10k timestamps), and skip the df copying if a no-op is expected (when there are no datetime columns).

samukweku: Sweeeeeet

ericmjl:
> - not iterating over rows, and instead using .applymap on relevant columns only yields about 60x improvement. (5+ seconds -> 80ms on a df with 10k timestamps).

😮! That is impressive, @thatlittleboy!

ericmjl (approved):
LGTM and I think we can merge. Thank you @thatlittleboy! @samukweku would you like to do the honors?

thatlittleboy:
Sorry, hold on one sec before merging, just realized one of the test docstring wasnt updated. i'll Need to push a quick fix

EDIT: okays, done!

PR #1036, opened 2022-03-05
[INF] Try to fix dependabot alerts

thatlittleboy:
**Dependabot error:** Could not find version that satisfies...

```
ERROR: Could not find a version that satisfies the requirement scipy==1.7.1 (from -r base.in (line 8)) (from versions: 0.8.0, 0.9.0, 0.10.0, 0.10.1, 0.11.0, 0.12.0, 0.12.1, 0.13.0, 0.13.1, 0.13.2, 0.13.3, 0.14.0, 0.14.1, 0.15.0, 0.15.1, 0.16.0, 0.16.1, 0.17.0, 0.17.1, 0.18.0, 0.18.1, 0.19.0, 0.19.1, 1.0.0b1, 1.0.0rc1, 1.0.0rc2, 1.0.0, 1.0.1, 1.1.0rc1, 1.1.0, 1.2.0rc1, 1.2.0rc2, 1.2.0, 1.2.1, 1.2.2, 1.2.3, 1.3.0rc1, 1.3.0rc2, 1.3.0, 1.3.1, 1.3.2, 1.3.3, 1.4.0rc1, 1.4.0rc2, 1.4.0, 1.4.1, 1.5.0rc1, 1.5.0rc2, 1.5.0, 1.5.1, 1.5.2, 1.5.3, 1.5.4, 1.6.0rc1, 1.6.0rc2, 1.6.0, 1.6.1, 1.7.2, 1.7.3, 1.8.0rc1, 1.8.0rc2)
Traceback (most recent call last):
  File "/usr/local/.pyenv/versions/3.10.0/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/.pyenv/versions/3.10.0/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/.pyenv/versions/3.10.0/lib/python3.10/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/local/.pyenv/versions/3.10.0/lib/python3.10/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/.pyenv/versions/3.10.0/lib/python3.10/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/.pyenv/versions/3.10.0/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func
[...]

PR #1041, opened 2022-03-13
[INF] Attempt to fix dependabot alerts again

thatlittleboy:
Dependabot alerts are complaining about versions of `urllib`, `numpy` and `ipython`. This PR attempts to fix the warnings.
As long as the versions are outside the specified versions in the warnings, it should resolve the dependabot alerts.

Follow-up from #1036 

You may just refer to the change list below:

Changes:
* Add `pip-tools` and remove `nox` from the `environment.yml` file. (since we're using `pytest` as our testing framework now)
* Add a  `compile-requirements` target in the Makefile, making use of the above-installed `pip-compile` (for ease of maintaining requirements.txt's in the future). This target was also used to update all the `.requirements/xxx.txt` files in this PR.
* The relevant changes in the `.in` files:
   * removed `sphinx` dependencies, which also removed dependency for `urllib`; and replaced with `mkdocs` and co.
   * constrained `mkdocstrings` version < 0.18 (to be fixed in future)
   * `numpy` and `ipython` are upgraded as per normal.


ericmjl:
🚀 Deployed on https://deploy-preview-1041--pyjanitor.netlify.app

samukweku (approved):

ericmjl (approved):
Awesome, @thatlittleboy! Let's see how this works :smile:.

PR #1046, opened 2022-03-18
[ENH] xlsx_cells

### LINKED ISSUE #929: Implement tidyxl for Excel files

samukweku:
# PR Description

- read in xlsx file without coercing into rectangle form
- each cell is assigned a row with relevant information (row number, column number, cell formatting, ...)
- updates to xlsx_table and read_commandline
- doctests added for xlsx_table and xlsx_cells

**This PR resolves #929 .**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1046--pyjanitor.netlify.app

janitor/io.py
```diff
@@ -99,18 +102,17 @@ def read_commandline(cmd: str, **kwargs) -> pd.DataFrame:
...
-    if outcome.returncode != 0:
-        raise JanitorError(outcome.stderr)
-    else:
-        outcome = outcome.stdout
-    return pd.read_csv(StringIO(outcome), **kwargs)
+    # adding check=True ensures that an explicit, clear error
+    # is raised, so that the user can see the reason for the failure
+    outcome = subprocess.run(
+        cmd, shell=True, capture_output=True, text=True, check=True
+    )
```
thatlittleboy, line 116: Missed this during the last PR when `read_commandline` was added, but just wanted to confirm if the plan for the next release of pyjanitor is to only support py3.7+ ? @ericmjl 

It's relevant here because `capture_output` and `text` were both only recently added in py3.7. The legacy equivalent are `stdout=subprocess.PIPE`, `stderr=subprocess.PIPE` and `universal_newlines`, which should be used if we still want to support py3.6.

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+        3           4    Competition
+        4           5  Long Distance
+        >>> out['dSalesReps'].head(3)
+           SalesRepID             SalesRep Region
+        0           1  Sioux Radcoolinator     NW
+        1           2        Tyrone Smithe     NE
+        2           3         Chantel Zoya     SW
+
+
+    :param path: Path to the Excel File. It can also be an openpyxl Workbook.
```
thatlittleboy, line 171: Let's update the annotation in the function signature as well? If Workbook is allowed.

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
             pip_install=True,
         )
-    wb = load_workbook(filename=path, read_only=False, keep_links=False)
-    check("sheetname", sheetname, [str])
-    ws = wb[sheetname]
+    if isinstance(path, Workbook):
+        if path.read_only:
+            raise ValueError(
+                "Accessing the tables require 'read_only' to be False."
+            )
```
thatlittleboy: is this check for read-only necessary?

If `path.read_only=True`, I'm assuming openpyxl would complain on their own if we try to run the `path[sheetname]` ? If so, I suggest leaving the error handling to them, unless their error message is very cryptic.

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
-                    3         DB        Darnell Booms  Burlington    VT            Darnell@Darnell.com}
-    ```
-
-    :param path: Path to the Excel File.
+        >>> import pandas as pd
+        >>> from janitor import xlsx_table
+        >>> filename = "../pyjanitor/tests/test_data/016-MSPTDA-Excel.xlsx"
+
+        # single table
+        >>> xlsx_table(filename, sheetname='Tables', table = 'dCategory')
```
thatlittleboy: Could you run `black` on all of these pycon examples? I think the spaces around the `=` should be removed, and to use double quotes instead of single quotes. (just to maintain a consistent look across the documentation)

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+        2       1              1         A2    2       1         n    False       General
+        3       2              2         B2    2       2         n    False       General
+        4       3              3         A3    3       1         n    False       General
+        5       4              4         B3    3       2         n    False       General
+        6       5              5         A4    4       1         n    False       General
+        7       6              6         B4    4       2         n    False       General
+
+        # Access cell formatting such as fill :
+
+        >>> out = xlsx_cells(filename, sheetnames = 'highlights', fill=True).select_columns('value', 'fill')
```
thatlittleboy: Same comments here about code formatting (spaces around the `=`).

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+        1  Height         Height         B1    1       2         s    False       General
+        2       1              1         A2    2       1         n    False       General
+        3       2              2         B2    2       2         n    False       General
+        4       3              3         A3    3       1         n    False       General
+        5       4              4         B3    3       2         n    False       General
+        6       5              5         A4    4       1         n    False       General
+        7       6              6         B4    4       2         n    False       General
+
+        # Access cell formatting such as fill :
+
```
thatlittleboy: Can we delete this empty line between the comment and the code? It renders as an empty line in the documentation (unnecessary).
And also to keep formatting consistent with the example above.

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+        2       1     {'patternType': None, 'fgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}, 'bgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}}
+        3       2     {'patternType': None, 'fgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}, 'bgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}}
+        4       3  {'patternType': 'solid', 'fgColor': {'rgb': 'FFFFFF00', 'type': 'rgb', 'tint': 0.0}, 'bgColor': {'rgb': 'FFFFFF00', 'type': 'rgb', 'tint': 0.0}}
+        5       4  {'patternType': 'solid', 'fgColor': {'rgb': 'FFFFFF00', 'type': 'rgb', 'tint': 0.0}, 'bgColor': {'rgb': 'FFFFFF00', 'type': 'rgb', 'tint': 0.0}}
+        6       5     {'patternType': None, 'fgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}, 'bgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}}
+        7       6     {'patternType': None, 'fgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}, 'bgColor': {'rgb': '00000000', 'type': 'rgb', 'tint': 0.0}}
+
+        # specific cell attributes can be accessed by using Pandas' series.str.get :
+
```
thatlittleboy: Can we delete this empty line between the comment and the code? It renders as an empty line in the documentation (unnecessary).
And also to keep formatting consistent with the example above.

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+
+    :param path: Path to the Excel File. It can also be an openpyxl Workbook.
+    :param sheetnames: Names of the sheets from which the cells are to be extracted.
+        If `None`, all the sheets in the file are extracted;
+        if it is a string, or list or tuple, only the specified sheets are extracted.
+    :param start_point: start coordinates of the Excel sheet. This is useful
+        if the user is only interested in a subsection of the sheet.
+    :param end_point: end coordinates of the Excel sheet. This is useful
+        if the user is only interested in a subsection of the sheet.
```
thatlittleboy: The way the code is written now, it seems you expect the end-user to either provide both, or provide neither, right? Might be worth specifying this restriction in the description for both params.

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+                path,
+                sheetname,
+                defaults,
+                parameters,
+                start_point,
+                end_point,
+                include_blank_cells,
+            )
+            for sheetname in path.sheetnames
+        }
```
thatlittleboy: There's a lot of duplicate code here, I'm thinking if we can make this a little more compact in the same vein as `xlsx_table`?
How about
```python
if not sheetnames:
    sheetnames = path.sheetnames
elif isinstance(sheetnames, str):
    sheetnames = [sheetnames]
else:
    check("sheetnames", sheetnames, [str, list, tuple])

out = {sheetname: _xlsx_cells(...) for sheetname in sheetnames}
if len(out) == 1:
    _, out = out.popitem()
```

By the way, is there ever a situation where `out` is empty `{}` here, or would this be avoided by the time we reach this point?

janitor/io.py
```diff
@@ -132,91 +134,412 @@ def xlsx_table(
...
+    end_point: Union[str, int],
+    include_blank_cells: bool,
+):
+    """
+    Function to process a single sheet.
+    Returns a DataFrame.
+
+    :param wb: Openpyxl Workbook.
+    :param sheetname: Name of the sheet
+        from which the cells are to be extracted.
```
thatlittleboy: How about just making the input here the worksheet itself, instead of workbook + sheetname? I don't see `wb` being used for any other purpose other than to access the worksheet.

tests/io/test_tidyxl.py
```diff
@@ -0,0 +1,171 @@
...
+def test_sheetname_presence():
+    """Raise error if sheet name does not exist."""
+    with pytest.raises(KeyError):
+        io.xlsx_cells(filename, "5")
+
+
+def test_comment_read_only():
+    """Raise error if comment and read_only is True."""
+    with pytest.raises(ValueError):
+        io.xlsx_cells(filename, "clean", comment=True)
```
thatlittleboy: For generic errors that we're raising (such as ValueError), let's try to use the `match` wherever possible? Because we're throwing ValueErrors for a lot of different reasons; we don't want to accidentally catch a ValueError from another error msg unknowingly and get a false negative.

Also, I understand `read_only=True` is default set to True, but I think if the test is explicitly meant to check the comment & read_only combination, that we should specify it explicitly in the function call here.

thatlittleboy (changes requested):
Sorry this took so long, :) Had to split up the review process over several nights.

janitor/io.py
samukweku: py3.6 is EOL and no longer supported; safe to say we can move on to 3.7 and above.

janitor/io.py
samukweku: would appreciate your advice on how to go about that, since at the moment we are importing the workbook in a try clause, which triggers an error message if openpyxl is not installed/not found

janitor/io.py
samukweku: this is why I love reviews ... fix flaws, make code better ... thanks @thatlittleboy for your selfless service ... really appreciate your efforts ... hope you don't get tired anytime soon

samukweku: not at all; out can't be empty - whatever issues would be resolved before then

janitor/io.py
samukweku: the user might have the workbook to access sheet names, hold on to a connection because he needs do more stuff; whereas within this function once we are done we close the connection

thatlittleboy: So, to be clear, my previous comment I was referring to the parameters of the private function `_xlsx_cells` rather than the public one.

I understood your point about maybe the user already having a workbook on hand, but I don't quite get where this --- "whereas within this function once we are done we close the connection" --- happens.

janitor/io.py
```diff
@@ -3,6 +3,7 @@
 from glob import glob
 from io import StringIO
 from typing import Iterable, Union, NamedTuple
+from openpyxl import Workbook
```
thatlittleboy: I think I saw a comment from you in the morning about this potentially causing problems since `openpyxl` is an optional import.

And I think it is a valid concern.

This is usually handled by using a `TYPE_CHECKING` variable from the `typing` module, which will only be True during type checking, and False otherwise. This is a common way to get around circular imports and/or optional imports like the situation we have here when dealing with importing objects only used for type checking, and nothing else.

```python
from __future__ import annotations  # only py3.7+
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from openpyxl import Workbook
```
Then we just need to make sure our testing environment has `openpyxl`, but it won't be a requirement when someone tries to import `janitor` (which in turn, imports `io`).

janitor/io.py
samukweku: @thatlittleboy if U don't mind, kindly post a visual of what U mean

samukweku: made an edit to the code; when you can, lemme know if that is what you have in mind.

samukweku:
something is broken in jinja, which is affecting mkdocs. Looks like a PR to fix it will soon be merged on Jinja's end.

janitor/io.py
thatlittleboy: @samukweku yep I think your changes are what I meant (having the parameters to `_xlsx_cells` as the worksheet itself rather than the full workbook + the sheetname.)
Thanks!

thatlittleboy (approved):
Thanks for being patient with my reviews! Other than the coverage issue (which honestly, I'm not quite sure how to fix), the rest LGTM!

ericmjl (approved):

PR #1050, opened 2022-03-19
[ENH] Allows multiple values_to when names_pattern is a list/tuple

### LINKED ISSUE #1048: `pivot_longer` support multiple melting with multiple `values_to`?
# Brief Description

I would like to propose extending `pivot_longer` to support multiple melting with multiple `values_to`.

[Source Example](https://stackoverflow.com/q/51519101/7175713)

# Example API

```python
df = pd.DataFrame({'City': ['Houston', 'Austin', 'Hoover'],
                   'State': ['Texas', 'Texas', 'Alabama'],
                   'Name':['Aria', 'Penelope', 'Niko'],
                   'Mango':[4, 10, 90],
                   'Orange': [10, 8, 14], 
                   'Watermelon':[40, 99, 43],
                   'Gin':[16, 200, 34],
                   'Vodka':[20, 33, 18]},
                 columns=['City', 'State', 'Name', 'Mango', 'Orange', 'Watermelon', 'Gin', 'Vodka'])

df

      City    State      Name  Mango  Orange  Watermelon  Gin  Vodka
0  Houston    Texas      Aria      4      10          40   16     20
1   Austin    Texas  Penelope     10       8          99  200     33
2   Hoover  Alabama      Niko     90      14          43   34     18

df.pivot_longer(index = ['City', 'State'], 
                column_names = slice('Mango', 'Vodka'), 
                names_to = ('Fruit', 'Drinks'), 
                values_to = ('Pounds', 'Ounces'), 
                names_pattern = [r"M|O|W", r"G|V"])

      City    State       Fruit  Pounds  Drink  Ounces
0   Austin    Texas       Mango      10    Gin   200.0
1   Hoover  Alabama       Mango      90    Gin    34.0
2  Houston    Texas       Mango       4    Gin    16.0
3   Austin    Texas      Orange       8  Vodka    33.0
4   Hoover  Alabama      Orange      14  Vodka    18.0
5  Houston    Texas      Orange      10  Vodka    20.0
6   Austin    Texas  Watermelon      99    NaN     NaN
7   Hoover  Alabama  Watermelon      43    NaN     NaN
8  Houston    Texas  Watermelon      40    NaN     NaN
```

samukweku:
# PR Description

- Allows multiple `values_to` when names_pattern is a list/tuple

**This PR resolves #1048.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1050--pyjanitor.netlify.app

janitor/functions/pivot.py
```diff
@@ -275,21 +318,24 @@ def _data_checks_pivot_longer(
...
-                    f"{word} is of type {type(word).__name__}."
-                )
-
-            if (word in unique_names_to) and (word != ".value"):
+            check(f"{word} in names_to", word, [str])
+            if (word in uniques) and (word != ".value"):
                 raise ValueError(f"{word} already exists in names_to.")
-            unique_names_to.add(word)  # noqa: PD005
-        unique_names_to = None
+            uniques.add(word)  # noqa: PD005
```
thatlittleboy:
```suggestion
            uniques.add(word)
```
PD005 ("Use arithmetic operator instead of method") is probably no longer relevant anymore here, let's remove it and see if the linter complains?

janitor/functions/pivot.py
```diff
@@ -16,7 +16,7 @@ def pivot_longer(
     index: Optional[Union[List, Tuple, str, Pattern]] = None,
     column_names: Optional[Union[List, Tuple, str, Pattern]] = None,
     names_to: Optional[Union[List, Tuple, str]] = None,
-    values_to: Optional[str] = "value",
+    values_to: Optional[Union[str, list, tuple]] = "value",
     column_level: Optional[Union[int, str]] = None,
     names_sep: Optional[Union[str, Pattern]] = None,
     names_pattern: Optional[Union[List, Tuple, str, Pattern]] = None,
```
thatlittleboy: I suggest we either just convert all of these type annotations to the built-ins (`list` instead of `List`, etc.) or maintain using `List` and `Tuple` for the type annotation of `values_to`.
To keep it consistent among all the parameter type annotations.

janitor/functions/pivot.py
```diff
@@ -313,13 +359,7 @@ def _data_checks_pivot_longer(
 
         if isinstance(names_pattern, (list, tuple)):
```
thatlittleboy: Shall we convert this into an `elif`? The `if` before this is mutually exclusive with this `if` check.

```suggestion
        elif isinstance(names_pattern, (list, tuple)):
```

janitor/functions/pivot.py
```diff
@@ -335,18 +375,39 @@ def _data_checks_pivot_longer(
...
+                    if word in names_to:
+                        raise ValueError(
+                            f"{word} in values_to already exists in names_to."
+                        )
+
+                    if word in uniques:
+                        raise ValueError(
+                            f"{word} already exists in values_to."
+                        )
+                    uniques.add(word)
```
thatlittleboy: I think `f"{word} already exists in values_to."` might not be a very helpful error message, perhaps mentioning that this word is duplicated in `values_to` would be less confusing.

janitor/functions/pivot.py
```diff
@@ -648,13 +713,14 @@ def _pivot_longer_dot_value(
...
-                "index parameter. Kindly use a unique name."
-            )
+        for word in index:
+            if word in exclude:
+                raise ValueError(
+                    f"{word}  associated with .value "
+                    "already exists as a column label "
+                    "assigned to the dataframe's index parameter. "
+                    "Kindly use a unique name."
+                )
```
thatlittleboy: Just curious why you decided to go with this change. Wouldn't it be better to show the user all of the "duplicated labels" at once, rather than raising at the very first error instance?

Because then, the user would be able to fix the errors all at once rather than one-by-one.

ericmjl (approved):
This is great work, thank you, @samukweku!

thatlittleboy (changes requested):

janitor/functions/pivot.py
ericmjl: Agreed, consistency is important.

janitor/functions/pivot.py
samukweku: My thought process was trigger error as early as possible. What's d pattern for these? All errors at once?

ericmjl: Hmmm, good points, @thatlittleboy and @samukweku. 

I think I can see why triggering as early as possible would be good, especially in the case where it takes a bit of time (~seconds to ~minutes) to collate together all of the errors. Raising early gives a bit more interactive feel.

Thinking about the user experience with `pandera`, there's perhaps something to learn there -- any rows that do not satisfy pandera checks are collected together and displayed at once.

If I were to vote, it would be:

- if the code chunk runs really fast, raise all errors all at once
- if the code chunk runs slowly, raise the first error while also letting users know, in the error message, that "we raise on first error, may be more" is the behaviour of the error handling.

ericmjl: @samukweku could I check, is this issue resolved, btw? I just realized I commented on an outdated comment thread :joy_cat:.

ericmjl:
@thatlittleboy if you think things are good here, please do the honours of merging! (Also don't worry about codecov, it's misbehaving again haha.)

janitor/functions/pivot.py
thatlittleboy: Sorry for the late reply. Yeah, there is definitely a place for both patterns (raising early / collating errors as a group and raising all at once).

I would tend to agree with Eric's general intuition (using "code runtime" as a rough gauge); For example, import errors are typically collated together (e.g. this function requires optional import `xxx`, `yyy` and `zzz`, please pip install them. [Pandas](https://github.com/pandas-dev/pandas/blob/main/pandas/__init__.py#L17) does this.). You can imagine how annoying it might be if this information gets relayed to you one package at a time.

In our case, we're essentially checking if there is an intersection between the two sets of names `index` and `exclude` (which was how the code was written before this PR). Not particularly computationally expensive I think, if that's the primary worry here.
The only instance you'll gain some time advantage is when there *is* an intersection (ala, raising the error early), but the trade-off in terms of user experience is too huge, IMO. 

Not to say there aren't situations where raising early is useful -- there are --; just not in this particular case, I think.

samukweku: thanks for the feedback; I made the changes to collate the errors into one.

thatlittleboy: Thanks, approving and merging!

thatlittleboy (approved):

PR #1060, opened 2022-04-06
[ENH] Add lazy imports to speed up the time taken to load pyjanitor.

### LINKED ISSUE #1059: Enable lazy importing for external dependencies
I'd like to propose that we use lazy imports for our external dependencies.

Currently, importing pyjanitor results in about ~1-2 seconds of delay. While insignificant for an interactive data exploration setting, it becomes a significant latency burden if there are projects that depend on `pyjanitor`. 

There is a SPEC for lazy importing, which means this idea is getting community traction in the SciPy world: https://scientific-python.org/specs/spec-0001/

There is also a tool for making lazy loading easy: https://pypi.org/project/lazy_loader/, thereby giving us a path towards lazy loading.

Would love to get thoughts from the dev team!

ericmjl:
# PR Description

- We switch some imports to be lazy imports.
- We also add a document that explains why we use lazy importing.

**This PR resolves #1059.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1060--pyjanitor.netlify.app

samukweku (approved):

mkdocs/development/lazy_imports.md
```diff
@@ -0,0 +1,52 @@
...
+
+You'll be redirected to your browser,
+where the web UI will allow you to see
+which imports are causing time delays.
+
+![Tuna's Web UI](./images/tuna.png)
+
+## Which imports to lazily load
+
+Generally speaking, the the _external_ imports are the ones that
```
thatlittleboy:
```suggestion
Generally speaking, the _external_ imports are the ones that
```
minor typo

mkdocs/development/lazy_imports.md
```diff
@@ -0,0 +1,52 @@
...
+python -X importtime -c "import janitor" 2> timing.log
+```
+
+Then, using the `tuna` CLI tool, we can view the timing log:
+
+```bash
+tun timing.log
+```
+
+Note: You may need to install tuna using `pip install -U tuna`.
```
thatlittleboy: 1. I believe the invocation is meant to be `tuna` rather than `tun`?
2. I think it would be suitable to also include a link to the tuna github / pypi in either of these sentences.

mkdocs/development/lazy_imports.md
```diff
@@ -0,0 +1,52 @@
...
+
+You'll be redirected to your browser,
+where the web UI will allow you to see
+which imports are causing time delays.
+
+![Tuna's Web UI](./images/tuna.png)
```
thatlittleboy, line 48: This is showing times after lazy loading (it seems the total loading time is 0.5s in the image)? Is it more suitable to show the timings *before* lazy loading instead? 😅 I'm curious to see what the graph looks like if we don't do lazy import, and can help motivate why we choose to lazy import some modules but not the others.

janitor/functions/add_columns.py
```diff
@@ -1,9 +1,11 @@
...
 
+import lazy_loader as lazy
 from janitor.utils import check, deprecated_alias
-import pandas as pd
-from typing import Union, List, Any, Tuple
-import numpy as np
+
+pf = lazy.load("pandas_flavor")
+pd = lazy.load("pandas")
+np = lazy.load("numpy")
```
thatlittleboy: Not sure if I am misunderstanding how lazy loading with `lazy_loader` should work, but does this help in any significant way to reduce import times? I would've thought that at least for `pf` and `pd`, they would be immediately loaded anyway when this file is loaded because they are used directly within this file. 😅

mkdocs.yml
```diff
@@ -19,7 +19,9 @@ theme:
 # in which pages show up.
 nav:
   - Home: index.md
-  - Development Guide: devguide.md
+  - Development:
+      - Dev Guide: development/devguide.md
```
thatlittleboy, line 23: This should impact the URL for the devguide, right? In which case we will need to update the url mentioned in the `mkdocs/index.md` file as well.

thatlittleboy (changes requested):
Really cool idea, TIL @ericmjl !

Requesting changes for now as I'm still mulling over some of the changes suggested here. Leaving my other discussion comments in the issue thread (#1059) so it's more visible. :)

mkdocs/development/lazy_imports.md
ericmjl: Done!

mkdocs/development/lazy_imports.md
ericmjl: Good point! The timings are drastically different. I will add a before/after comparison.

Edit: I remember taking import times down from 2 sec to 0.5, but I can't replicate the old behaviour. Back at 2 seconds, the import time was reliably long, though now, on `dev`, the import times that I'm measuring are around 0.8 seconds. On this branch, we're reliably at 0.5 seconds.

janitor/functions/add_columns.py
ericmjl: I've mostly gone empirical with this profiling activity, so I don't think I have a first-principles way of explaining how lazy-loader should work 😂. On the other hand, I have also done another empirical test where I lazy-import pandas-flavor early on in `__init__.py` -- that has the same import speed effect, I think.

ericmjl:
Apologies to everyone who helped review this PR. I would like to close this PR without merging. My reasons are below.

Firstly, the PR was borne out of a need at work, and that need has gone away through a redesign around a slow import.

Secondly, lazy imports are gaining ground in the broader Python community; I am hopeful that lazy imports will become more and more commonplace and therefore built-in to the language. Given this momentum, I think we should ride the wave rather than be ahead of it.

Thirdly, I've neglected the PR and we've hence had a ton of conflicts accumulate. It might take a bit of time to resolve them.

If we're interested in implementing lazy importing in the library for other good reasons, we can revisit the topic.

PR #1064, opened 2022-04-15
[ENH/DOC] Improve transform_columns

### LINKED ISSUE #1063: transform_columns ignores `column_names` when `new_column_names` is specified
# Brief Description

`transform_columns` ignores the cols specified in `column_names` if `new_column_names` is specified (wrongly). For example, See the MWE.

# Minimally Reproducible Code

```python
>>> import numpy as np
>>> import pandas as pd
>>> import janitor
>>> df = pd.DataFrame({
...     "col1": [5, 10, 15],
...     "col2": [3, 6, 9],
...     "col3": [10, 100, 1_000],
... })
>>> df
    col1  col2  col3
0     5     3    10
1    10     6   100
2    15     9  1000

>>> df.transform_columns(
...     ["col1", "col3"],
...     np.log10,
...     new_column_names={"col1": "transform1", "col2": "transform2"},
... )
   col1  col2  col3  transform1  transform2
0     5     3    10    0.698970    0.477121
1    10     6   100    1.000000    0.778151
2    15     9  1000    1.176091    0.954243
```
Note that I specified that I wanted to transform "col1" and "col3" with log10, but "wrongly" specified the keys in the dictionary of `new_column_names` as "col1" and "col2".
Yet, the transformation is applied on "col1" and "col2" (obvious, because the expected output of `log10(col3)` is `[1, 2, 3]` which is not present in the output).

thatlittleboy:
# PR Description

- Bug fix for `transform_columns` by setting up the `dest_column_names` dictionary using the columns in `column_names` instead of just using the user-entered `new_column_names`. Also added a test to avoid a similar regression in the future. Resolves #1063 .
- Make `transform_column(s)` not mutate the original dataframe.
- Check the column to be transformed actually exists, if not raise the error as early as possible.
- Add MWE for `transform_column(s)`. Related to #972 .

### DIFF
--- a/janitor/functions/transform_columns.py
+++ b/janitor/functions/transform_columns.py
@@ -172,7 +172,9 @@ def transform_columns(
         to apply the transformation function elementwise (True)
         or columnwise (False).
     :param new_column_names: An explicit mapping of old column names in
-        `column_names` to new column names.
+        `column_names` to new column names. If any column specified in
+        `column_names` is not a key in this dictionary, the transformation
+        will happen in-place for that column.
     :returns: A pandas DataFrame with transformed columns.
     :raises ValueError: If both `suffix` and `new_column_names` are
         specified.
@@ -184,22 +186,21 @@ def transform_columns(
             "Only one of `suffix` or `new_column_names` should be specified."
         )
 
-    dest_column_names = dict(zip(column_names, column_names))
-
     if suffix:
         check("suffix", suffix, [str])
-        for col in column_names:
-            dest_column_names[col] = col + suffix
+        dest_column_names = {col: col + suffix for col in column_names}
     elif new_column_names:
         check("new_column_names", new_column_names, [dict])
         dest_column_names = new_column_names
+    else:
+        dest_column_names = {}
 
-    for old_col, new_col in dest_column_names.items():
+    for old_col in column_names:
         df = transform_column(
             df,
             old_col,
             function,
-            new_col,
+            dest_column_name=dest_column_names.get(old_col, old_col),
             elementwise=elementwise,
         )
 
--- a/janitor/functions/transform_columns.py
+++ b/janitor/functions/transform_columns.py
@@ -12,10 +12,8 @@ def _get_transform_column_result(
 ) -> pd.Series:
     """Perform the actual computation for Series transformation."""
     if elementwise:
-        result = series.apply(function)
-    else:
-        result = function(series)
-    return result
+        return series.apply(function)
+    return function(series)
 
 
 @pf.register_dataframe_method
@@ -116,8 +114,7 @@ def transform_column(
         elementwise,
     )
 
-    df = df.assign(**{dest_column_name: result})
-    return df
+    return df.assign(**{dest_column_name: result})
 
 
 @pf.register_dataframe_method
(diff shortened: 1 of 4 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1064--pyjanitor.netlify.app

samukweku (approved):

samukweku:
@thatlittleboy can we also do a column check for the new columns, to ensure they are not already in the dataframe. thanks

thatlittleboy:
> @thatlittleboy can we also do a column check for the new columns, to ensure they are not already in the dataframe. thanks

Ok done, I've taken the liberty to skip the checks if `dest_column_name == column_name` though. I thought we could also give the end-user the option to specify an explicit in-place transformation (it'll be the same effect as having `dest_column_name=None`)
@samukweku

samukweku (approved):

janitor/functions/transform_columns.py
```diff
@@ -121,95 +118,97 @@ def transform_columns(
...
+    elif new_column_names:
         check("new_column_names", new_column_names, [dict])
         dest_column_names = new_column_names
+    else:
+        dest_column_names = {}
 
-    # Now, transform columns.
-    for old_col, new_col in dest_column_names.items():
+    for old_col in column_names:
         df = transform_column(
```
samukweku: multiple copies here, since it is calling transform_column (each assign will create a copy). Is it possible to do it once? maybe build a dictionary of column names and values, then assign once to df? ... this might require creating a sub function within transform_column, that precludes the assign step.

thatlittleboy: True, I did a bit of refactoring. Kindly re-review the `transform_columns.py` file again, thanks!

janitor/functions/transform_columns.py
```diff
@@ -5,6 +5,19 @@
 from janitor.utils import check, check_column, deprecated_alias
 
 
+def _get_transform_column_result(
+    series: pd.Series,
+    function: Callable,
+    elementwise: bool,
+) -> pd.Series:
+    """Perform the actual computation for Series transformation."""
+    if elementwise:
```
samukweku, line 14: maybe this :  
```py
    if elementwise:
         return series.apply(function)
    return function(series)
```

janitor/functions/transform_columns.py
```diff
@@ -97,10 +110,11 @@ def transform_column(
...
-        result = df[column_name].apply(function)
-    else:
-        result = function(df[column_name])
+    result = _get_transform_column_result(
+        df[column_name],
+        function,
+        elementwise,
+    )
 
     df = df.assign(**{dest_column_name: result})
```
samukweku: we can just return here, without the assignment:  ``return df.assign(**{dest_column_name: result})``

janitor/functions/transform_columns.py
```diff
@@ -198,17 +213,21 @@ def transform_columns(
...
+        if old_col != new_col:
+            check_column(df, new_col, present=False)
+        results[new_col] = _get_transform_column_result(
+            df[old_col],
             function,
-            dest_column_name=dest_column_names.get(old_col, old_col),
             elementwise=elementwise,
         )
 
+    df = df.assign(**results)
```
samukweku: same here, no need for the reassignment : ``return df.assign(**results)``

ericmjl (approved):

samukweku:
LGTM ... If there are no other updates then I'll merge

PR #1066, opened 2022-04-15
[ENH/DOC] Fix case_sensitive bug in `count_cumulative_unique` and add MWE

### LINKED ISSUE #1065: Incorrectly modified output of `count_cumulative_unique` when `case_sensitive` is turned off
# Brief Description

The `case_sensitive` parameter of `count_cumulative_unique` when set to False, would modify the column which is being counted, and this change is then (wrongly) propagated to the output.

The implementation is currently written like so:
```python
    if not case_sensitive:
        # Make it so that the the same uppercase and lowercase
        # letter are treated as one unique value
        df[column_name] = df[column_name].astype(str).map(str.lower)
```
Which modifies the column_name for the intermediate calculation of `count_cumulative_unique`, which is incorrect. See MWE.

# Minimally Reproducible Code

```python
In [4]: df = pd.DataFrame({"ccu": np.array(["a", "b", "c", "A", "B", "C", "a", "b", "c"])})

In [5]: df
Out[5]: 
  ccu
0   a
1   b
2   c
3   A
4   B
5   C
6   a
7   b
8   c

In [6]: df.count_cumulative_unique("ccu", "counting", case_sensitive=False)
Out[6]: 
  ccu  counting
0   a         1
1   b         2
2   c         3
3   a         3
4   b         3
5   c         3
6   a         3
7   b         3
8   c         3
```
Note that the `ccu` column in the original df has upper case letters, but the `ccu` column in the output does not. Changes made to the dataframe for intermediate calculations (i.e., `str.lower()`) should not be propagated to the final output.

thatlittleboy:
# PR Description

- Fixed bug: `count_cumulative_unique` now no longer modifies the column being counted in the output when `case_sensitive` argument is set to False, resolves #1065. See the bug description in the linked issue.
- And add tests to avoid similar regression in the future.
- Add MWE, related to #972 .

### DIFF
--- a/janitor/functions/count_cumulative_unique.py
+++ b/janitor/functions/count_cumulative_unique.py
@@ -77,6 +77,7 @@ def count_cumulative_unique(
         count of unique values from another column.
     """
     check_column(df, column_name)
+    check_column(df, dest_column_name, present=False)
 
     if not case_sensitive:
         # Make it so that the the same uppercase and lowercase
--- a/janitor/functions/count_cumulative_unique.py
+++ b/janitor/functions/count_cumulative_unique.py
@@ -1,6 +1,9 @@
+"""Implementation of count_cumulative_unique."""
 from typing import Hashable
 import pandas_flavor as pf
 import pandas as pd
+
+from janitor.utils import check_column
 
 
 @pf.register_dataframe_method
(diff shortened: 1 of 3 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1066--pyjanitor.netlify.app

samukweku (commented):
Maybe add another example for case_sensitive?

janitor/functions/count_cumulative_unique.py
```diff
@@ -12,73 +15,85 @@ def count_cumulative_unique(
...
     :param dest_column_name: The name of the new column containing the
         cumulative count of unique values that will be created.
     :param case_sensitive: Whether or not uppercase and lowercase letters
-        will be considered equal (e.g., 'A' != 'a' if `True`).
-
+        will be considered equal.
     :returns: A pandas DataFrame with a new column containing a cumulative
         count of unique values from another column.
     """
+    check_column(df, column_name)
```
samukweku, line 85: can we add a check here for the dest_column_name, something like `check_column(df, dest_column_name, present=True)` to raise an error if `dest_column_name` already exists.

samukweku (approved):

samukweku:
@thatlittleboy apologies, just playing it in my head and realise reindex might fail if there is a duplicate index

thatlittleboy:
> @thatlittleboy apologies, just playing it in my head and realise reindex might fail if there is a duplicate index

@samukweku What do you propose here? Should we check if df has duplicated index and throw an error earlier on in the function? Or you have something else in mind?

samukweku:
@thatlittleboy I'll avoid it altogether and use a combination of cumcount and cumsum:

```py
counter = df[column_name] # copy maybe?
# convert to numpy to avoid assignment for duplicate index
# might not be necessary though ... just a guard ... plus speed maybe?
counter = counter.groupby(counter, sort = False).cumcount().to_numpy(copy = False)
counter = np.cumsum(counter == 0)
df.assign(**{dest_column_name : counter})
```

I also think we should not convert the column to a string; if it isn't a string, it should error out, or we can do a check to ensure it is a string (`pd.is_string_dtype` maybe). converting an integer to a string then lower feels unnecessary, so maybe assuming it is a string and erroring out might be better?

thatlittleboy:
@samukweku Raising a TypeError now, but I'm opting for a try-except instead of checking with `is_string_dtype`. As that doesn't cover categoricals (for example), and `bytes` type will actually throw a TypeError even though it is string, because of [this](https://github.com/pandas-dev/pandas/blob/7dd22b8de0c006f99963e39426abce2916290f90/pandas/core/strings/accessor.py#L2952).

Refer to the test cases in the new test function `test_not_case_sensitive_but_nonstring()` to see my considerations.

ericmjl (approved):
@thatlittleboy thank you for the bugfix and MWE! My many apologies for not reviewing earlier. I think we're in a good spot here, and will merge asap.

PR #1097, opened 2022-05-04
[EHN] `min_max_scale` drop `old_min` and `old_max` to fit sklearn's method API

### LINKED ISSUE #1068: Discussion: arguments `old_min` and `old_max` should be removed from `min_max_scale`
```python
>>> import pandas as pd
>>> import janitor

# Use one column dataframe to avoid scaling the entire data or the column data problem

>>> df = pd.Series([0, 1, 2]).to_frame()

# use the minimum and maximum value of data

>>> df.min_max_scale()
     0
0  0.0
1  0.5
2  1.0

# Overwrite the value of data. The result seems wired for the user, but it's ok for the formula view.
# Question 1: 0 should be scaled or not? 0 is out range of [old_min 1, old_max 2].
# Question 2: I already define the new_min (0) and new_max of value. Why do there have -1?
# Question 3: The API differs from sklearn.preprocessing.MinMaxScaler

# Min-Max normalization formula
# X_std = (X - X.min(axis=0)) / (X.max(axis=0) - X.min(axis=0))
# X_scaled = X_std * (max - min) + min

>>> df.min_max_scale(old_min=1, old_max=2)
     0
0 -1.0
1  0.0
2  1.0
```

In the end, it's hard to trace the `min_max_scale` [committing history](https://github.com/pyjanitor-devs/pyjanitor/commits/dev/janitor/functions/min_max_scale.py) to know the reason why added these options.

Zeroto521:
# PR Description

- Drop `old_min` and `old_max` parameters.
- Merge `new_min` and `new_max` two parameters to one `feature_range`.

**This PR resolves #1068.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


janitor/functions/min_max_scale.py
```diff
@@ -8,11 +10,12 @@
 @deprecated_alias(col_name="column_name")
 def min_max_scale(
     df: pd.DataFrame,
-    old_min=None,
```
Zeroto521, line 11: I thought there should be a _typing.py script that could contain some common typing.

janitor/functions/min_max_scale.py
```diff
@@ -8,11 +10,12 @@
...
-    old_max=None,
-    column_name=None,
-    new_min=0,
-    new_max=1,
+    feature_range: tuple[int | float, int | float] = (0, 1),
+    column_name: str
+    | int
+    | list[str | int]
+    | tuple[str | int]
+    | pd.Index = None,
```
Zeroto521: I thought there should be a _typing.py script that could contain some common typing such as `column_name`.

janitor/functions/min_max_scale.py
```diff
@@ -72,45 +68,42 @@ def min_max_scale(
...
+    ):
+        raise ValueError(
+            "`feature_range` should be a range type contains number element, "
+            "the first element must be greater than the second one"
+        )
 
+    new_min, new_max = feature_range
     new_range = new_max - new_min
 
     if column_name:
```
Zeroto521, line 94: `df[column_name]`, `column_name` could also be `pandas.Index`. But `if column_name` would raise an error. `pandas object can't be easily handled like that.

```python
ValueError: The truth value of a Index is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

Zeroto521:
Since this PR would drop `old_min`, `old_max`, `new_min`, and `new_max` four parameters.
Is there a warning message to notify users the API has changed?

ericmjl:
> Since this PR would drop `old_min`, `old_max`, `new_min`, and `new_max` four parameters.
> Is there a warning message to notify users the API has changed?

Good point, @Zeroto521!

If my memory serves me right, there is a function that's available for deprecating arguments/keyword arguments. Could you see if that would do the trick?

If not, something we can do here might be to capture the old kwargs in `**kwargs` and raise an error if any of the old kwargs are used. In the error, I can foresee that explaining the rationale is going to be quite helpful as well (something more diplomatic than "we would like to just follow sklearn", but conveying the same meaning).

ericmjl (approved):
I'm pre-approving the PR. It's small enough that I can follow along easily, and we've had extensive discussions. One thing that should be done before we merge is raising an exception (or warning) for users, since this is a breaking API change.

Zeroto521:
According to this condition, we build a new decorator called `deprecated_kwargs` like `deprecated_alias` in a new PR or directly raise the error in the source code?

ericmjl:
For now, I think directly raising the error makes sense, unless you think you've got a good generalizable pattern. In the latter case, we can make a `deprecated_kwargs` decorator directly. I'll leave the decision in your hands, both ways sound reasonable to me!

PR #1101, opened 2022-05-10
[ENH] faster groupby top k

### LINKED ISSUE #1093: groupby_topk performance
it is possible to use `nlargest/nsmallest` instead of sort_values for `groupby_topk`, this should offer more performance by avoiding a full sort (which is what nlargest/nsmallest achieves)

samukweku:
# PR Description

- Use nlargest/nsmallest instead ... avoids sorting the entire dataframe
- change column names to align more with pandas description and other existing functions (open to suggestions, or possible revert to existing names)
- deprecate sort_kwargs, since there is no use of `sort_values`
-  add `ascending` parameter to match existing parameter in the deprecated `sort_kwargs`

**This PR resolves #1093 .**

# performance comparison ... view with a pinch of salt, as always

```py
# adapted from  https://stackoverflow.com/a/36702926/7175713
np.random.seed(3)
df = pd.DataFrame({
   'dep': np.random.choice(list('ABC'), 2_000_000),
  'manager_id': np.random.randint(0, 200_000, 2_000_000),
   'salary': np.random.randint(5000, 5006, 2_000_000)
 })

df.head()
  dep  manager_id  salary
0   C       12586    5002
1   A      153719    5004
2   B      178720    5004
3   A      162938    5001
4   A      122360    5000

# --------------------> PR

df.groupby_topk('dep', 'salary', 5)
   dep  manager_id  salary
0    C       44168    5000
1    C      135457    5000
2    C       42892    5000
3    C      130785    5000
4    C      155488    5000
5    A      122360    5000
6    A      136722    5000
7    A       14219    5000
8    A      162417    5000
9    A      115906    5000
10   B       66927    5000
11   B       79314    5000
12   B      150702    5000
13   B      186580    5000
14   B       34819    5000

df.groupby_topk('dep', 'salary', 5, ignore_index = False) 
    dep  manager_id  salary
11    C       44168    5000
13    C      135457    5000
42    C       42892    5000
73    C      130785    5000
118   C      155488    5000
4     A      122360    5000
14    A      136722    5000
30    A       14219    5000
46    A      162417    5000
49    A      115906    5000
29    B       66927    5000
33    B       79314    5000
38    B      150702    5000
51    B      186580    5000
76    B       34819    5000

%timeit df.groupby_topk('dep', 'salary', 5)
162 ms ± 12.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit df.groupby_topk('dep', 'salary', 5, ignore_index = False)
154 ms ± 1.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)



# ---------------------------> dev

df.groupby_topk('dep', 'salary', 5)
            dep  manager_id  salary
dep                                
A   1999996   A      131800    5000
    548719    A       57184    5000
    548720    A       59279    5000
    1585713   A       87340    5000
    548727    A      196661    5000
B   716069    B       89977    5000
    306266    B       34652    5000
    743267    B      125335    5000
    1877756   B      166187    5000
    1703182   B      101396    5000
C   1999997   C      180787    5000
    292147    C      105282    5000
    1609388   C       52935    5000
    1609401   C       55700    5000
    292136    C      111160    5000

df.groupby_topk('dep', 'salary', 5).droplevel(0) 
        dep  manager_id  salary
1999996   A      131800    5000
548719    A       57184    5000
548720    A       59279    5000
1585713   A       87340    5000
548727    A      196661    5000
716069    B       89977    5000
306266    B       34652    5000
743267    B      125335    5000
1877756   B      166187    5000
1703182   B      101396    5000
1999997   C      180787    5000
292147    C      105282    5000
1609388   C       52935    5000
1609401   C       55700    5000
292136    C      111160    5000

%timeit df.groupby_topk('dep', 'salary', 5)
297 ms ± 4.18 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
 ```

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### CODE REVIEW

**Comment by ericmjl** [comment]
🚀 Deployed on https://deploy-preview-1101--pyjanitor.netlify.app

File: janitor/functions/groupby_topk.py [hunk 1]
```diff
@@ -41,59 +44,78 @@ def groupby_topk(
...
-                  age  id result
-        result
-        fail   2   22   6   fail
-               4   21   5   fail
-        pass   3   43   2   pass
-               1   23   4   pass
+
+    Ascending top 3:
+
+        >>> df.groupby_topk(by = 'result', column = 'age',  k = 3)
```
thatlittleboy: A small request to use `black` formatting even within code examples to keep a consistent look in the documentation. That is to say, no spaces around `=`, double quotes, and trailing commas where necessary.

janitor/functions/groupby_topk.py
```diff
@@ -41,59 +44,78 @@ def groupby_topk(
...
+
+    Descending top 2 :
+
+        >>> df.groupby_topk(
+        ...     by = 'result',
+        ...     column = 'age',
+        ...     k = 2,
+        ...     ascending = False,
+        ...     ignore_index = False
+        ...     )
```
thatlittleboy: Same here, + trailing comma on the final kwarg and the indentation of the final parenthesis is wrong.

janitor/functions/groupby_topk.py
```diff
@@ -41,59 +44,78 @@ def groupby_topk(
...
+    :returns: A pandas DataFrame with top `k` rows per `column`, grouped by `by`.
     :raises ValueError: if `k` is less than 1.
-    :raises ValueError: if `groupby_column_name` not in DataFrame `df`.
-    :raises ValueError: if `sort_column_name` not in DataFrame `df`.
-    :raises KeyError: if `inplace:True` is present in `sort_values_kwargs`.
     """  # noqa: E501
 
-    # Convert the default sort_values_kwargs from None to empty Dict
-    sort_values_kwargs = sort_values_kwargs or {}
+    # TODO: allow multiple columns for `columns`
```
thatlittleboy: Is this TODO comment still relevant?

janitor/functions/groupby_topk.py
```diff
@@ -41,59 +44,78 @@ def groupby_topk(
...
-        and sort_values_kwargs["inplace"]
-    ):
-        raise KeyError("Cannot use `inplace=True` in `sort_values_kwargs`.")
+    indices = df.groupby(by=by, dropna=dropna, sort=False, observed=True)
+    indices = indices[column]
+
+    if ascending:
+        indices = indices.nsmallest(n=k)
+    else:
+        indices = indices.nlargest(n=k)
```
thatlittleboy: This isn't, strictly speaking, equivalent to the initial implementation involving sorting and taking the head/tail though.

`nsmallest`/`nlargest` don't work with string / object types, for example.
So at the moment, if `column` passed in is of a non-numeric dtype, `pandas` will raise an error. (It's a limitation imposed on by pandas, but I'm not quite sure why.)

thatlittleboy (changes requested):

janitor/functions/groupby_topk.py
samukweku: fair point ... based on this, it is fair to abandon this PR

ericmjl: Before we close the PR, it may be worth thinking about the range of use cases. 

`groupby_topk` probably implies a numeric ordering, rather than lexicographic ordering, right? 

If so, then I think the implementation here, while more constrained in scope, fulfills the intent of the function more closely, and we can close out the PR by merging.

If not, then let's close the PR without merging.

samukweku: One option could be to put a check ... If it is a numeric column use nlargest/smallest, else sort? Another option would be to use `np.partition`, though I am not sure it is worth the trouble

thatlittleboy: re. the scope of the `groupby_topk` function, there is kind of a gray area given the way the function is named. I'll be okay with going with either interpretation (only allowing numeric ordering, OR allowing generic ordering which includes lexicographical or even custom ordering).

I'm also not in favour of closing the PR, I think there is value with what Sam is proposing here.

I'm only bringing this up in the review because, given the previous implementation allowed for generic ordering, there might be some users making use of this behaviour. If we decide to go with purely numeric ordering, we need to give ample warning to the users (in the CHANGELOG & in the docstring).

That said, I would vote for Sam's first proposal (if numeric, then use nlargest/smallest; while having `groupby_topk` still supporting generic ordering for other types, via the old implementation).

ericmjl: @samukweku @thatlittleboy great points. Yeah, I agree with @thatlittleboy also, that we include an if/else block to call on the appropriate chunk of code based on the `dtype` of the column being sorted. @samukweku, do you know if it would be too much work to reinstate the original code for the non-numerical sorting use case?

samukweku: I could do that. One thing with the original code though is that the grouping column becomes the index as well; I propose that the dataframe index remains as is and the grouper stays within the dataframe

ericmjl: Yes, that sounds like a reasonable change to make, I think. Let's go with that.

samukweku:
made the update with the old code; not sure if I'm missing something else. by the way @thatlittleboy how do you format the docs with black? or do you copy the code snippet, run it in a .py and copy back?

thatlittleboy:
> made the update with the old code; not sure if I'm missing something else. by the way @thatlittleboy how do you format the docs with black? or do you copy the code snippet, run it in a .py and copy back?

I don't typically run `black` if the snippets are small enough to "fix by hand" (in general, I know what the "black" style should look like, so I just write in that manner when coding up the MWEs).
There might be tools like `blacken-docs` that we can consider integrating, though that one doesn't seem to support the way of declaring code snippets that we are using. (EDIT: ignore me.. that only runs on md/rst files, not within docstrings)

thatlittleboy (approved):
Nice! thanks!

ericmjl (approved):

ericmjl:
As all of the tests were green before the conflict resolution, I'm going to merge. Thank you, @samukweku!

PR #1112, opened 2022-06-02
[EHN] Add `jointly` option for `min_max_scale`

### LINKED ISSUE #1067: [EHN] `min_max_scale` transforms each column values not entire values
# Brief Description

I thought this method should transform each column values sperately not all values when `column_name` was set to `None`.
From `DataFrame` view, each column represtent differnt data.

```python
def min_max_scale(
    df: pd.DataFrame,
    old_min=None,
    old_max=None,
    column_name=None,
    new_min=0,
    new_max=1,
) -> pd.DataFrame:
    ...
```

# Example

```python
import pandas as pd
import janitor  #noqa

df = pd.DataFrame({'a':[1, 2], 'b':[0, 1]})
df.min_max_scale(new_max=10)
# not this one
#       a    b
# 0   5.0  0.0
# 1  10.0  5.0

# excepted result
#       a     b
# 0   0.0   0.0
# 1  10.0  10.0

# To get excepted result I have to add `column_name=df.columns.tolist()`
df.min_max_scale(new_max=10, column_name=df.columns.tolist())
#       a     b
# 0   0.0   0.0
# 1  10.0  10.0

# bad for chaining
(
    df.eval("c = a+b")
    .min_max_scale(new_max=10, column_name=df.columns.tolist())  # 'c' column was missed.
)
#       a     b  c
# 0   0.0   0.0  1
# 1  10.0  10.0  3
```

Zeroto521:
# PR Description

- Add an option for `min_max_scale` support to transform each column values or entire values
- Default transform each column, similar behavior to [sklearn.preprocessing.MinMaxScaler](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.MinMaxScaler.html)

**This PR resolves #1067.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions/min_max_scale.py
+++ b/janitor/functions/min_max_scale.py
@@ -142,6 +142,9 @@ def min_max_value(df: pd.DataFrame, entire_data: bool) -> tuple:
     Return the minimum and maximum of DataFrame.
 
     Use the `entire_data` flag to control returning entire data or each column.
+
+    .. # noqa: DAR101
+    .. # noqa: DAR201
     """
 
     if entire_data:
@@ -169,6 +172,9 @@ def apply_min_max(
     - Inputting minimum and maximum type
         - int or float : It will apply minimax to the entire DataFrame.
         - Series : It will apply minimax to each column.
+
+    .. # noqa: DAR101
+    .. # noqa: DAR201
     """
 
     old_range = old_max - old_min
--- a/janitor/functions/min_max_scale.py
+++ b/janitor/functions/min_max_scale.py
@@ -23,11 +23,10 @@ def min_max_scale(
     df: pd.DataFrame,
     feature_range: tuple[int | float, int | float] = (0, 1),
     column_name: str | int | list[str | int] | pd.Index = None,
+    entire_data: bool = False,
 ) -> pd.DataFrame:
     """
     Scales data to between a minimum and maximum value.
-
-    This method mutates the original DataFrame.
 
     If `minimum` and `maximum` are provided, the true min/max of the
     `DataFrame` or column is ignored in the scaling process and replaced with
@@ -102,23 +101,61 @@ def min_max_scale(
             "the first element must be greater than the second one"
         )
 
-    new_min, new_max = feature_range
-    new_range = new_max - new_min
-
     if column_name is not None:
-        old_min = df[column_name].min()
-        old_max = df[column_name].max()
-        old_range = old_max - old_min
+        df = df.copy()  # Avoid to change the original DataFrame.
 
-        df = df.copy()
-        df[column_name] = (
-            df[column_name] - old_min
-        ) * new_range / old_range + new_min
+        old_feature_range = df[column_name].pipe(min_max_value, entire_data)
+        df[column_name] = df[column_name].pipe(
+            apply_min_max,
+            *old_feature_range,
+            *feature_range,
+        )
     else:
-        old_min = df.min().min()
-        old_max = df.max().max()
-        old_range = old_max - old_min
-
-        df = (df - old_min) * new_range / old_range + new_min
+        old_feature_range = df.pipe(min_max_value, entire_data)
+        df = df.pipe(
+            apply_min_max,
+            *old_feature_range,
+            *feature_range,
+        )
 
     return df
+
+
+def min_max_value(df: pd.DataFrame, entire_data: bool) -> tuple:
+    """
+    Return the minimum and maximum of DataFrame.
+
+    Use the `entire_data` flag to control returning entire data or each column.
+    """
+
+    if entire_data:
+        mmin = df.min().min()
+        mmax = df.max().max()
+    else:
+        mmin = df.min()
+        mmax = df.max()
+
+    return mmin, mmax
+
+
+def apply_min_max(
+    df: pd.DataFrame,
+    old_min: int | float | pd.Series,
+    old_max: int | float | pd.Series,
+    new_min: int | float | pd.Series,
+    new_max: int | float | pd.Series,
+) -> pd.DataFrame:
+    """
+    Apply minimax scaler to DataFrame.
+
+    Notes
+    -----
+    - Inputting minimum and maximum type
+        - int or float : It will apply minimax to the entire DataFrame.
+        - Series : It will apply minimax to each column.
+    """
+
+    old_range = old_max - old_min
+    new_range = new_max - new_min
+
+    return (df - old_min) * new_range / old_range + new_min
--- a/janitor/functions/min_max_scale.py
+++ b/janitor/functions/min_max_scale.py
@@ -59,7 +59,7 @@ def min_max_scale(
                a      b
         0    0.0    0.0
         1  100.0  100.0
-        >>> df.min_max_scale(feature_range=(0, 100), enture_data=True)
+        >>> df.min_max_scale(feature_range=(0, 100), entire_data=True)
                a     b
         0   50.0   0.0
         1  100.0  50.0
--- a/janitor/functions/min_max_scale.py
+++ b/janitor/functions/min_max_scale.py
@@ -43,6 +43,10 @@ def min_max_scale(
         >>> df = pd.DataFrame({'a':[1, 2], 'b':[0, 1]})
         >>> df.min_max_scale()
              a    b
+        0  0.0  0.0
+        1  1.0  1.0
+        >>> df.min_max_scale(entire_data=True)
+             a    b
         0  0.5  0.0
         1  1.0  0.5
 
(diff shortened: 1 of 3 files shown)


janitor/functions/min_max_scale.py
```diff
@@ -23,23 +23,18 @@ def min_max_scale(
     df: pd.DataFrame,
     feature_range: tuple[int | float, int | float] = (0, 1),
     column_name: str | int | list[str | int] | pd.Index = None,
+    entire_data: bool = False,
```
Zeroto521: I don't think `entire_data` is a better name.
Need to be added to the changelog file.

janitor/functions/min_max_scale.py
```diff
@@ -84,6 +98,7 @@ def min_max_scale(
     :param df: A pandas DataFrame.
     :param feature_range: (optional) Desired range of transformed data.
     :param column_name: (optional) The column on which to perform scaling.
+    :param entire_data: (bool) Scale the entire data if Ture.
```
Zeroto521: Does mkdoc have any markup like `versionadded`, `deprecated` like [sphinx](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#paragraph-level-markup=)?

janitor/functions/min_max_scale.py
```diff
@@ -102,23 +117,67 @@ def min_max_scale(
...
+
+
+def min_max_value(df: pd.DataFrame, entire_data: bool) -> tuple:
+    """
+    Return the minimum and maximum of DataFrame.
+
+    Use the `entire_data` flag to control returning entire data or each column.
+
+    .. # noqa: DAR101
+    .. # noqa: DAR201
```
Zeroto521, line 152: The inner function doesn't have more descriptions of parameters and a return section.

Zeroto521:
Well, GitHub Action style checking is failing but pre-commit.ci app is okay.
I thought it was time to cut the duplicate style checkings #1113

image

ericmjl:
@Zeroto521 looks like we're good. The action didn't fail b/c of your code; it failed b/c of a system error.

janitor/functions/min_max_scale.py
ericmjl: I agree, though I also need a bit of time and space to think of a better name. I wonder if others on the dev team have ideas? @pyjanitor-devs/core-devs

samukweku: not sure of a good name to use. maybe `keep` like in Pandas?

Zeroto521: You mean 'keep' its value could be 'column' or 'all'?

But this variable is better as a boolean type.

janitor/functions/min_max_scale.py
ericmjl: I don't think `mkdocs` has this functionality. We could simply add that in as part of the text, maybe?

janitor/functions/min_max_scale.py
samukweku: yea, horrible parameter name. maybe `scale_all` or `scale_all_columns`?

janitor/functions/min_max_scale.py
Zeroto521: Maybe we could add the changing list at the end like the following?

```python
"""
    ...

    :param feature_range: (optional) Desired range of transformed data.
    :param column_name: (optional) The column on which to perform scaling.
    :param entire_data: (bool) Scale the entire data if Ture.
    :returns: A pandas DataFrame with scaled data.
    :raises ValueError: if `feature_range` isn't tuple type.
    :raises ValueError: if the length of `feature_range` isn't equal to two.
    :raises ValueError: if the element of `feature_range` isn't number type.
    :raises ValueError: if `feature_range[1]` <= `feature_range[0]`.

    Changed in version 0.24.0: Deleted "old_min", "old_max", "new_min", and "new_max" options.
    Changed in version 0.24.0: Added "feature_range", and "entire_data" options.
"""
```
image

janitor/functions/min_max_scale.py
ericmjl: Hmmm, @Zeroto521, on second thought, I think we need a bit better definition of the semantics. `min_max_scale` currently operates with the assumption of operating on one column, so the use of `column_name` makes sense here. Below is my attempt at reasoning through the multiple ways we could use `min_max_scale`.

- Scale one column.
- Scale multiple columns independently.
- Scale multiple columns, but jointly (so they are all scaled to the same min and max)

`entire_data` is a special case of scaling multiple columns _jointly_. 

Since we're working on this function, it's a good chance to change the API to be flexible yet also sensible. What if the API was, instead the following?

```python
def min_max_scale(df, feature_range, column_names: Iterable[Hashable] | Callable, jointly: bool):
```

If `jointly` is True, then the `column_names` provided are jointly scaled; otherwise, they are not. 

I wanted to point out a new behaviour that we might be able to support across the rest of the API -- by making `column_names` accept a Callable that has the signature:

```python
def column_names_callable(df) -> Iterable[Hashable]:
```

we can enable `min_max_scale` on all columns by doing:

```python
df.min_max_scale(feature_range=(0, 1), column_names = lambda df: df.columns, jointly=True)
```

This is pretty concise without resorting to needing to maintain string mappings for special-case behaviour.

I'm glad we didn't rush to merge this PR, giving us the time and space to think clearly about the semantics of the API. 

@samukweku and @Zeroto521 what do you think about this?

Zeroto521: > I wanted to point out a new behaviour that we might be able to support across the rest of the API -- by making column_names accept a Callable that has the signature:
> ```python
> def column_names_callable(df) -> Iterable[Hashable]:
> ```

I'm sorry I still can't get the point why `column_names` could accept a callback type argument.

The usage of `column_names` is to get dataframe's columns like `df[column_names]`.
So if `column_names` has some column names which is not in `df.columns`.
There will raise an error. Whatever `column_names` is `Iterable[Hashable]` or callback type could return `Iterable[Hashable]`.

> we can enable min_max_scale on all columns by doing:
> ```python
> df.min_max_scale(feature_range=(0, 1), column_names = lambda df: df.columns, jointly=True)
> ```

To scale all columns, I thought we could use the default argument as `None` for `column_names` without no more inputting.

```python
df.min_max_scale(feature_range=(0, 1), column_names=None, jointly=True)
```

Are there more examples to show the importance of the callback type?

ericmjl: Thanks for the comments, @Zeroto521! 

Regarding why we might want to allow `column_names` to be a Callable, I had the idea that it helps support being explicit over implicit, which is in the Zen of Python. Setting `column_names=None` makes selecting all columns implicit, whereas setting `column_names=lambda df: df.columns` makes selecting all columns explicit. In addition, it allows the selection of arbitrary subsets of column names programmatically, without needing to hard-code those names.

On further thought, I can see how `column_names=None` actually follows the pattern established in other places in the library, so I think, for now, we can:

1. Use `column_names=None` to imply selection of all columns, and
2. Talk more about `column_names: Callable` in the issue tracker, deferring the implementation till later.

What do you all think about `jointly=True` as the keyword for triggering whether to _independently_ scale each column or to _jointly_ scale all columns specified in `column_names`? @Zeroto521 if you're in agreement with the keyword argument, then I think, let's get that specified in this PR, then we can close out the PR!

janitor/functions/min_max_scale.py
ericmjl: Yes! This sounds like a good idea!

At the same time, don't forget to add your contribution to the CHANGELOG.

ericmjl (approved):
@Zeroto521 I'm going to pre-approve the PR. Regarding whether to use `jointly` as the kwarg name, please take a decision and we'll run with it.

janitor/functions/min_max_scale.py
Zeroto521: I totally agree with using `jointly`.

About `column_names` whether could receive a callable type or not.
I can understand now. `column_names=lambda df: df.columns` is an implicit style and also a trick.

| Select columns                 | Using callable type                                                   | Using Iterable type | `df.columns`                 |
| ------------------------------ | --------------------------------------------------------------------- | ------------------- | ---------------------------- |
| Select the first three columns | `lambda df: df.columns[:3]`                              | `['a', 'b', 'c']`   | pd.Index(list('abcde'))      |
| Select str type columns        | `lambda df: [i for i in df.columns if instance(i, str)]` | `['a', 'b', 'c']`   | pd.Index(['a', 'b', 'c', 1]) |

As you said, we can put it aside at present.
Once the parameter `column_names` of `min_max_scale` could receive, other functions also need to do the same thing.

janitor/functions/min_max_scale.py
```diff
@@ -84,11 +99,16 @@ def min_max_scale(
...
+    :param jointly: (bool) Scale the entire data if Ture.
     :returns: A pandas DataFrame with scaled data.
     :raises ValueError: if `feature_range` isn't tuple type.
     :raises ValueError: if the length of `feature_range` isn't equal to two.
     :raises ValueError: if the element of `feature_range` isn't number type.
     :raises ValueError: if `feature_range[1]` <= `feature_range[0]`.
+
+    Changed in version 0.24.0: Deleted "old_min", "old_max", "new_min", and "new_max"
+    options.
+    Changed in version 0.24.0: Added "feature_range", and "jointly" options.
```
Zeroto521, line 111: If this renders well, then we can merge this PR.

janitor/functions/min_max_scale.py
Zeroto521: More discussions move to #1115

PR #1139, opened 2022-08-07
[BUG] Force `softmax` returning Series

Zeroto521:
# PR Description

```python
115         >>> import pandas as pd
116         >>> import janitor
117         >>> s = pd.Series([0, 1, 3], name="numbers")
118         >>> s.softmax()
Expected:
    0    0.042010
    1    0.114195
    2    0.843795
    Name: numbers, dtype: float64
Got:
    array([0.04201007, 0.1141952 , 0.84379473])
```

**This PR resolves https://github.com/pyjanitor-devs/pyjanitor/runs/7703616556?check_suite_focus=true .**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/math.py
+++ b/janitor/math.py
@@ -124,7 +124,8 @@ def softmax(s: pd.Series) -> pd.Series:
     :param s: Input Series.
     :return: Transformed Series.
     """
-    return scipy_softmax(s)
+
+    return pd.Series(scipy_softmax(s), index=s.index, name=s.name)
 
 
 @pf.register_series_method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 -   [ENH] Add `names_expand` and `index_expand` parameters to `pivot_wider` for exposing missing categoricals. Issue #1108 @samukweku
 -   [ENH] Add fix  for slicing error when selecting columns in `pivot_wider`. Issue #1134 @samukweku
 -   [ENH] `dropna` parameter added to `pivot_longer`. Issue #1132 @samukweku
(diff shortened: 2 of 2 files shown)


janitor/math.py
```diff
@@ -124,7 +124,8 @@ def softmax(s: pd.Series) -> pd.Series:
     :param s: Input Series.
     :return: Transformed Series.
     """
-    return scipy_softmax(s)
+
+    return pd.Series(scipy_softmax(s), index=s.index, name=s.name)
```
Zeroto521, line 128: Actually I'd like use `np.exp(s) / sum(np.exp(s))` instead of using scipy

samukweku (approved):

janitor/math.py
ericmjl: I think the SciPy implementation comes with numerical stability checks for free. Is there a reason why you would prefer to use a NumPy implementation?

ericmjl (approved):
I am good to go. I think we should keep the SciPy soft Mac implementation rather than switch to numpy, for numerical stability reasons (IIRC).

janitor/math.py
Zeroto521: To totally replace scipy. Since we only use it in `math.py` and `functions/impute.py`
- `mode` function in `functions/impute.py` could replace by `Series.value_counts`
- `math.py`
  - `sigmoid` (`expit`) could use `np.exp`
  - `softmax` (`scipy.special.softmax`) could use `np.exp`
  - `logit` (`scipy.special.softmax`) could use `np.log`
  - `normal_cdf` (`scipy.stats.norm.cdf`) and `probit` (`scipy.stats.norm.ppf`) can't replace easily.

PR #1133, opened 2022-07-14
[INF] Simplify environment

Zeroto521:
# PR Description

Please describe the changes proposed in the pull request:

- To keep minimal dependencies.
- Split environment
    - required: base, work for `functions`
    - optional: work for others
    - testing: work for testing
    - dev: work for the developer in local
    - doc: work for doc building

this `/environment-dev.yaml` will be kept for the developer to quickly install pyjanitor environment.
And `/ci/envs/*` will be created work for CI to test pyjanitor compatibility.

**This PR resolves #1127.**

# PR Checklist

Please ensure that you have done the following:

1. [ ] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [ ] If you're not on the contributors list, add yourself to `AUTHORS.md`.

3. [ ] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

# Automatic checks

There will be automatic checks run on the PR. These include:

- Building a preview of the docs on Netlify
- Automatically linting the code
- Making sure the code is documented
- Making sure that all tests are passed
- Making sure that code coverage doesn't go down.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #1147, opened 2022-08-14
[INF/DOC] Documentation bugfix by pinning mkdocstrings to 0.18.1

ericmjl:
@zeroto521 and @samukweku please merge as soon as the docs are built correctly on the docs preview.

### DIFF
--- a/tests/test_documentation_build.py
+++ b/tests/test_documentation_build.py
@@ -1,0 +1,24 @@
+"""Tests for documentation build."""
+
+import os
+
+
+def test_docs_general_functions_present():
+    """Test that all docs pages build correctly.
+
+    TODO: There has to be a better way to automatically check that
+    all of the functions are present in the docs.
+    This is an awesome thing that we could use help with in the future.
+    """
+    # Build docs using mkdocs
+    os.system("mkdocs build --clean")
+
+    # We want to check that the following keywords are all present.
+    # I put in a subsample of general functions.
+    # This can be made much more robust.
+    rendered_correctly = False
+    with open("./site/api/functions/index.html", "r+") as f:
+        for line in f.readlines():
+            if "add_columns" in line or "update_where" in line:
+                rendered_correctly = True
+    assert rendered_correctly
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           auto-update-conda: true
           miniforge-variant: Mambaforge
-          environment-file: mkdocs/environment.yaml
+          environment-file: environment-dev.yml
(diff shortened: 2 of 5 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1147--pyjanitor.netlify.app

Zeroto521:
Now documentation environment is `mkdocs/environment.yaml`.
I apply these you did to `environment-dev.yml` to `mkdocs/environment.yaml`

ericmjl:
I see, @Zeroto521. Thanks for the push. 

In working through this PR, I'm starting to see possible undesirable consequences of having separate environment definitions. Complexity is the main driver here. 

Previously, we could all rely on a single source of truth for the environment definition -- `environment-dev.yml`. With a separate `environment.yml` for docs, I didn't think to look at it (`mkdocs/environment.yaml`), because (1) it wasn't stored at the top level (which makes sense given what you were trying to accomplish) and (2) the GitHub actions configurations were not fully updated (i.e. only `docs-preview.yml` was updated, but not `docs.yml`). I should have reviewed your PR more stringently, so I'm partially responsible for this mess I found myself in.

One of the principles we have in the pyjanitor project is to make contributing easy for newcomers; having different `environment.yml` files may lead to downstream confusion if newcomers don't know where to look. As an old-timer on the project, I also missed out on this point; maybe I need to update my knowledge.

Going forward, I think we should have a deeper discussion about having separate environments for docs and development. Just as we try to standardize development, testing, and deployment environments with a single dev container, having a single `environment.yml` follows the same spirit. 

For now, I'm going to revert some of the changes that you made. This is just to fix broken docs, though, so I hope you don't feel that I'm rejecting the change outright. We can reintroduce them in a more comprehensive PR; if you could help me here by articulating (in PR #1133) what advantages we would get by your proposed changes, that will help stimulate the discussion further.

ericmjl:
@Zeroto521 and @samukweku I added a very rudimentary automated check for the docs. We may want to see whether there are better options out there, though. I will tag you both in a comment on the relevant file.

tests/test_documentation_build.py
```diff
@@ -0,0 +1,24 @@
+"""Tests for documentation build."""
```
ericmjl, line 1: @Zeroto521 and @samukweku this is the automated hack I have in place to check that our docs are built correctly automatically... at least for the general functions page. We may want to do a more comprehensive thing later.

ericmjl:
Finally, confirming that the docs built correctly on the deploy preview above:

image

ericmjl:
Ok, I'm going to take an "executive" decision and merge, so that our friends in the Eastern Hemisphere timezones can access usable docs as their work week begins.

Issue #1102, opened 2022-05-11
improvement suggestions for conditional_join


samukweku:
- [x] allow selection/renaming of final columns within the function
- [x] implement a numba alternative, for hopefully faster processing
- [x] improve performance for left and right joins

samukweku:
reopening this for the last issue with left/right joins ...

PR #1151, opened 2022-08-17
[ENH] Numba support for single non-equi joins

samukweku:
# PR Description

- Numba implementation for single non-equi joins (including support for first/last/all matches)
- Single file for all numba implementations, separate from the utils file
- Serves as a build layer for multiple non equi joins, with more efficiency

Speed tests ... YMMV ... usual disclaimer ... pinch of salt 

```py
In [47]: np.random.seed(3)
    ...: df = pd.DataFrame({'start':np.random.randint(100_000, size=1_000),
    ...:                    'end':np.random.randint(100_000, size=1_000)})
    ...: dd = pd.DataFrame({'ID':np.random.randint(100_000, size=50_000)})

In [48]: df.conditional_join(dd, ('start', 'ID', '<'), use_numba = True)
Out[48]: 
          start    end     ID
0            53  95502    110
1            53  95502     86
2            53  95502    121
3            53  95502     63
4            53  95502     72
...         ...    ...    ...
25318626  99838  92334  99892
25318627  99838  92334  99998
25318628  99838  92334  99999
25318629  99996  21475  99998
25318630  99996  21475  99999

[25318631 rows x 3 columns]

In [49]: %timeit df.conditional_join(dd, ('start', 'ID', '<'), use_numba = True)
502 ms ± 10.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [50]: %timeit df.conditional_join(dd, ('start', 'ID', '<'), use_numba = False)
685 ms ± 11.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [51]: %timeit df.merge(dd, how = 'cross').loc[lambda df: df.start < df.ID]
8.73 s ± 311 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [52]: A = df.conditional_join(dd, ('start', 'ID', '<'), use_numba = True)

In [53]: B = df.conditional_join(dd, ('start', 'ID', '<'), use_numba = False)

In [54]: C = df.merge(dd, how = 'cross').loc[lambda df: df.start < df.ID]

In [55]: A = A.sort_values(A.columns.tolist(), ignore_index = True)
    ...: C = C.sort_values(C.columns.tolist(), ignore_index=True)
    ...: B = B.sort_values(B.columns.tolist(), ignore_index = True)

In [56]:  B.equals(C)
Out[56]: True

In [57]: A.equals(C)
Out[57]: True

In [58]: A.equals(B)
Out[58]: True
```

**This PR is related to #1102 .**

# PR Checklist

2. [ ] If you're not on the contributors list, add yourself to `AUTHORS.md`.

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions/_numba.py
+++ b/janitor/functions/_numba.py
@@ -18,8 +18,8 @@ class _KeepTypes(Enum):
 
 
 def _numba_single_join(
-    left_c: pd.Series,
-    right_c: pd.Series,
+    left: pd.Series,
+    right: pd.Series,
     strict: bool,
     keep: str,
     op_code: int,
@@ -29,9 +29,9 @@ def _numba_single_join(
         # for the not equal join, we combine indices
         # from strictly less than and strictly greater than indices
         # as well as indices for nulls, if any
-        left_nulls, right_nulls = _numba_not_equal_indices(left_c, right_c)
+        left_nulls, right_nulls = _numba_not_equal_indices(left, right)
         dummy = np.array([], dtype=int)
-        result = _numba_less_than_indices(left_c, right_c)
+        result = _numba_less_than_indices(left, right)
         if result is None:
             lt_left = dummy
             lt_right = dummy
@@ -39,7 +39,7 @@ def _numba_single_join(
             lt_left, lt_right = _numba_generate_indices_ne(
                 *result, strict, keep, op_code=1
             )
-        result = _numba_greater_than_indices(left_c, right_c)
+        result = _numba_greater_than_indices(left, right)
         if result is None:
             gt_left = dummy
             gt_right = dummy
@@ -47,35 +47,37 @@ def _numba_single_join(
             gt_left, gt_right = _numba_generate_indices_ne(
                 *result, strict, keep, op_code=0
             )
-        left_c = np.concatenate([lt_left, gt_left, left_nulls])
-        right_c = np.concatenate([lt_right, gt_right, right_nulls])
-        if (not left_c.size) & (not right_c.size):
+        left = np.concatenate([lt_left, gt_left, left_nulls])
+        right = np.concatenate([lt_right, gt_right, right_nulls])
+        if (not left.size) & (not right.size):
             return None
         if keep == _KeepTypes.ALL.value:
-            return left_c, right_c
-        indexer = np.argsort(left_c)
-        left_c, pos = np.unique(left_c[indexer], return_index=True)
+            return left, right
+        indexer = np.argsort(left)
+        left, pos = np.unique(left[indexer], return_index=True)
         if keep == _KeepTypes.FIRST.value:
-            right_c = np.minimum.reduceat(right_c[indexer], pos)
+            right = np.minimum.reduceat(right[indexer], pos)
         else:
-            right_c = np.maximum.reduceat(right_c[indexer], pos)
-        return left_c, right_c
+            right = np.maximum.reduceat(right[indexer], pos)
+        return left, right
 
     if op_code == 1:
-        result = _numba_less_than_indices(left_c, right_c)
+        result = _numba_less_than_indices(left, right)
     else:
-        result = _numba_greater_than_indices(left_c, right_c)
+        result = _numba_greater_than_indices(left, right)
     if result is None:
         return None
     result = _get_regions(*result, strict, op_code)
     if result is None:
         return None
-    if keep == _KeepTypes.ALL.value:
-        return _numba_single_non_equi(*result)
     left_index, right_index, left_region, right_region = result
-    right_index, right_region = _prep_numba_first_last(
+    right_index, right_region = _prep_numba_sort_right(
         right_index, right_region
     )
+    if keep == _KeepTypes.ALL.value:
+        return _numba_single_non_equi(
+            left_index, right_index, left_region, right_region
+        )
     return _numba_single_non_equi_keep_first_last(
         left_index, right_index, left_region, right_region, keep
     )
@@ -102,22 +104,26 @@ def _numba_generate_indices_ne(
     )
     if result is None:
         return dummy, dummy
-    if keep == _KeepTypes.ALL.value:
-        return _numba_single_non_equi(*result)
     left_index, right_index, left_region, right_region = result
-    right_index, right_region = _prep_numba_first_last(
+    right_index, right_region = _prep_numba_sort_right(
         right_index, right_region
     )
+    if keep == _KeepTypes.ALL.value:
+        return _numba_single_non_equi(
+            left_index, right_index, left_region, right_region
+        )
     return _numba_single_non_equi_keep_first_last(
         left_index, right_index, left_region, right_region, keep
     )
 
 
-def _prep_numba_first_last(right_index, right_region):
+def _prep_numba_sort_right(right_index, right_region):
     """
-    Preparatory function if keep = 'first'/'last'
+    Sort right_index and right_region for faster searching
+    via binary search.
     """
-    if not pd.Series(right_region).is_monotonic_increasing:
+    bools = np.all(right_region[1:] >= right_region[:-1])
+    if not bools:
         indexer = np.lexsort((right_index, right_region))
         right_region = right_region[indexer]
         right_index = right_index[indexer]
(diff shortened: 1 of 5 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1151--pyjanitor.netlify.app

samukweku:
The docs are not showing the latest updates; I suspect it is just static at the moment.

janitor/functions/utils.py
```diff
@@ -425,87 +425,17 @@ def _column_sel_dispatch(columns_to_select, df):  # noqa: F811
     return filtered_columns
 
 
-class _KeepTypes(Enum):
+def _convert_to_numpy_array(
+    left_c: np.ndarray, right_c: np.ndarray
```
Zeroto521: Can't we directly use 'left' and 'right'?
What does the 'c' mean?

janitor/functions/_numba.py
```diff
@@ -0,0 +1,450 @@
+"""Various Functions powered by Numba"""
+
+import numpy as np
+import pandas as pd
+from enum import Enum
+from janitor.functions.utils import _convert_to_numpy_array
+from numba import njit, prange
+
+
+class _KeepTypes(Enum):
```
Zeroto521: Since we create `_numba.py` with `_` as the prefix.
I think we can remove `_` from the class and the method names.
like `KeepTypes` and `numba_single_join`.

Zeroto521: I'm not sure this change whether impact docstring rendering.

janitor/functions/_numba.py
```diff
@@ -0,0 +1,450 @@
...
+    ALL = "all"
+    FIRST = "first"
+    LAST = "last"
+
+
+def _numba_single_join(
+    left_c: pd.Series,
+    right_c: pd.Series,
+    strict: bool,
+    keep: str,
```
Zeroto521, line 13: Since py3.8, python support `Literal` like `keep: typing.Literal["all", "first", "last"]`.
It would be cleaner than `Enum` and `str`.

janitor/functions/_numba.py
```diff
@@ -0,0 +1,450 @@
...
+            check = arr[mid] > value
+        if check:
+            low = mid + 1
+        else:
+            high = mid
+    return high - 1
+
+
+@njit()
+def _searchsorted_right(arr: np.ndarray, value: int, op_code: int) -> int:
```
Zeroto521, line 411: numba can't speed up this function.

In my local test like the following case. 
```python
a = np.arange(1e7)
searchsorted_right(a, a[len(a)//2], 1)
```
with `njit` costing 1s, without `njit` (pure python) costing 300ms.

janitor/functions/_numba.py
```diff
@@ -0,0 +1,450 @@
...
+            high = _searchsorted_left(left_c, value, op_code)
+        else:
+            high = _searchsorted_right(left_c, value, op_code)
+        indices[num] = high
+
+    return indices
+
+
+@njit()
+def _searchsorted_left(arr: np.ndarray, value: int, op_code: int) -> int:
```
Zeroto521, line 389: it seems we can merge `_searchsorted_left` and `_searchsorted_right` into one method `_searchsorted`

Zeroto521 (commented):
Thanks for doing that.
Well, this is such a big change. And It's hard for me to review all parts at once. Plus I use `conditional_join` not very much.
I'm reviewing a small part of this PR first. From `typing.Literal`, name style, and `_searchsorted_*` starting.

janitor/functions/utils.py
samukweku: right is already a variable meant for the right dataframe, so we can't use that. hmmm, for the `_c` I can't even remember really. I am open to a better naming convention

janitor/functions/_numba.py
samukweku: not sure of the doctoring rendering, plus this is minor right

janitor/functions/_numba.py
samukweku: the goal is in its repeated use, plus they are just a modification which differs slightly from your regular `np.searchsorted`.

janitor/functions/_numba.py
samukweku: The comments/explanations in `_get_regions` are the reason for these two functions. The modification differs slightly from the regular bisect/np.searchsorted. Hopefully, the comments there are helpful

samukweku:
@Zeroto521 sure, think of `conditional_join` as an extension of merge but for inequality joins .... usually relevant for temporal joins or range joins; they are not as popular as your bread and butter merge joins, but quite relevant in some other fields. Efficient implementation avoids a cross join which can be very memory intensive and sometimes `O(n2)` time complexity. It is used often within SQL, and just feels like an obvious gap in the pandas ecosystem.  This [article]( https://duckdb.org/2022/05/27/iejoin.html) by DuckDB explains its value much better. 

The algorithm implementation here is quite efficient and should fill in the gap for this type of joins. The function name might in retrospect be a bad name, I guess it's too late (I don't even know what to change it to :grin:)

janitor/functions/_numba.py
ericmjl: Wow, I didn't realize that `Literal` was a thing. Thanks for educating, this is going to be much easier for documentation everywhere!

janitor/functions/_numba.py
ericmjl: Oh, I see. `njit` jit-compiles the thing, so on subsequent executions, the thing will be, pardon the greek, _bloody fast_.

janitor/functions/utils.py
ericmjl: `left_c` and `right_c` are just local variable names, is that right, @samukweku? If so, then we might be able to simplify a bit and switch to `left` and `right` just for the context of this function.

ericmjl (approved):

janitor/functions/_numba.py
samukweku: @Zeroto521 how do I use the Literal in code. I understand how it works in terms of type hinting and mypy (not sure we use mypy), but I do not understand how to use it in code. With the Enums, the values are declared once and then I can always access via `.value` to do comparison or some other operation. I can't, it seems, do that with the Literal type.

Zeroto521: How to use:

```python
from typing import Literal, get_args

# `Literal` is just an annotation to tell the user what option they can input.
# It's not a type, not a data structure.
KEEP_OPS = Literal["all", "first", "last", None, 1, "OrAnyOtherOption"]

def func(keep: KEEP_OPS="all"):
    # need to check first by us
    # there are some runtime 3rd packages to check use inputting and avoid us manually checking
    # but they are all in the developing stage.
    if keep not in get_args(KEEP_OPS):
        raise ValueError()

    if keep == 'all':  # more readable than Enum
        # do something

    return keep

# What you give to `keep`, that `keep` is you what gives.
func('all')  # -> 'all'
func(None)  # -> None
```


Something I worry about:

Currently, janitor requires py3.6. Actually could be py3.7 or py3.8. It's still not sure. Or doesn't test yet.
But this feature requires py3.8.
Lower python will raise importerror even via `from __future__ import annotations`.

janitor/functions/_numba.py
```diff
@@ -29,53 +29,55 @@ def _numba_single_join(
...
+        if (not left.size) & (not right.size):
             return None
         if keep == _KeepTypes.ALL.value:
-            return left_c, right_c
-        indexer = np.argsort(left_c)
-        left_c, pos = np.unique(left_c[indexer], return_index=True)
+            return left, right
+        indexer = np.argsort(left)
+        left, pos = np.unique(left[indexer], return_index=True)
         if keep == _KeepTypes.FIRST.value:
```
Zeroto521: Like this we could change to `if keep == 'first':`

samukweku: so basically get rid of `enum` and just use the strings directly?

Zeroto521: Yes, `Enum` is a little bit heavy for `keep`. `keep` only has three values. `Enum` is used to build a relationship between string and number, in my opinion.
Here we could just use string to replace `Enum`.
Once the required python version of janitor is set to 38, we just update documentation and typing

janitor/functions/_numba.py
ericmjl: Got it, let's raise the minimum Python from 3.6 to 3.8 in the next infra PR.

Zeroto521: Based on multi latest dependencies environments show that janitor can work with py3.6 and pandas 1.4.x.

But pandas 1.3.x requires py3.7.1 and pandas 1.4.x requires py3.8.
We should requires py3.7 at least although janitor can work well with py3.6 now.

But requiring py3.8 is optional.
I thought we could decide it after testing minimal dependencies environments on Ubuntu.

ericmjl:
@Zeroto521 I think we're good to merge. Next PR we should bump the minimum Python versions in the relevant places. I can get to work on that.

janitor/functions/conditional_join.py
```diff
@@ -483,220 +458,198 @@ def _conditional_join_compute(
         sort_by_appearance,
         df_columns,
         right_columns,
-        keep,
     )
 
 
 def _keep_output(keep: str, left_c: np.ndarray, right_c: np.ndarray):
```
Zeroto521:
```suggestion
def _keep_output(keep: str, left: np.ndarray, right: np.ndarray):
```

janitor/functions/conditional_join.py
```diff
@@ -483,220 +458,198 @@ def _conditional_join_compute(
...
     """return indices for left and right index based on the value of `keep`."""
-    if keep == _KeepTypes.ALL.value:
+    if keep == "all":
         return left_c, right_c
```
Zeroto521:
```suggestion
        return left, right
```

janitor/functions/conditional_join.py
```diff
@@ -483,220 +458,198 @@ def _conditional_join_compute(
...
     grouped = pd.Series(right_c).groupby(left_c)
```
Zeroto521:
```suggestion
    grouped = pd.Series(right).groupby(left)
```

Zeroto521 (approved):
After updating changelog file, we can merge this PR.

A small patches: removed `_c`

ericmjl:
@Zeroto521 would you like to do the honors?

PR #1143, opened 2022-08-11
[INF] Set multi-test env to compat dependency and system

Zeroto521:
# PR Description

Please describe the changes proposed in the pull request:

Aim:
- Set multi-test environment
- Compat different dependencies and systems

ToDo/Doing:
- [x] Set latest env: test pyjanitor work with the latest dependencies to get the minimal python version
- [ ] Set minimal env: get the minimal version of dependencies

Part of #1133

# PR Checklist

Please ensure that you have done the following:

1. [x] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [x] If you're not on the contributors list, add yourself to `AUTHORS.md`.

3. [x] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

# Automatic checks

There will be automatic checks run on the PR. These include:

- Building a preview of the docs on Netlify
- Automatically linting the code
- Making sure the code is documented
- Making sure that all tests are passed
- Making sure that code coverage doesn't go down.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #1157, opened 2022-08-21
Update Python to be minimum 3.9

ericmjl:
This PR updates our minimum Python version to be 3.9.

Key changes:

- We matrix Python version testing across 3.9, and 3.10 (current stable version and two before).
- We note everywhere relevant that the minimum Python version should be 3.9.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         test-subset: ["turtle", "not turtle"]
+        python-version: ["3.8", "3.9", "3.10"]
     runs-on: ubuntu-latest
     name: Run pyjanitor test suite
 
@@ -34,6 +35,7 @@ jobs:
       - name: Run unit tests
         run: |
           conda activate pyjanitor-dev
+          mamba install python=${{ matrix.python-version }}
           python -m pip install -e .
           pytest -m "${{ matrix.test-subset }}"
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         test-subset: ["turtle", "not turtle"]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10"]
     runs-on: ubuntu-latest
     name: Run pyjanitor test suite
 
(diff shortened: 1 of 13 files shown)


ericmjl:
@Zeroto521 I just realized this PR might overlap with your other PR #1143. I'll just use the PR to test some syntax things needed to make Python >=3.8 work. After that, we can discard my PR, and you can pick-and-choose whatever changes seem to make sense into your PR.

ericmjl:
🚀 Deployed on https://deploy-preview-1157--pyjanitor.netlify.app

Zeroto521:
Directly requires py3.9. Would that be a little bit aggressive?
pandas main branch still requires py3.8.
I'm okay to deprecate py3.6 and py3.7.

.github/workflows/tests.yml
```diff
@@ -34,6 +35,7 @@ jobs:
       - name: Run unit tests
         run: |
           conda activate pyjanitor-dev
+          mamba install python=${{ matrix.python-version }}
```
Zeroto521, line 38: those all cis actually use python 3.9.
https://github.com/pyjanitor-devs/pyjanitor/runs/7942582404?check_suite_focus=true#step:3:905

ericmjl:
Good question, @Zeroto521. Type annotations in Python 3.9 are more flexible, i.e. we can use built-in `tuple` or `list` as types like this:

```python
def func(df, arg1: tuple[int]):
    pass 
```

We can drop down to Python 3.8 if we fix the types to be:

```python
from typing import Tuple # (and/or List)

def func(df, arg1: Tuple[int]):
    pass
```

Zeroto521:
Py3.7+ with `__future__` also has that feature.

```python
from __future__ import annotations

int_array: list[int] = list(range(3))
None_or_bool: bool | None = True  # equals to `Optional[bool]`
```

thatlittleboy:
I also suggest using asottile's [`pyupgrade`](https://github.com/asottile/pyupgrade) to upgrade whatever syntax we can to the minimum version we are supporting. I also suggest sticking to py38 for now instead of jumping straight to py39. py37 EOL is coming this year.

Less manual work :)

ericmjl:
@thatlittleboy I'm in favour of more automation! 😺

Lemme take care of that in a bit.

ericmjl:
This PR is going stale. I'm going to close it and re-do it with minimum py38, as per @Zeroto521's and @thatlittleboy's suggestions.

PR #1163, opened 2022-08-31
[ENH/DOC] Support multi-column inputing, `change_type`

Zeroto521:
# PR Description

`col` could receive a single column or multi-column.
But currently, the documentation of `change_type` only shows supporting a single column.

```python
df[col] = df[col].astype(dtype)
```

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions/change_type.py
+++ b/janitor/functions/change_type.py
@@ -51,6 +51,16 @@ def change_type(
         1    1   5.0
         2    2   1.0
 
+        It also could be used to change the type of multiple columns at once.
+        If you want to change all columns to a specific type, please use
+        `DataFrame.astype` instead.
+
+        >>> df.change_type(['col1', 'col2'], str)
+          col1  col2
+        0    0     m
+        1    1     5
+        2    2  True 
+
     :param df: A pandas DataFrame.
     :param column_name: The column(s) in the dataframe.
     :param dtype: The datatype to convert to. Should be one of the standard
--- a/janitor/functions/change_type.py
+++ b/janitor/functions/change_type.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Hashable
+from typing import Any, Hashable
 
 import pandas as pd
 import pandas_flavor as pf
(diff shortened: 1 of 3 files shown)


janitor/functions/change_type.py
```diff
@@ -62,15 +65,21 @@ def change_type(
...
 
-        def convert(x, dtype):
-            """Casts item `x` to `dtype` or None if not possible."""
-            try:
-                return dtype(x)
-            except ValueError:
-                return None
-
-        df[column_name] = df[column_name].apply(lambda x: convert(x, dtype))
+        df[column_name] = df[column_name].apply(
```
Zeroto521, line 71: A little patch to `fillna`. Since we use `dtype(x)` logic. This only can handle a single element at once.
So we need to iterate the series, then iterate its values.

janitor/functions/change_type.py
```diff
@@ -62,15 +65,21 @@ def change_type(
...
+            lambda s: s.map(lambda x: _convert(x, dtype))
+        )
```
Zeroto521: A little patch to `fillna`. Since we use `dtype(x)` logic. This only can handle a single element at once.
So we need to iterate the series, then iterate its values.

Zeroto521: refactor via `.applymap` to convert all values

ericmjl (approved):
Approving! This is a great direction to take the semantics, @Zeroto521 -- anytime we want to make a change to columns, we can do it multiple columns at a time.

ericmjl:
When all tests pass, I will hit the merge button :smile:.

janitor/functions/change_type.py
```diff
@@ -50,7 +52,7 @@ def change_type(
         2    2   1.0
 
     :param df: A pandas DataFrame.
-    :param column_name: A column in the dataframe.
+    :param column_name: The column(s) in the dataframe.
```
thatlittleboy, line 68: How about also adding an example in the docstrings demonstrating that multiple columns are supported?

Because it's not obvious from the parameter name (`column_name`) that this is so

thatlittleboy (approved):
Lgtm, just a small suggestion docstring-wise

PR #1158, opened 2022-08-25
[ENH] Numba support for multiple non-equi joins

samukweku:
# PR Description

Please describe the changes proposed in the pull request:

- Numba implementation for multiple non-equi joins
- Much faster than the existing `_range_indices` implementation (which is limited to range joins)

### Speed Tests 
Disclaimer: take with a pinch of salt ... YMMV

data taken from [DuckDB](https://github.com/duckdb/duckdb/blob/master/benchmark/micro/join/iejoin_events.benchmark), which was referenced in this [DuckDB article](https://duckdb.org/2022/05/27/iejoin.html#simple-measurements)

```py
In [14]: events = pd.read_csv('https://github.com/samukweku/data-wrangling-blog/raw/master/_notebooks/Data_files/results.csv', parse_dates=['start', 'end'])

In [15]: events = events.iloc[:, 1:]

In [25]: (events
    ...: .conditional_join(
    ...:     events,
    ...:     ('start', 'end', '<='),
    ...:     ('end', 'start', '>='),
    ...:     ('id', 'id', '!='),
    ...:     df_columns='id',
    ...:     right_columns='id',
    ...:     use_numba=False)
    ...: )
Out[25]:
       left  right
         id     id
0        10   2345
1        15  11178
2        17  19605
3        26   8218
4        35   6916
...     ...    ...
3697  29966  29375
3698  29971  24173
3699  29978    981
3700  29984  19051
3701  29995  12296

[3702 rows x 2 columns]

In [26]: %%timeit
    ...: (events
    ...: .conditional_join(
    ...:     events,
    ...:     ('start', 'end', '<='),
    ...:     ('end', 'start', '>='),
    ...:     ('id', 'id', '!='),
    ...:     df_columns='id',
    ...:     right_columns='id',
    ...:     use_numba=False)
    ...: )
    ...:
    ...:
1.55 s ± 67.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [27]: %%timeit
    ...: (events
    ...: .conditional_join(
    ...:     events,
    ...:     ('start', 'end', '<='),
    ...:     ('end', 'start', '>='),
    ...:     ('id', 'id', '!='),
    ...:     df_columns='id',
    ...:     right_columns='id',
    ...:     use_numba=True)
    ...: )
    ...:
    ...:
21.1 ms ± 245 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [28]: (events
    ...: .conditional_join(
    ...:     events,
[...]

PR #1165, opened 2022-09-08
[ENH] explicit default parameter for case_when

### LINKED ISSUE #1159: improvement suggestions for `case_when`
~~- [ ] standalone case_when function, that can be used within assign; this can be handy for multiple case_whens to different column names~~
- [x] More explicit default parameter, currently it is inferred from the variable arguments
- [ ] performance improvement possibly, instead of iterating over all the options, just create matching rows only when needed?

samukweku:
# PR Description

~~- standalone function (`if_else`) that can be imported for use in pandas.assign~~
- explicit default parameter added

**This PR resolves #1159 .**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,305 +1,305 @@
 # Changelog
 
-## [Unreleased][Unreleased]
+## [Unreleased]
 
-- [DOC] Updated developer guide docs.
-- [ENH] Allow column selection/renaming within conditional_join. Issue #1102. Also allow first or last match. Issue #1020 @samukweku.
-- [ENH] New decorator `deprecated_kwargs` for breaking API. #1103 @Zeroto521
-- [ENH] Extend select_columns to support non-string columns. Also allow selection on MultiIndex columns via level parameter. Issue #1105 @samukweku
-- [ENH] Performance improvement for groupby_topk. Issue #1093 @samukweku
-- [ENH] `min_max_scale` drop `old_min` and `old_max` to fit sklearn's method API. Issue #1068 @Zeroto521
-- [ENH] Add `jointly` option for `min_max_scale` support to transform each column values or entire values. Default transform each column, similar behavior to `sklearn.preprocessing.MinMaxScaler`. (Issue #1067, PR #1112, PR #1123) @Zeroto521
-- [INF] Require pyspark minimal version is v3.2.0 to cut duplicates codes. Issue #1110 @Zeroto521
-- [ENH] Added support for extension arrays in `expand_grid`. Issue #1121 @samukweku
-- [ENH] Add `names_expand` and `index_expand` parameters to `pivot_wider` for exposing missing categoricals. Issue #1108 @samukweku
-- [ENH] Add fix  for slicing error when selecting columns in `pivot_wider`. Issue #1134 @samukweku
-- [ENH] `dropna` parameter added to `pivot_longer`. Issue #1132 @samukweku
-- [INF] Update `mkdocstrings` version and to fit its new coming features. PR #1138 @Zeroto521
-- [BUG] Force `math.softmax` returning `Series`. PR #1139 @Zeroto521
-- [INF] Set independent environment for building documentation. PR #1141 @Zeroto521
-- [DOC] Add local documentation preview via github action artifact. PR #1149 @Zeroto521
-- [ENH] Enable `encode_categorical` handle 2 (or more ) dimensions array. PR #1153 @Zeroto521
-- [ENH] Faster computation for a single non-equi join, with a numba engine. Issue #1102 @samukweku
-- [INF] Cancel old workflow runs via Github Action `concurrency`. PR #1161 @Zeroto521
-- [ENH] Add standalone function for `case_when`, as well an explicit default parameter. Issue #1159 @samukweku
+-   [DOC] Updated developer guide docs.
+-   [ENH] Allow column selection/renaming within conditional_join. Issue #1102. Also allow first or last match. Issue #1020 @samukweku.
+-   [ENH] New decorator `deprecated_kwargs` for breaking API. #1103 @Zeroto521
+-   [ENH] Extend select_columns to support non-string columns. Also allow selection on MultiIndex columns via level parameter. Issue #1105 @samukweku
+-   [ENH] Performance improvement for groupby_topk. Issue #1093 @samukweku
+-   [ENH] `min_max_scale` drop `old_min` and `old_max` to fit sklearn's method API. Issue #1068 @Zeroto521
+-   [ENH] Add `jointly` option for `min_max_scale` support to transform each column values or entire values. Default transform each column, similar behavior to `sklearn.preprocessing.MinMaxScaler`. (Issue #1067, PR #1112, PR #1123) @Zeroto521
+-   [INF] Require pyspark minimal version is v3.2.0 to cut duplicates codes. Issue #1110 @Zeroto521
+-   [ENH] Added support for extension arrays in `expand_grid`. Issue #1121 @samukweku
+-   [ENH] Add `names_expand` and `index_expand` parameters to `pivot_wider` for exposing missing categoricals. Issue #1108 @samukweku
+-   [ENH] Add fix  for slicing error when selecting columns in `pivot_wider`. Issue #1134 @samukweku
+-   [ENH] `dropna` parameter added to `pivot_longer`. Issue #1132 @samukweku
+-   [INF] Update `mkdocstrings` version and to fit its new coming features. PR #1138 @Zeroto521
+-   [BUG] Force `math.softmax` returning `Series`. PR #1139 @Zeroto521
+-   [INF] Set independent environment for building documentation. PR #1141 @Zeroto521
+-   [DOC] Add local documentation preview via github action artifact. PR #1149 @Zeroto521
+-   [ENH] Enable `encode_categorical` handle 2 (or more ) dimensions array. PR #1153 @Zeroto521
+-   [ENH] Faster computation for a single non-equi join, with a numba engine. Issue #1102 @samukweku
+-   [INF] Cancel old workflow runs via Github Action `concurrency`. PR #1161 @Zeroto521
+-   [ENH] Add standalone function for case_when, as well an explicit default parameter. Issue #1159 @samukweku
 
-## [v0.23.1][v0.23.1] - 2022-05-03
+## [v0.23.1] - 2022-05-03
 
-- [DOC] Updated `fill.py` and `update_where.py` documentation with working examples.
-- [ENH] Deprecate `num_bins` from `bin_numeric` in favour of `bins`, and allow generic `**kwargs` to be passed into `pd.cut`. Issue #969. @thatlittleboy
-- [ENH] Fix `concatenate_columns` not working on category inputs @zbarry
-- [INF] Simplify CI system @ericmjl
-- [ENH] Added "read_commandline" function to janitor.io @BaritoneBeard
-- [BUG] Fix bug with the complement parameter of `filter_on`. Issue #988. @thatlittleboy
-- [ENH] Add `xlsx_table`, for reading tables from an Excel sheet. @samukweku
-- [ENH] minor improvements for conditional_join; equality only joins are no longer supported; there has to be at least one non-equi join present. @samukweku
-- [BUG] `sort_column_value_order` no longer mutates original dataframe.
-- [BUG] Extend `fill_empty`'s `column_names` type range. Issue #998. @Zeroto521
-- [BUG] Removed/updated error-inducing default arguments in `row_to_names` (#1004) and `round_to_fraction` (#1005). @thatlittleboy
-- [ENH] `patterns` deprecated in favour of importing `re.compile`. #1007 @samukweku
-- [ENH] Changes to kwargs in `encode_categorical`, where the values can either be a string or a 1D array. #1021 @samukweku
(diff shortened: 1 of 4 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1165--pyjanitor.netlify.app

CHANGELOG.md
```diff
@@ -1,304 +1,305 @@
...
-    default by @mralbu
--   [ENH] add camelCase conversion to snake_case on `clean_names` by @ericmjl,
-    h/t @jtaylor for sharing original
--   [ENH] Added `null_flag` function which can mark null values in rows.
-    Implemented by @anzelpwj
--   [ENH] add engineering submodule with unit conversion method by @rahosbach
--   [DOC] add PyPI project description
--   [ENH] add example notebook with use of finance submodule methods
-    by @rahosbach
+## [Unreleased][Unreleased]
```
Zeroto521: I'm not sure whether you want to use 2-spaces to replace 4-space
If you want to use a 2-spaces style it's better to use cherry-pick this to another PR.

janitor/functions/case_when.py
```diff
@@ -1,14 +1,17 @@
...
 import pandas as pd
+from pandas.core.dtypes.inference import is_array_like
 
 from janitor.utils import check
 
 
 @pf.register_dataframe_method
-def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
+def case_when(
+    df: pd.DataFrame, *args, default: Optional[Any] = 0, column_name: str
```
Zeroto521:
```suggestion
    df: pd.DataFrame, *args, *, default: Optional[Any] = 0, column_name: str
```

I suggest using positional argument syntax.
It may avoid some hidden problems since `case_when` could receive `*args`.

janitor/functions/case_when.py
```diff
@@ -99,26 +102,51 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
-    :raises ValueError: If the condition fails to evaluate.
     :returns: A pandas DataFrame.
     """
-    conditions, targets, default = _case_when_checks(df, args, column_name)
+    check("column_name", column_name, [str])
+    default = if_else(df, args, default)
+    return df.assign(**{column_name: default})
+
+
+def if_else(df: pd.DataFrame, args, default: Optional[Any]) -> pd.Series:
```
Zeroto521: Whether renamed to `_if_else`, `if_else` isn't an accessor.

Zeroto521:
```python
default: Optional[Any]
```
The optional argument needs a default value.
`if_else(df, args)` without `default` will raise an error.

janitor/functions/case_when.py
```diff
@@ -128,71 +156,79 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
-            replacement = apply_if_callable(replacement, df)
-        targets.append(replacement)
+    booleans = [
+        apply_if_callable(condition, df)
+        if callable(condition)
+        else df.eval(condition)
+        if isinstance(condition, str)
+        else condition
+        for condition in booleans
+    ]
```
Zeroto521: I prefer the old one. The new one has 4 conditions which are hard to read.

janitor/functions/case_when.py
```diff
@@ -128,71 +156,79 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
+        apply_if_callable(replacement, df)
+        if callable(replacement)
+        else replacement
+        for replacement in replacements
+    ]
 
     if callable(default):
         default = apply_if_callable(default, df)
-    if not is_list_like(default):
+    if pd.api.types.is_scalar(default):
```
Zeroto521: directly use `is_scalar`?

janitor/functions/case_when.py
```diff
@@ -99,26 +102,51 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
+    """
 
+    booleans, replacements, default = _case_when_checks(df, args, default)
     # ensures value assignment is on a first come basis
-    conditions = conditions[::-1]
-    targets = targets[::-1]
-    for condition, value, index in zip(conditions, targets, count()):
+    booleans = booleans[::-1]
+    replacements = replacements[::-1]
+    for condition, value, index in zip(booleans, replacements, count()):
```
Zeroto521: use `enumerate` replace `itertools.count`?
So we can remove `import itertools` line.

janitor/functions/case_when.py
```diff
@@ -128,71 +156,79 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
+    if pd.api.types.is_scalar(default):
         default = pd.Series([default]).repeat(len(df))
-        default.index = df.index
-    if not hasattr(default, "shape"):
-        default = pd.Series([*default])
+    if not is_array_like(default):
+        raise TypeError(
+            "The argument for the `default` parameter "
+            "should evaluate to an array-like object, "
+            f"instead got {type(default)}"
```
Zeroto521:
```suggestion
            f"instead got {type(default)!r}"
```

This could output with `''`

Zeroto521 (changes requested):
Small patches need to be fixed.

janitor/functions/case_when.py
samukweku: Not sure what you mean here @Zeroto521

Zeroto521: This isn't a big deal. For this long path method, we would like 
to import is_scalar first at the script top. Then use use it directly.
You can ignore this. I'm okay to keep that.

janitor/functions/case_when.py
samukweku: this is so that it can be imported and used as a standalone within assign:

```py
from janitor.case_when import if_else
df.assign(a = if else(...), b = if else(...). ...)
```

Of course, if it is not a good idea, we can drop it

janitor/functions/case_when.py
```diff
@@ -99,100 +114,126 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
-    conditions, targets, default = _case_when_checks(df, args, column_name)
+    check("column_name", column_name, [str])
+    default = if_else(df, args, default)
+    return df.assign(**{column_name: default})
+
 
-    if len(conditions) == 1:
-        default = default.mask(conditions[0], targets[0])
-        return df.assign(**{column_name: default})
+def if_else(df: pd.DataFrame, args, default: Optional[Any] = 0) -> pd.Series:
```
ericmjl: @samukweku this is a question more than a critique -- what would make more semantic sense for the default value of the `default` argument? Should it be None/`np.nan` instead of 0? What are your thoughts here? I think you might have given more thought to this problem that I have, so I'm happy to defer to you here.

ericmjl (commented):
Just one question from me, @samukweku. Once we resolve it, I'm happy to approve! Other than that, we'll just need to resolve conflicts and then we're good.

janitor/functions/case_when.py
samukweku: I am open to criticism :grin: ... I just chose `numpy.select` default value ... thinking about it I'd say None is a better fit. open to suggestions

Zeroto521:
Two threads (https://github.com/pyjanitor-devs/pyjanitor/runs/8295244245, https://github.com/pyjanitor-devs/pyjanitor/runs/8295231459) CI failed due to new PR #1158.

`case_when` testcases raised some errors like 

```python
=========================== short test summary info ============================
ERROR tests/functions/test_case_when.py - hypothesis.errors.InvalidArgument: ...
ERROR tests/functions/test_case_when.py - hypothesis.errors.InvalidArgument: ...
ERROR tests/functions/test_case_when.py - hypothesis.errors.InvalidArgument: ...
ERROR tests/functions/test_case_when.py - hypothesis.errors.InvalidArgument: ...
```

janitor/functions/case_when.py
samukweku: I'll edit and let the user pass a default value instead

janitor/functions/case_when.py
samukweku: since args is already variable arg, this is fine, and no issues should arise

janitor/functions/case_when.py
samukweku: dropping the `if_else` idea altogether; user can do multiple case_when if necessary, similar to what obtains in SQL. In the future, users can change this, if there is enough interest

janitor/functions/case_when.py
```diff
@@ -99,100 +102,97 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
         For a 1-D array, it should have the same length as the DataFrame.
-        The `default` argument applies if none of `condition0`,
-        `condition1`, ..., evaluates to `True`.
-        Value can be a scalar, a callable, or a 1-D array. if `default` is a
-        callable, it should evaluate to a 1-D array.
+    :param default: scalar, 1-D array or callable.
+        If callable, it should evaluate to a 1-D array.
         The 1-D array should be the same length as the DataFrame.
+        The element inserted in the output when all conditions
+        evaluate to False.
```
thatlittleboy: My only concern is that this is a breaking change which probably warrants a deprecation cycle: users who depended on providing `2n+1` args (with the last arg treated as the "default"), will now be met with a TypeError due to a missing `default` kwarg.

I'm thinking if a better solution here is to (1) make `default` default to some value like `"__pyjanitor_default__"` (it can't be None since I suppose None is a valid default value), (2) introduce a DeprecationWarning to when we detect `default` is the special value from step1. Point the user towards using the explicit `default` kwarg. while manually popping the last arg into `default`. (3) Perform the remaining checks as if the user had passed in `default`, i.e., if `len(args)` is odd, then raise error.

Thoughts?

thatlittleboy (changes requested):
Just 1 concern on my end w.r.t. breaking changes. Hope we can address it before we merge. Thanks

janitor/functions/case_when.py
samukweku: Been a while @thatlittleboy hope you are well and good. 

I'm thinking of closing this PR and leave things as they are; the core is not really changed, maybe a future PR can improve performance. But yea, the breaking change is not worth it

thatlittleboy: Hey @samukweku , thanks, it's been a while. Hoping to jump back in some time soon to assist more, probably during the holiday season when work's a little lighter.

Back on topic: I do like the idea of specifying the `default` kwarg though, I think it's a step in the right direction. I've always found the (2n+1) "API" for `case_when` a little confusing to read when written out as code. The breaking change just needs a little careful handling.

I'll suggest keeping the PR open instead of closing for now, someone else can pick up from where you left off (I'll volunteer, if I can find the time :x ).

samukweku: @thatlittleboy made some edits to the code to cater for deprecation. Would love your feedback on this. @ericmjl @Zeroto521 pls feel free to chime in when you have some bandwidth

tests/functions/test_case_when.py
```diff
@@ -64,6 +64,21 @@ def test_args_even(dataframe):
...
+    is odd and `default` is None.
+    """
+    with pytest.warns(DeprecationWarning):
+        dataframe.case_when(
+            dataframe.a < 10,
+            "less_than_10",
+            dataframe.a == 5,
+            default=None,
+            column_name="a",
+        )
```
thatlittleboy: I think it's more important to ensure in this test that `DeprecationWarning` is raised when the user is writing in the style from before. i.e., I suggest we remove the `default=None`.

janitor/functions/case_when.py
```diff
@@ -99,100 +104,100 @@ def case_when(df: pd.DataFrame, *args, column_name: str) -> pd.DataFrame:
...
         default = pd.Series([default]).repeat(len(df))
-        default.index = df.index
-    if not hasattr(default, "shape"):
-        default = pd.Series([*default])
+    if not is_array_like(default):
+        raise TypeError(
+            "The argument for the `default` parameter "
+            "should evaluate to an array-like object, "
+            f"instead got {type(default)!r}"
+        )
```
thatlittleboy: Just curious why we are raising and error rather than converting the list/tuple/etc. to a Series (as we've done before)?

thatlittleboy (approved):
Thanks Sam, just a couple questions left from me, but nothing major. Pre-approving.

smiling-cat-face-lgtm

tests/functions/test_case_when.py
samukweku: @thatlittleboy Do you mind scribbling what U mean so I can better understand?

janitor/functions/case_when.py
samukweku: Good catch. Lemme fix

tests/functions/test_case_when.py
thatlittleboy:
```python
    with pytest.warns(DeprecationWarning):
        dataframe.case_when(
            dataframe.a < 10,
            "less_than_10",
            dataframe.a == 5,
            column_name="a",
        )
```
I just mean this ^ :)

(When I said "style from before", I meant "before this PR / before the breaking change is introduced".)

ericmjl:
@thatlittleboy would you like to do the honors here? a -0.01% decrease in coverage is marginal, we can handle it :smile:

thatlittleboy:
yep okies, merging

PR #1166, opened 2022-09-08
[BUG] Modify `concatenate_columns`  ignore_empty output

### LINKED ISSUE #1164: [BUG] `concatenate_columns` Ignoring null values is not available
## Source Code
```Python
@pf.register_dataframe_method
@deprecated_alias(columns="column_names")
def concatenate_columns(
    df: pd.DataFrame,
    column_names: List[Hashable],
    new_column_name: Hashable,
    sep: str = "-",
    ignore_empty: bool = True,
) -> pd.DataFrame:

    if len(column_names) < 2:
        raise JanitorError("At least two columns must be specified")


################# Need  modified  ###################
    df[new_column_name] = (
        df[column_names].astype(str).fillna("").agg(sep.join, axis=1)
    )
######## ############ to ###############################
    df[new_column_name] = (
        df[column_names].astype('string').fillna("").agg(sep.join, axis=1)
    )


    if ignore_empty:

        def remove_empty_string(x):
            """Ignore empty/null string values from the concatenated output."""
            return sep.join(x for x in x.split(sep) if x)

        df[new_column_name] = df[new_column_name].transform(
            remove_empty_string
        )

    return df
```
## Like this, np.nan becomes 'nan' in astype str
image

## If  change it to this, there is no problem
image

Fu-Jie:
# PR Description

- modify concatenate_column  ignore_empty output (pd.NA,pd.NaT,None,np.nan => "")
- modify concatenate_columns test case

**This PR resolves #1164.**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


AUTHORS.md
```diff
@@ -109,3 +109,4 @@ Contributors
 - [@ethompsy](https://github.com/ethompsy) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues?q=is%3Aclosed+mentions%3Aethompsy)
 - [@apatao](https://github.com/apatao) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues?q=is%3Aclosed+mentions%3Aapatao)
 - [@OdinTech3](https://github.com/OdinTech3) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/pull/1094)
+- [@Fu-Jie](https://github.com/Fu-Jie) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues/1164)
```
Zeroto521: This could index all merged PRs.

```suggestion
- [@Fu-Jie](https://github.com/Fu-Jie) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/pulls?q=is%3Aclosed+mentions%3AFu-Jie)
```

janitor/functions/concatenate_columns.py
```diff
@@ -52,7 +52,10 @@ def concatenate_columns(
         raise JanitorError("At least two columns must be specified")
```
Zeroto521, line 53:
```suggestion

    df = df.copy()  # avoid mutating original data
```

I note this function also mutates the original `df`. This is a bug also.
Please update this into changelog. Could you have a test?

```python
>>> df = pd.DataFrame([[1, 2], [3, None]], columns=['a', 'b'])
>>> df.concatenate_columns(
...     column_names=["a", "b"], new_column_name="m",
... )
   a    b      m
0  1  2.0  1-2.0
1  3  NaN  3-nan
>>> df  # changed
   a    b      m
0  1  2.0  1-2.0
1  3  NaN  3-nan
```

Zeroto521 (changes requested):
Thanks for fixing this problem.

I thought `ignore_empty=True` we could fill nan.

```python
    df[new_column_name] = (
        df[column_names].fillna("").astype(str).agg(sep.join, axis=1)
        if ignore_empty
        else df[column_names].astype(str).agg(sep.join, axis=1)
    )
```


And It's better to do the fill operation and then do the change type operation.
It's hard to say how many nan strings there have.

```python
df.fillna("").astype(str)

df.astype(str).replace(["NaT", "nan", "<NA>"], "")
```

janitor/functions/concatenate_columns.py
```diff
@@ -52,7 +52,10 @@ def concatenate_columns(
...
 
     df[new_column_name] = (
-        df[column_names].astype(str).fillna("").agg(sep.join, axis=1)
+        df[column_names]
+        .astype(str)
+        .replace(["NaT", "nan", "<NA>"], "")
+        .agg(sep.join, axis=1)
```
ericmjl: We might want an argument here to toggle between old and new behaviours. Would you be open to doing so?

ericmjl (changes requested):
@Fu-Jie thank you for your contribution here! I'm noticing that the change will likely be a breaking change, i.e. it modifies the old expected behaviour of the function. Can we ensure that the suggested changes are toggleable via function arguments?

AUTHORS.md
Fu-Jie: Fillna first, then astype, can not handle columns of type int or float

janitor/functions/concatenate_columns.py
Fu-Jie: I'm sorry, the translation software I used may not have described it clearly。
For ignoring null values, my idea comes from the implementation of Excel, feeling that the implementation of Excel is more in line with the actual use
https://support.microsoft.com/en-us/office/textjoin-function-357b449a-ec91-49d0-80c3-0e8fc845691c

Fu-Jie: @ericmjl 
It is my understanding that if null values are not ignored, then this is more reasonable.
##  Pseudo-code
```Python  
split = ','
if ignore_empty =  False   then  1,2,pd.NA  ->  1,2,
if ignore_empty =  True    then  1,2,pd.NA  ->  1,2
```

Fu-Jie:
> And It's better to do the fill operation and then do the change type operation. It's hard to say how many nan strings there have.
> 
> ```python
> df.fillna("").astype(str)
> 
> df.astype(str).replace(["NaT", "nan", "<NA>"], "")
> ```
 @Zeroto521  I am test None、pd.NA、pd.NaT、np.nan，there should be nothing else.

Fu-Jie:
> Thanks for fixing this problem.
> 
> I thought `ignore_empty=True` we could fill nan.
> 
> ```python
>     df[new_column_name] = (
>         df[column_names].fillna("").astype(str).agg(sep.join, axis=1)
>         if ignore_empty
>         else df[column_names].astype(str).agg(sep.join, axis=1)
>     )
> ```
@Zeroto521  
At first, I also thought of this solution，but there are some problems:
 Fillna first, then astype, can not handle columns of type int or float。

Zeroto521:
```python
import numpy as np
import pandas as pd


def fillna_astype(df, sep="-"):
    return df.fillna("").astype(str).agg(sep.join, axis=1)


def astype_fillna(df, sep="-"):
    return df.astype(str).replace(["NaT", "nan", "<NA>"], "").agg(sep.join, axis=1)

# normal case
# both of them passed, but `astype_fillna` need to replace `None` and `'NaN'`.
pd.DataFrame(
    {
        "a": ["string", 1, 1.5, np.nan],
        "b": ["another_string", 0, pd.NA, None],
    }
).pipe(fillna_astype)
# 0    string-another_string
# 1                      1-0
# 2                     1.5-
# 3                        -
# dtype: object

pd.DataFrame(
    {
        "a": ["string", 1, 1.5, np.nan],
        "b": ["another_string", 0, pd.NA, None],
    }
).pipe(astype_fillna)
# 0    string-another_string
# 1                      1-0
# 2                     1.5-
# 3                    -None
# dtype: object


# this one is a special case. `astype_fillna` is failed.
# we only want to fill na value.
pd.DataFrame(
    {
        "a": ["string", np.nan, pd.NA, None],
        "b": ["another_string", "nan", "<NA>", "None"],
    }
).pipe(fillna_astype)
# 0    string-another_string
# 1                     -nan
# 2                    -<NA>
# 3                    -None
# dtype: object

pd.DataFrame(
    {
        "a": ["string", np.nan, pd.NA, None],
        "b": ["another_string", "nan", "<NA>", "None"],
    }
).pipe(astype_fillna)
# 0    string-another_string
# 1                        -  # wrong
# 2                        -  # wrong
# 3                None-None  # wrong
# dtype: object
```

Fu-Jie:
@Zeroto521 

### normal case
It could be a pandas(1.3.5) version issue,my env both None and np.nan astype for "nan" ,Should need to increase the na value.
```Python

def astype_fillna(df, sep="-"):
    return df.astype(str).replace(["NaT", "nan", "<NA>","None"], "").agg(sep.join, axis=1)


```
### about fillna astype float or int issue
```Python

import pandas as pd 
def fillna_astype(df, sep="-"):
    return df.fillna("").astype(str).agg(sep.join, axis=1)
pd.DataFrame(
    {
        "b": [1, 0, pd.NA, 3],
    }
    ,dtype=pd.Float32Dtype()
).pipe(fillna_astype)
##
## TypeError: <U1 cannot be converted to a FloatingDtype
```

### special case
In my opinion, what should be dealt with is the null value of the column, not the text that represents the empty meaning.

Zeroto521:
I think that isn't a good example for `fillna_astype`.
Once you point out the type of data, some of the methods you can't use. This is what's the dtype meaning.

> ```python
> import pandas as pd 
> def fillna_astype(df, sep="-"):
>     return df.fillna("").astype(str).agg(sep.join, axis=1)
> pd.DataFrame(
>     {
>         "b": [1, 0, pd.NA, 3],
>     }
>     ,dtype=pd.Float32Dtype()
> ).pipe(fillna_astype)
> ##
> ## TypeError: <U1 cannot be converted to a FloatingDtype
> ```

Fu-Jie:
> I think that isn't a good example for `fillna_astype`. Once you point out the type of data, some of the methods you can't use. This is what's the dtype meaning.
> 
> > ```python
> > import pandas as pd 
> > def fillna_astype(df, sep="-"):
> >     return df.fillna("").astype(str).agg(sep.join, axis=1)
> > pd.DataFrame(
> >     {
> >         "b": [1, 0, pd.NA, 3],
> >     }
> >     ,dtype=pd.Float32Dtype()
> > ).pipe(fillna_astype)
> > ##
> > ## TypeError: <U1 cannot be converted to a FloatingDtype
> > ```
This situation may appear in the read_sql int column（pandas dafaultl parse） or read_excel specify the dtype = 'int'
In addition to this problem, I also think it will be better to use Fillna first。

samukweku:
Any thoughts on the progress of this PR @thatlittleboy @ericmjl @Zeroto521 ? @Fu-Jie kindly rebase so that  this PR is updated to the latest

janitor/functions/concatenate_columns.py
```diff
@@ -51,8 +51,12 @@ def concatenate_columns(
     if len(column_names) < 2:
         raise JanitorError("At least two columns must be specified")
 
+    df = df.copy()  # avoid mutating original data
```
thatlittleboy, line 54: Now that we've added this, the function is no longer mutating.Can we also update the docstring?

tests/functions/test_concatenate_columns.py
```diff
@@ -28,7 +28,7 @@ def test_concatenate_columns_null_values(missingdata_df):
         new_column_name="index",
         ignore_empty=True,
     )
-    expected_values = ["1.0-1", "2.0-2", "nan-3"] * 3
+    expected_values = ["1.0-1", "2.0-2", "3"] * 3
```
thatlittleboy, line 31: Please also update the docstrings for the test.

tests/functions/test_concatenate_columns.py
```diff
@@ -28,7 +28,7 @@ def test_concatenate_columns_null_values(missingdata_df):
...
     assert expected_values == df["index"].tolist()
```
thatlittleboy, line 34: I also think it might be worth writing a test merging a custom dataframe with a float column (NaN), a datetime column (NaT) and a string column (None/NA?).

And assert the expected output accordingly.

Then, mention this PR or the attached issue in the test docstring as well, please.

thatlittleboy (changes requested):
I think we're close! just need to clear up some inconsistencies against the docstrings

Issue #1176, opened 2022-10-16
visualization of pyjanitor chained method calls


asmirnov69:
# Brief Description

I'd like to propose some new features to pyjanitor with focus to visualization and data organization. I've made proof-of-concept repo to explain what exactly is the proposal contains: https://github.com/asmirnov69/pyjviz-poc

To be really brief - this example of [pyjanitor example](https://github.com/asmirnov69/pyjviz-poc/blob/main/scripts/dirty-clean.py) with corresponding [diagram](https://github.com/asmirnov69/pyjviz-poc/blob/main/scripts/rdflog/dirty-clean.ttl.dot.png)

# Examples

more examples are [here](https://github.com/asmirnov69/pyjviz-poc/tree/main/scripts)

# A bit more about what is goal beyond immediate focus

Immediate focus is to have png diagram files generation working using rdflib with provided SPARQL and graphviz. There is a bigger idea of rdf logs and similarily collected data to be stored in graph database. This would be database of research&production activity which uses SPARQL and/or opencypher to provide the way to connect collected data to other knowledge graph systems (e.g. Obsidian).

asmirnov69:
I was unable to find the way to attach labels. My intent was to label this issue as 'disscussion-needed'

ericmjl:
Hl @asmirnov69! Thanks for posting this issue. It's a really cool idea! I especially like being able to visualize both pandas and janitor functions simultaneously.

Because the POC implementation is a tad complex, would you be kind enough to point out where the key changes were needed in order to enable keeping track of which function/method calls were made? I read through the code, but was confused, as I didn't see something like a globally-instantiated NetworkX graph object (or analogous thing).

Additionally, I saw the use of a new `ChainedMethodPipe` object. Is the intent with the ChainedMethodPipe to keep transformations visually isolated from one another?

I'd love to see this functionality implemented. We could probably house this in `pyjanitor`; if so, I can foresee the PR review process needing to be a bit longer than usual so that there's enough time for other maintainers to be brought up-to-speed on the implementation; we'd probably also need time to have developer/maintainer documentation written, focused particularly on:

1. what are the most likely ways this code would break in the future, and what's the most likely place to go and fix, and 
2. what are desired possible future improvements not covered in this PR, and where should the code changes be made?
3. answering questions that we bring up during the review process, for e.g. architecture docs answering the questions I raised above.

Zeroto521:
There is a reference from sklearn's Pipeline

[...]

PR #1180, opened 2022-10-18
[ENH] Add lazy imports to speed up the time taken to load pyjanitor (part 2)

### LINKED ISSUE #1059: Enable lazy importing for external dependencies

asmirnov69:
# PR Description

**This PR resolves #1059**

Code changes are taken from original PR https://github.com/pyjanitor-devs/pyjanitor/pull/1060 and placed to current dev branch head.

# PR Checklist

1. [X] PR in from a fork off your branch.
2. [X] If you're not on the contributors list, add yourself to `AUTHORS.md`.
3. [X] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.

# Automatic checks

# Relevant Reviewers

@ericmjl

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -109,3 +109,4 @@ Contributors
 - [@ethompsy](https://github.com/ethompsy) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues?q=is%3Aclosed+mentions%3Aethompsy)
 - [@apatao](https://github.com/apatao) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues?q=is%3Aclosed+mentions%3Aapatao)
 - [@OdinTech3](https://github.com/OdinTech3) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/pull/1094)
+- [@asmirnov69](https://github.com/asmirnov69) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues/1059)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+-   [ENH] Add lazy imports to speed up the time taken to load pyjanitor (part 2)
 -   [DOC] Updated developer guide docs.
 -   [ENH] Allow column selection/renaming within conditional_join. Issue #1102. Also allow first or last match. Issue #1020 @samukweku.
 -   [ENH] New decorator `deprecated_kwargs` for breaking API. #1103 @Zeroto521
--- a/janitor/__init__.py
+++ b/janitor/__init__.py
@@ -29,4 +29,4 @@ def get_features_targets(*args, **kwargs):
     return _get_features_targets(*args, **kwargs)
 
 
-__version__ = "0.23.1-lazy"
+__version__ = "0.23.1"
--- a/janitor/__init__.py
+++ b/janitor/__init__.py
@@ -1,15 +1,24 @@
-"""Top-level janitor API lives here."""
-try:
-    import janitor.xarray  # noqa: F401
-except ImportError:
-    pass
+"""Top-level janitor API lives here.
 
+Lazy loading used here to speed up imports.
+"""
+import lazy_loader as lazy
+
+
+from .accessors import *  # noqa: F403, F401
 from .functions import *  # noqa: F403, F401
 from .io import *  # noqa: F403, F401
 from .math import *  # noqa: F403, F401
 from .ml import get_features_targets as _get_features_targets
 from .utils import refactored_function
-from .accessors import *  # noqa: F403, F401
+
+
+_ = lazy.load("pandas_flavor")
+
+try:
+    jxr = lazy.load("janitor.xarray")  # noqa: F401
+except ImportError:
+    pass
 
 
 @refactored_function(
@@ -20,4 +29,4 @@ def get_features_targets(*args, **kwargs):
     return _get_features_targets(*args, **kwargs)
 
 
-__version__ = "0.23.1"
+__version__ = "0.23.1-lazy"
--- a/janitor/functions/impute.py
+++ b/janitor/functions/impute.py
@@ -1,12 +1,14 @@
 """Implementation of `impute` function"""
 from typing import Any, Hashable, Optional
 
+import lazy_loader as lazy
 import numpy as np
 import pandas_flavor as pf
 import pandas as pd
-from scipy.stats import mode
 
 from janitor.utils import deprecated_alias
+
+ss = lazy.load("scipy.stats")
 
 
 @pf.register_dataframe_method
@@ -100,7 +102,7 @@ def impute(
         "mean": np.mean,
         "average": np.mean,  # aliased
         "median": np.median,
-        "mode": mode,
+        "mode": ss.mode,
         "minimum": np.min,
         "min": np.min,  # aliased
         "maximum": np.max,
--- a/.requirements/base.in
+++ b/.requirements/base.in
@@ -6,3 +6,4 @@ natsort
 pandas_flavor
 multipledispatch
 scipy
+lazy_loader
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -55,3 +55,4 @@ dependencies:
   - xorg-libxrender
   - pip:
       - mknotebooks
+      - lazy-loader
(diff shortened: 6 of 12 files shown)


ericmjl:
Hi @asmirnov69, thanks for taking the initiative on this PR!

Can I ask, when you benchmark import times on your own machine, do you observe a speedup in import times?

I'm asking b/c on a fresh Codespace, running the import benchmark gets me:

image

However, a subsequent run gets me:

image

As you probably can tell, the absolute first clean import on a machine will take ~6 seconds, and subsequent imports will be fast. I'm not sure whether this will be desired behaviour or not, though. What are your thoughts here?

asmirnov69:
Hi @ericmjl 

Looks like first run is busy caching bytecode into `__pycache__` dirs - in your case it is way too slow. Difference and the timings you observe are much bigger that I can see on my ubuntu 22.04 4 cpu 16Gb in aws. All my tests are faster than 1 sec - including test with no cached bytecode. 
I will try to do one more test on fresh instance, will let you know if I will observe significant slow down on first time use - but I think something is happening in your system.

Few immediate observations:
- tuna.png doesn't go through pre-commit - file is ~800k which is bigger than allowed 500k
- simple timing shows improvement in import time ~30%. Removing of `__pycache__` dirs slow both tests down ~8%. The slowdown if -B cli options is used to prevent bytecode caching is the same 8% as expected
```
time python -c 'import janitor' # real time ~0.7 secs
time python -B -c 'import janitor' # 
python -X importtime -c 'import janitor' |& tail # look at the last line
python -B -X importtime -c 'import janitor' |& tail # 
```
30% improvement is smaller gain than you've seen (~50%). But looks like it make sense to use lazy_import in pyjanitor since gain is here and if anything goes wrong with future development it should be easy to change import statements.

<hr>
In general slow startup time of some script due to huge package dependencies could become the problem elsewhere too.
E.g. usual executables or shared objects sometimes loaded before program start using LD_PRELOAD (also used if you have circular dependencies). Maybe python3 team will add something like this in python itself in future.
<br/><br/>
But the real solution could be proper split of pyjanitor to pluggable modules (as mentioned in #826). pyjanitor list of mandatory dependencies is too long.

In long term I am for packing of pyjanitor using [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) which I think is proper answer on how to implement the ask of #826: 

> Is the following possible?
> pip install pyjanitor[base]

BTW, I can explore what kind of visualization support can be added to my proposal #1176 to start addressing package dependencies problem. The idea could be to collect import dependencies into rdf triple form and then maybe have some useful diagram out of that - e.g. pyjanitor submodules dependencies graph.

asmirnov69:
What should be done about codecov checks failures? Is it mandatory precondition to have all checks green to have contribution accepted?

ericmjl:
@asmirnov69 thanks for the detailed response, and for putting in the time and effort to do profiling on your end.

Considering what you've written, I think we're good with the PR. 

With respect to codecov, the big ones to look out for are that all tests pass (primarily), and that code coverage on the *project* doesn't decrease significantly. In this case, a decrease by 0.41% is acceptable.

ericmjl (approved):
Approving the PR. Let's see how this works out on `dev` (i.e. make sure all tests still pass).

asmirnov69:
@ericmjl do i need to do anything at this point? who is going to complete this PR with merge commit to dev?

ericmjl:
We'll need one more review from the team. I just asked for @samukweku, @thatlittleboy, and @Zeroto521. As a rule of thumb, we try to go for 2 reviews (but time-bound to ~1-1.5 weeks of having a PR open.)

samukweku (approved):

AUTHORS.md
```diff
@@ -109,3 +109,4 @@ Contributors
 - [@ethompsy](https://github.com/ethompsy) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues?q=is%3Aclosed+mentions%3Aethompsy)
 - [@apatao](https://github.com/apatao) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues?q=is%3Aclosed+mentions%3Aapatao)
 - [@OdinTech3](https://github.com/OdinTech3) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/pull/1094)
+- [@asmirnov69](https://github.com/asmirnov69) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/issues/1059)
```
Zeroto521, line 112: It could index all.
```suggestion
- [@asmirnov69](https://github.com/asmirnov69) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/pulls?q=is%3Aclosed+mentions%3Aasmirnov69)
```

Zeroto521, line 112: It could index all.
```suggestion
- [@asmirnov69](https://github.com/asmirnov69) | [contributions](https://github.com/pyjanitor-devs/pyjanitor/pulls?q=is%3Aclosed+mentions%3Aasmirnov69)
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+-   [ENH] Add lazy imports to speed up the time taken to load pyjanitor (part 2)
```
Zeroto521, line 5:
```suggestion
-   [ENH] Add lazy imports to speed up the time taken to load pyjanitor. Issue #1059 @asmirnov69.
```

janitor/__init__.py
```diff
@@ -20,4 +29,4 @@ def get_features_targets(*args, **kwargs):
     return _get_features_targets(*args, **kwargs)
 
 
-__version__ = "0.23.1"
+__version__ = "0.23.1-lazy"
```
Zeroto521: I'm unsure whether adding a suffix will have an impact on `bumpversion`.
Once the CI of `auto-release.yml` is trigged, the version will be like `0.24.0` or `0.24.0-lazy`?

janitor/functions/impute.py
```diff
@@ -1,13 +1,15 @@
 """Implementation of `impute` function"""
 from typing import Any, Hashable, Optional
 
+import lazy_loader as lazy
 import numpy as np
 import pandas_flavor as pf
 import pandas as pd
```
Zeroto521, line 7: I saw all packages in `janitor/math.py` were wrapped by `lazy.load`.
But `janitor/functions/impute.py` only wrapped `scipy.stats`.

Zeroto521 (changes requested):
Some small patches.

Since `lazy_loader` is imported in `janitor/__init__.py`. It's a necessary package.
So It need to add into `.requirements/base.in`.

janitor/__init__.py
asmirnov69: suffix -lazy to __version__ value  was introduced by me since I did't know what to do about version in the case 
of PR branch.
so feel free to drop it and set __version__ appropriately to current situation

janitor/functions/impute.py
asmirnov69: i didn't try to go beyond original @ericmjl code changes. 
it could be true that some newer additions on dev would require changes in imports. i assume we can address it later in subsequent PRs.

janitor/__init__.py
Zeroto521: I'd like to reverse this change to version first.

janitor/functions/impute.py
ericmjl: Good point, @Zeroto521. We can continue to improve coverage of lazy loading, so no worries here. Plus we might also benefit from Python having lazy loading by default! (I'm thinking of [PEP690](https://peps.python.org/pep-0690/) here.) Let's stick with what we have for now.

janitor/__init__.py
ericmjl: I agree with you, @Zeroto521, very astute catch that I didn't see before! @asmirnov69 could you remove the `-lazy` here? We'll take care of versioning in our build pipeline. That is probably the only thing I'll ask of before we merge.

asmirnov69: will do:
- \_\_version\_\_ will be 0.23.1
- lazy_loader will be added to .requirements/base.in

asmirnov69: done

Zeroto521 (approved):

ericmjl:
Three approvers! I am going to hit the merge button :smile:.

PR #1185, opened 2022-10-31
[INF] Merge `codecov.yml` into `tests.yml`

Zeroto521:
# PR Description

Please describe the changes proposed in the pull request:

Fix code coverage difference in the dev branch, PR, and envs.
Get the same coverage result from the dev branch and PR.

The way of launching pytest is thiny different. And that causes the code coverage sometimes passed sometimes failed.
image
https://github.com/pyjanitor-devs/pyjanitor/blob/8445dc00c42728df18b461d0f7df337b7cc96f18/.github/workflows/tests.yml#L42
https://github.com/pyjanitor-devs/pyjanitor/blob/8445dc00c42728df18b461d0f7df337b7cc96f18/.github/workflows/codecov.yml#L37

# PR Checklist

Please ensure that you have done the following:

1. [x] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [x] If you're not on the contributors list, add yourself to `AUTHORS.md`.

3. [x] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

# Automatic checks

There will be automatic checks run on the PR. These include:

- Building a preview of the docs on Netlify
- Automatically linting the code
- Making sure the code is documented
- Making sure that all tests are passed
- Making sure that code coverage doesn't go down.

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #1190, opened 2022-11-03
[INF/CI] Add `--cov-append` for `pytest`

Zeroto521:
# PR Description

> Another reason code coverage failed is that pytest doesn't add `--cov-append` option.
`--cov-append` can get a sum coverage. I'll add this option in the next PR.
First let us merge `codecov.yml` into `tests.yml`. Keep the same test logic for the dev branch or a PR.

_Originally posted by @Zeroto521 in https://github.com/pyjanitor-devs/pyjanitor/issues/1185#issuecomment-1296479926_

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           conda activate pyjanitor-dev
           python -m pip install -e .
-          pytest -m "${{ matrix.test-subset }}"
+          pytest -v -r a -n auto --color=yes --durations=0 --cov=janitor --cov-append --cov-report term-missing --cov-report xml --doctest-modules tests -m "${{ matrix.test-subset }}"
 
       # https://github.com/codecov/codecov-action
       - name: Upload code coverage
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           conda activate pyjanitor-dev
           python -m pip install -e .
-          pytest -v -r a -n auto --color=yes --durations=0 --cov=janitor --cov-append --cov-report term-missing --cov-report xml --doctest-modules tests -m "${{ matrix.test-subset }}"
+          pytest -v -r a -n auto --color=yes --durations=0 --cov=janitor --cov-append --cov-report term-missing --cov-report xml --doctest-modules --doctest-glob="janitor/*.py" --ignore-glob="nbconvert_config.py" tests -m "${{ matrix.test-subset }}"
 
       # https://github.com/codecov/codecov-action
       - name: Upload code coverage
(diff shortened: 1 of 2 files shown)


pytest.ini
```diff
@@ -1,6 +1,4 @@
 [pytest]
-# always check coverage of janitor.
-addopts = --cov=janitor --cov-report term-missing --cov-report xml --durations=0 --doctest-modules --doctest-glob="janitor/*.py" --ignore-glob="nbconvert_config.py" -v -n auto --color=yes
```
Zeroto521, line 3: I don't like `addopts`. That will force all running to use that configuration.
If I only want to do a single test like `tests/test_df_registration.py`, I do not need `pytest-xdist` to enable the multi-core feature.

.github/workflows/tests.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: |
           conda activate pyjanitor-dev
           python -m pip install -e .
-          pytest -m "${{ matrix.test-subset }}"
```
Zeroto521, line 48: Later we could update these to the Makefile.

Zeroto521: And we also could split normal test cases and doctest cases.
Currently, they are mixed up.

pytest.ini
```diff
@@ -1,6 +1,4 @@
 [pytest]
```
Zeroto521, line 1: `[pytest]` maybe could be update into `pyproject.toml`.

ericmjl: This sounds good. Would you prefer to sneak it in this PR or do it in a separate PR?

pytest.ini
ericmjl: I see. This sounds reasonable, though I think we do want to have sane defaults here; can you think of a way to enable:

1. Sane default that includes multi-core,
2. Flexibility to disable certain addopts,
3. Without excessive duplication of commands across files?

Multi-core testing will really speed up testing in the CI and locally.

pytest.ini
Zeroto521: I thought we better fix coverage problems first. So merge #1191 then merge this PR

pytest.ini
Zeroto521: We use addopts because of we are too lazy to key in long commands.
So the solution is Makefile. Set some common commands. Other conditions we could just key in pytest path/to/test_file.py

Zeroto521:
ready to be merged

samukweku:
@Zeroto521 any idea why codecov/patch is still running?

Zeroto521:
It may reopen and then reset causing that.
Let's merge the branch dev into this PR. See what happens.

Zeroto521:
Well, another weird thing is the color whoever @ me (@Zeroto521).
In the middle of this year, I have a long "busy" status.
After I canceled that status, the color is still there.

samukweku (approved):

ericmjl (approved):
Approving!

pytest.ini
Zeroto521: done at #1204

.github/workflows/tests.yml
Zeroto521: splitting doctests tests and other tests were done at #1205

PR #1194, opened 2022-11-08
Infra-updates

ericmjl:
This PR adds in infrastructure upgrades.

Firstly, we add recommended VSCode extensions into `devcontainer.json`. 

Secondly, we add in a workflow to automatically update PR branches as soon as the target branch is updated.


samukweku (approved):

ericmjl:
🚀 Deployed on https://deploy-preview-1194--pyjanitor.netlify.app

Zeroto521:
There is a simple option.
To enable 
image

ericmjl:
@Zeroto521 that's a great find! I think the automatic merging can mutually co-exist with the button to update branch, as it'll be a quality-of-life improvement. What do you think?

.github/workflows/auto-update.yml
```diff
@@ -0,0 +1,14 @@
...
+  # This will trigger on all pushes to all branches.
+  push: {}
+jobs:
+  autoupdate:
+    name: autoupdate
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```
Zeroto521: I'm okay with both of them.
- workflow auto merges dev into PR
- we set that option and let the user update the PR

I have used this action. Based on my experience, there are somethings we need to do.
- Default github token can't work to bot opened PR, see [details](https://github.com/marketplace/actions/auto-update#limitations). A solution is using a personal key to replace it.
- Not all PRs need to update. Use a label to trigger this workflow or other conditions. Need to set `PR_FILTER`, `PR_READY_STATE` and `PR_LABELS`.

ericmjl: Great comments, @Zeroto521! I took a look at the Auto Update [docs](https://github.com/marketplace/actions/auto-update) and I decided on the following configuration:

- Still do all PRs, b/c I don't think we are at the level of sophistication yet where we leverage labels on PRs, and
- Only do them on PRs that are "ready for review", as I've seen examples of us do PRs in a draft state first.

.github/workflows/auto-update.yml
```diff
@@ -0,0 +1,15 @@
+# This workflow automatically updates PR branches with latest changes on target branch.
+# See: https://github.com/marketplace/actions/auto-update
+name: autoupdate
+on:
+  # This will trigger on all pushes to all branches.
+  push: {}
```
Zeroto521: `push: {}` means any branch.
I prefer to have specific the branch via:

```yaml
on:
  push:
    branches: [dev]
```

Zeroto521 (approved):

ericmjl:
@Zeroto521 when the CI system passes, please do the honors of merging!

PR #1207, opened 2022-11-20
[ENH] Minor fix for `conditional_join`

samukweku:
# PR description:
- fixes scenarios where an array/pd.Series/pd.Index is passed to `df_columns/right_columns`; 
- Add more comments in code section 
- Improve `_range_indices` to break early where possible


ericmjl:
🚀 Deployed on https://deploy-preview-1207--pyjanitor.netlify.app

thatlittleboy (approved):
Thanks Sam, I'll approve.

I can see the logic in doing so, but just curious if you've done any preliminary verification for the benefit of breaking early. It *feels* to me like there would be excessive memory reallocations as a result of L1181-1182. (Naively speaking, of course :) I'm not that well-versed in numpy under-the-hood optimizations)

simon-cowell-agt-thumbsup

samukweku:
@thatlittleboy fair point, I'll have a look and verify

samukweku:
memory doesn't seem to spike - i may be misinterpreting the output though ...: 

image

samukweku:
thanks to @thatlittleboy i think going via the numpy broadcasting route is the way to go; the numba option is always available for users

samukweku:
did some more tests, thanks to prodding from @thatlittleboy (this is why i love code reviews 😄), and because of the early breaks, on some really large data (d` is 20,000 rows, d2 is  100,000 rows) the python for loop is about 3x faster than np.func.outer, and offers more memory efficiency:

image


the first run - cell 12 - uses np.less.outer, while the second cell - cell 13 - uses the for loop. Of course, the `_range_indices` relies on the fact that there are little or no overlaps in the right side for it to be effective. The `numba` option is more robust though, and can handle overlaps pretty well.

This was just an adhoc test, there might be cases where the logic falls flat on its face. So far so good, I think I'll leave the for loop. As always, I am open to feedback, reviews and improvements. Thanks @thatlittleboy.   I sure do know how to write an epistle, right?

ericmjl (approved):

thatlittleboy:
Wow, that's a pretty sizeable improvement. Once again proving that one can't go with intuition with these things. 😅

Maybe we could build a test around this, similar to coverage, to give us a before and after profiling report (runtime and mem usage stats), in the future.
Definitely not here though, gotta stay true to the title of the PR :) 

I'll merge since Eric has also approved~

ericmjl:
Thank you for taking the initiative and merging, @thatlittleboy! And thank you also for the awesome PR and work done here, @samukweku!

PR #1219, opened 2022-12-02
[ENH] minor fixes

samukweku:
# PR Description

- minor fixes for  `drop_constant_columns` and `get_dupes`
- improve performance for `select` when all entries are scalars and are the same dtype
- `move` is more flexible with the `select_columns` syntax - multiple columns/rows can be moved at once
- avoid mutation in `collapse_levels`
- `impute` now supports multiple columns, making it easy to deprecate `fill_empty`
-  fix deprecation warning for `np.bool8`
- simplify column selection for dtypes (is_numeric_dtype, ...)

- @ericmjl

### DIFF
--- a/janitor/functions/move.py
+++ b/janitor/functions/move.py
@@ -68,8 +68,8 @@ def move(
     :returns: The dataframe with the Series moved.
     :raises ValueError: If `axis` is not `0` or `1`.
     :raises ValueError: If `position` is not `before` or `after`.
-    :raises ValueError: If  `source` row or column is not in dataframe.
-    :raises ValueError: If `target` row or column is not in dataframe.
+    :raises KeyError: If  `source` or `target` row or column
+        is not in dataframe.
     """
     if axis not in [0, 1]:
         raise ValueError(f"Invalid axis '{axis}'. Can only be 0 or 1.")
@@ -79,40 +79,23 @@ def move(
             f"Invalid position '{position}'. Can only be 'before' or 'after'."
         )
 
-    df = df.copy()
-    if axis == 0:
-        names = list(df.index)
-
-        if source not in names:
-            raise ValueError(f"Source row '{source}' not in dataframe.")
-
-        if target not in names:
-            raise ValueError(f"Target row '{target}' not in dataframe.")
-
-        names.remove(source)
-        pos = names.index(target)
-
-        if position == "after":
-            pos += 1
-        names.insert(pos, source)
-
-        df = df.loc[names, :]
+    if not axis:
+        names = df.index.tolist()
     else:
-        names = list(df.columns)
+        names = df.columns.tolist()
+    mapping = {0: "row", 1: "column"}
 
-        if source not in names:
-            raise ValueError(f"Source column '{source}' not in dataframe.")
+    if source not in names:
+        raise KeyError(f"Source {mapping[axis]} '{source}' not in dataframe.")
 
-        if target not in names:
-            raise ValueError(f"Target column '{target}' not in dataframe.")
+    if target not in names:
+        raise KeyError(f"Target {mapping[axis]} '{target}' not in dataframe.")
 
-        names.remove(source)
-        pos = names.index(target)
+    names.remove(source)
+    pos = names.index(target)
 
-        if position == "after":
-            pos += 1
-        names.insert(pos, source)
+    if position == "after":
+        pos += 1
+    names.insert(pos, source)
 
-        df = df.loc[:, names]
-
-    return df
+    return df.loc(axis=axis)[names]
--- a/janitor/functions/move.py
+++ b/janitor/functions/move.py
@@ -75,7 +75,7 @@ def move(
     :param source: Columns or rows to move.
     :param target: Columns or rows to move adjacent to.
         If `None` and `position == 'before'`, `source`
-        is moved to the beginning; if `position == 'after`,
+        is moved to the beginning; if `position == 'after'`,
         `source` is moved to the end.
     :param position: Specifies the destination of the columns/rows.
         Values can be either `before` or `after`; defaults to `before`.
--- a/janitor/functions/move.py
+++ b/janitor/functions/move.py
@@ -1,21 +1,25 @@
 """Implementation of move."""
 import pandas_flavor as pf
 import pandas as pd
+import numpy as np
 
-from typing import Union
+from typing import Any
+from janitor.functions.utils import _select_index, _index_converter
 
 
 @pf.register_dataframe_method
 def move(
     df: pd.DataFrame,
-    source: Union[int, str],
-    target: Union[int, str],
+    source: Any,
+    target: Any = None,
     position: str = "before",
     axis: int = 0,
 ) -> pd.DataFrame:
     """
-    Moves a column or row to a position adjacent to another column or row in
-    the dataframe.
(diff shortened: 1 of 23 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1219--pyjanitor.netlify.app

tests/functions/test_impute.py
```diff
@@ -32,12 +32,24 @@ def test_impute_statistical(missingdata_df, statistic, expected):
 @pytest.mark.functions
```
thatlittleboy, line 42: Could you add a test when `column_names` is a list / tuple of column names rather than just a column name?

janitor/functions/move.py
```diff
@@ -1,21 +1,25 @@
...
     position: str = "before",
     axis: int = 0,
 ) -> pd.DataFrame:
     """
-    Moves a column or row to a position adjacent to another column or row in
-    the dataframe.
+    Changes rows or columns positions in the dataframe. It uses the
+    [`select_columns`][janitor.functions.select.select_columns] or
+    [`select_rows`][janitor.functions.select.select_rows] syntax,
+    making it easy to move blocks of rows or columns at once.
```
thatlittleboy, line 22: Nice, can we modify just one of the examples below to demonstrate movement of multiple columns? (Presumably, it's by passing in a list of column names)

Secondly, I'm not sure if this is the intended behaviour? (When source and target are BOTH lists)

```pycon
>>> import janitor as jn, numpy as np, pandas as pd
>>> df = pd.DataFrame(np.random.random(size=(5, 10)), columns=list("abcdefghij"))
>>> df.move(source=["a", "c"], target=["f", "e"], axis=1)

          b         d         e         a         c         f         g         h         i         j
0  0.205516  0.522320  0.316623  0.071786  0.603323  0.424672  0.779201  0.519310  0.154003  0.138308
1  0.899164  0.721393  0.668475  0.468555  0.865373  0.210095  0.006999  0.907745  0.155087  0.089782
2  0.352662  0.195532  0.439537  0.149423  0.777204  0.696971  0.158383  0.794357  0.708664  0.003781
3  0.746061  0.873447  0.914278  0.790982  0.304805  0.811771  0.568839  0.370157  0.658724  0.205267
4  0.688929  0.348658  0.883594  0.043638  0.513286  0.363843  0.287933  0.776863  0.366071  0.885217
```
But given this is a "nonsensical" input, I guess it's not that big a deal. But can we at least add a test relating to this, so that we can track when this behaviour changes/regresses?

tests/functions/test_encode_categorical.py
```diff
@@ -62,10 +62,10 @@ def test_encode_categorical_missing_columns(df):
 @settings(deadline=None)
 def test_encode_categorical_invalid_input(df):
     """
-    Raise ValueError for wrong input type
+    Raise KeyError for wrong input type
     for `column_names`.
     """
-    with pytest.raises(ValueError):
+    with pytest.raises(KeyError):
         df.encode_categorical(1)
```
thatlittleboy: Not quite sure I understand the point of this test anymore, since if there was a datafame with column name (1, the int), it would work fine. So essentially this test is the same as the `test_encode_categorical_missing_column()` test above.

tests/functions/test_encode_categorical.py
```diff
@@ -74,10 +74,10 @@ def test_encode_categorical_invalid_input(df):
 @settings(deadline=None)
 def test_encode_categorical_invalid_input_2(df):
     """
-    Raise TypeError for wrong input type
+    Raise KeyError for wrong input type
     for `column_names`.
     """
-    with pytest.raises(TypeError):
+    with pytest.raises(KeyError):
         df.encode_categorical({"names"})
```
thatlittleboy: any way we could support this instead of letting it throw an error? seems like we should be able to accept any generic iterable (like sets), at least in this case.

mkdocs.yml
```diff
@@ -31,12 +31,15 @@ nav:
...
+      # - Utils: api/utils.md
       - XArray: api/xarray.md
   - Changelog: CHANGELOG.md
   - Authors: AUTHORS.md
 
 plugins:
+  - exclude:
+      glob:
+        - api/utils.md
   - search
```
thatlittleboy: Rather than introducing an additional dependency (mkdocs-exclude), how about we just remove the `api/utils.md` file entirely? Would that work?

thatlittleboy (changes requested):

mkdocs.yml
samukweku: so we move relevant files out of utils.md? dont know if that would work ... open to ideas on how to do it ... cos i dont know

janitor/functions/move.py
samukweku: for a list it will move to the front of the first column in the list, or the back of the last column (if position is after) - if the list is ordered as it is in the dataframe, the output makes sense a bit: 

```py
df.move(source=["a", "c"], target=["e", "f"], axis=1)
Out[19]: 
          b         d         a         c         e         f         g         h         i         j
0  0.501875  0.133830  0.010374  0.495773  0.142111  0.218559  0.418508  0.248101  0.084060  0.345499
1  0.878559  0.038748  0.166776  0.950964  0.699107  0.572760  0.898007  0.666899  0.547838  0.702427
2  0.694439  0.465668  0.386477  0.824870  0.940524  0.803740  0.983359  0.169636  0.521085  0.945517
3  0.860832  0.185542  0.649540  0.970846  0.243116  0.873401  0.906277  0.758278  0.346902  0.463535
4  0.496282  0.435555  0.610538  0.477974  0.870408  0.187359  0.687674  0.703529  0.925200  0.039455
```

There is a `test_move_reorder_columns` for columns - does that suffice?

mkdocs.yml
thatlittleboy: Just to confirm: the point of your proposed change here is so that utils doesn't show up in the "API reference" section right?

If so, you can just 
- delete the file `mkdocs/api/utils.md`
- and remove the `api/utils.md` line from the `nav:` section in `mkdocs.yml`

I don't think there's a need for the `exclude` plugin.

janitor/functions/move.py
thatlittleboy: Not quite, `test_move_reorder_columns` only tests if `source` is a list with `target` None.

I was referring to the case when both source and target are lists. We can add a test case when `source=["j", "a"]` and `target=["c", "e"]`, and check column output should be `['b', 'j', 'a', 'c', 'd', 'e', 'f', 'g', 'h', 'i']` for "before" and `['b', 'c', 'd', 'e', 'j', 'a', 'f', 'g', 'h', 'i']` for "after".

samukweku:
apologies, my rebasing included files that shouldnt be in there ... i def have to go back to basics for git again

mkdocs.yml
samukweku: we probably should move some of the relevant functions out of `utils`, like `unionize_dataframe_categories` into its own section, or a section of useful tools. open to suggestions on how that should look like

samukweku: ah now i remember why i used `exclude` - mkdocs builds everything, i then have to exclude the ones I dont want

thatlittleboy (approved):
asian-man-cubicle-thumbsup

ericmjl (approved):

Issue #1032, opened 2022-02-26
[DOC/INF] Migrate mkdocs documentation style from sphinx to google


thatlittleboy:
# The current state of affairs

1. The migration of pyjanitor's documentation from `Sphinx` to `mkdocs` happened late last year (#894).
    And with `mkdocstrings`, we’re [currently using the “restructured-text” docstring-style](https://github.com/pyjanitor-devs/pyjanitor/blob/eabc98858b56474d919e8f089dddb57c4b725ec5/mkdocs.yml#L49) option (which is not actually reST but rather "sphinx"-style docstrings, with `:param :`, `:returns:` etc.).
2. As of time of writing, there is an ongoing sprint on incorporating MWEs in all of our (public) APIs. (#972).
    So, having proper and nicely-formatted example blocks in our documentation is becoming more important moving forward.

# The issue(s)

### Upstream support is poor

From my interactions with the core-dev of `mkdocstrings`, and having gotten to know the `mkdocs` ecosystem better over the past week or so, I’ve realized that the restructured-text/sphinx docstring style **is not as well-supported** as the other styles (namely, google / numpy) by `mkdocstrings`.
For example, both the legacy python handler and new experimental python handler does not properly support the “examples” section in sphinx style at the moment (see [here-legacy](https://mkdocstrings.github.io/python-legacy/usage/#sections_1), and [here-new](https://mkdocstrings.github.io/griffe/docstrings/#section-support)).

And the new option to remove doctest flags (recently added to [pytkdocs](https://github.com/mkdocstrings/pytkdocs/pull/134) and [griffe](https://github.com/mkdocstrings/griffe/pull/48)) has only been implemented for Google/Numpy styles, but NOT for the reST / sphinx styles. --- as a direct consequence of the docstring collectors not supporting "examples" section mentioned above.

### Broken / Incomplete functionality

On the `pyjanitor` front: The following points could serve as their own individual issues/tickets on their own right, but just as an overview of what I’ve observed:
- doctest flags are visible in the rendered HTML (#1000);
- Extraneous space after the code block if there are no text after. (Note the space above the "Parameters" section)
   <img width="500" alt="examples-current" src="https://user-images.githubusercontent.com/30731072/155834151-1ccac03e-d548-4f3c-b170-276e4553b2b8.png">
[...]

PR #1231, opened 2023-01-22
[INF] update some mkdocs infrastructure

thatlittleboy:
# PR Description

Couple of minor infrastructure changes:

- I've deleted most of the linting tools in `environment-dev.yml` to reduce the bloat (esp for our CIs), and to promote the use of `pre-commit` (which sets up its own isolated environments for each tool). Two exceptions:
    - `black`: since the vscode container uses the black binary to format on save
    - `isort`: this isn't in pre-commit config yet, to be fixed in the future
- Since we don't pin `mkdocs` and friends, there're some new compatibility changes with v0.9 mkdocs-material and mkdocs v1.4. E.g. the `watch` option is recommended to use the global mkdocs one. `pymdownx` navigation extension names were wrong... etc.
- Added some more mkdocs-material features which I think will help the navigation experience
    - [navigation.top](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#back-to-top-button)
    - [toc.follow](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#anchor-following)
    - [content.code.copy](https://squidfunk.github.io/mkdocs-material/upgrade/#contentcodecopy) -- though I'm not fully convinced we need this, since alot of our code blocks are not "copy-friendly". But we can test run this for a bit to see how it goes.
- I moved the `Development Guide` section to below `API reference`, since the latter is most likely what our users want to see (first).

# PR Checklist

1. [x] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/mkdocs/css/apidocs.css
+++ b/mkdocs/css/apidocs.css
@@ -1,14 +1,46 @@
 /* https://mkdocstrings.github.io/theming/#css-classes */
-.doc-property { border-radius: 15px; padding: 0 5px; }
-.doc-property-special { background-color: blue; color: white; }
-.doc-property-private { background-color: red; color: white; }
-.doc-property-property { background-color: green; color: white; }
-.doc-property-read-only { background-color: yellow; color: black; }
+.doc-property {
+  border-radius: 15px;
+  padding: 0 5px;
+}
+.doc-property-special {
+  background-color: blue;
+  color: white;
+}
+.doc-property-private {
+  background-color: red;
+  color: white;
+}
+.doc-property-property {
+  background-color: green;
+  color: white;
+}
+.doc-property-read-only {
+  background-color: yellow;
+  color: black;
+}
 
 /* https://mkdocstrings.github.io/handlers/python/#recommended-style-material */
 /* Indentation. */
 div.doc-contents:not(.first) {
-    padding-left: 25px;
-    border-left: 4px solid rgba(230, 230, 230);
-    margin-bottom: 80px;
-  }
+  padding-left: 25px;
+  border-left: 4px solid rgba(230, 230, 230);
+  margin-bottom: 80px;
+}
+
+/* add a keyboard shortcut icon for search bar,
+ * https://github.com/squidfunk/mkdocs-material/issues/2574#issuecomment-821979698
+ */
+[data-md-toggle="search"]:not(:checked) ~ .md-header .md-search__form::after {
+  position: absolute;
+  top: 0.3rem;
+  right: 0.3rem;
+  display: block;
+  padding: 0.1rem 0.4rem;
+  color: var(--md-default-bg-color--lighter);
+  font-weight: bold;
+  font-size: 0.8rem;
+  border: 0.05rem solid var(--md-default-bg-color--lighter);
+  border-radius: 0.1rem;
+  content: "/";
+}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,11 @@ repos:
   hooks:
   - id: black
     args: [--config, pyproject.toml]
+# - repo: https://github.com/pycqa/isort
+#   rev: 5.11.2
+#   hooks:
+#     - id: isort
+#       name: isort (python)
 - repo: https://github.com/econchick/interrogate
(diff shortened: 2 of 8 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1231--pyjanitor.netlify.app

samukweku (approved):
any ideas on the mkdocs-legacy? ok to remove and just have the new mkdocs?

thatlittleboy:
> any ideas on the mkdocs-legacy? ok to remove and just have the new mkdocs?

@samukweku 

Yea, I was thinking about this recently as well, since `mkdocstrings` recently just got updated to v0.20. I don't think it's okay to remove it just yet, it hinges on #1032 being resolved -- because the new python parser that mkdocstrings uses (griffe) doesn't support the sphinx style that well.

So I'm imagining that we'll need to:
- update all doc styles to google style (it's the most feature complete one)
- update mkdocstrings to v0.20+, along with relevant changes to `mkdocs.yml` -- selection/rendering options are now in one single list of options, etc.
- and remove the version pinning (jinja2 to use 3.10+, etc.)

---

**Correction**: hmm nvm, ignore me. It might be possible to just do an upgrade first (while still using the legacy parser), I'll investigate and report back.

thatlittleboy:
> > any ideas on the mkdocs-legacy? ok to remove and just have the new mkdocs?
> 
> @samukweku
> 
> Yea, I was thinking about this recently as well, since `mkdocstrings` recently just got updated to v0.20. I don't think it's okay to remove it just yet, it hinges on #1032 being resolved -- because the new python parser that mkdocstrings uses (griffe) doesn't support the sphinx style that well.
> 
> So I'm imagining that we'll need to:
> 
>     * update all doc styles to google style (it's the most feature complete one)
> 
>     * update mkdocstrings to v0.20+, along with relevant changes to `mkdocs.yml` -- selection/rendering options are now in one single list of options, etc.
> 
>     * and remove the version pinning (jinja2 to use 3.10+, etc.)
> 
> 
> **Correction**: hmm nvm, ignore me. It might be possible to just do an upgrade first (while still using the legacy parser), I'll investigate and report back.

We can remove the pin on mkdocstrings to use v0.19+ (not quite, need an [upstream bug](https://github.com/mkdocstrings/python-legacy/issues/4) to be fixed first), but I wouldn't make the jump to drop the "mkdocstrings-python-legacy" handler just yet. I rather we move to Google style along (#1032) with the upgrade to the new "mkdocstrings-python" handler together as a whole.

I'll create a separate PR to upgrade our `mkdocstrings` once the aforementioned bug is fixed.

thatlittleboy:
Just realized there's been multiple related PRs to this current one (some overlap), #1133, #1147  . I'll like your opinion on the changes I'm making, if possible @ericmjl @Zeroto521  🙏🏽

mkdocs/css/apidocs.css
```diff
@@ -1,14 +1,46 @@
...
+  right: 0.3rem;
+  display: block;
+  padding: 0.1rem 0.4rem;
+  color: var(--md-default-bg-color--lighter);
+  font-weight: bold;
+  font-size: 0.8rem;
+  border: 0.05rem solid var(--md-default-bg-color--lighter);
+  border-radius: 0.1rem;
+  content: "/";
+}
```
thatlittleboy, line 46: image

**For review purposes**: I only added this block of CSS to add a keyboard shortcut icon in the search bar. The rest of the diff in this `apidocs.css` file is a result of my autoformatter.

ericmjl: Thank you for leaving this hint here! Without it, I would have been confused as to where the big diff came from 😅.

ericmjl:
@thatlittleboy thank you for handling this PR!

I've reviewed the changes and I'm good to go as well.

Regarding docstring style, I'm also happy to make a move to Google-style docstrings now that AI coding tools are more generally available, as they can help with the migration. I have paid for Tabnine, but we can also edit ChatGPT's output. Without them, reformatting would be a chore.

ericmjl (approved):

PR #1235, opened 2023-01-29
[DOC] Migrate to google docstring style

### LINKED ISSUE #1032: [DOC/INF] Migrate mkdocs documentation style from sphinx to google

thatlittleboy:
# PR Description

- Migrate all "visible" (i.e., rendered into mkdocs) docstrings to Google style. For better compatibility with `mkdocstrings`.
- With this change, we can finally drop the dependency on the legacy python handler and use the newer python handler with `mkdocstrings`.
- Added a guidelines page on docstring style for future developers / contributors to follow.
- Update changelog.

**This PR relates to #1032.**

(There are other ideas raised in that issue that I would like to fix as well, hence not closing, but that's for a future PR)

# PR Checklist

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1235--pyjanitor.netlify.app

thatlittleboy:
`darglint` is struggling with checking all of these files, and `pre-commit-CI` is timing out after 180s. And I can't find a way to extend this timeout.

I've run the pre-commit hooks locally, from 088b3f3 (dev) to 61dc66d (current HEAD before the changelog commit)

image

As for using ruff to replace darglint, it's not currently implemented yet so that's not an option for now.

---

But anyway, in better news, I think I've migrated all the public docstrings to Google style already (no guarantees on the private ones, though, I think I got most of them as well). PR is now ready for review! 🎉 

**Key review points**:

- updated the dependencies and relevant `mkdocs.yml` config
- look through the rendered documentation and see if there's anything that looks off


Thanks!

ericmjl:
@thatlittleboy I spot-checked the Functions page, and it looks great!

This is quite a heroic effort 🎉

Thank you for handling this migration!

I think we shouldn't worry about pre-commit timing out unless it gives us a ton of problems in the future. In that case, we can always run our own pre-commit Action.

ericmjl (approved):

ericmjl:
@samukweku and @Zeroto521 - the PR diff is large but conceptually, it's a small change. Between the two of you, whoever sees my comment first can just merge straight away if you don't want to spot-check things yourself. That way we'll get the updated docs into `dev` asap!

samukweku (approved):

PR #1237, opened 2023-02-03
[DOC] Prevent selection of `>>>` and output in code blocks

### LINKED ISSUE #1032: [DOC/INF] Migrate mkdocs documentation style from sphinx to google

thatlittleboy:
# PR Description

Final few finishing touches relating to the plan to improve our docs, laid out in #1032.

- Prevent selection of `>>>` and output in code blocks using https://mkdocstrings.github.io/recipes/#prevent-selection-of-prompts-and-output-in-python-code-blocks
    - Note that this entails dropping `highlight.js`, and thus the Nord css theme. For now, just using the default colour scheme that comes with Mkdocs Material.
- Fixed a broken mkdocs theme setting (wrong accent color)
- Squished the last few sphinx-style docstyles remaining in some of our private functions.

**Before**

image

**After**

image

**This PR resolves #1032 .**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 -   [INF] Fixed issue with missing PyPI README. PR #1216 @thatlittleboy
 -   [INF] Update some `mkdocs` compatibility code. PR #1231 @thatlittleboy
 -   [INF] Migrated docstring style from Sphinx to Google for better compatibility with `mkdocstrings`. PR #1235 @thatlittleboy
+-   [INF] Prevent selection of chevrons (`>>>`) and outputs in Example code blocks. PR #1237 @thatlittleboy
 -   [DEPR] Add deprecation warnings for `process_text`, `rename_column`, `rename_columns`, `filter_on`, `remove_columns`, `fill_direction`. #1045 @samukweku
 -   [ENH] `pivot_longer` now supports named groups where `names_pattern` is a regular expression. A dictionary can now be passed to `names_pattern`, and is internally evaluated as a list/tuple of regular expressions. Issue #1209 @samukweku
 
--- a/janitor/functions/clean_names.py
+++ b/janitor/functions/clean_names.py
@@ -179,18 +179,20 @@ def _strip_underscores(
 
     Underscores can be stripped from the beginning, end or both.
 
-    Example usage:
+    Examples:
+        ```python
+        df = _strip_underscores(df, strip_underscores='left')
+        ```
 
-    ```
-    df = _strip_underscores(df, strip_underscores='left')
-    ```
+    Args:
+        df: The pandas DataFrame object.
+        strip_underscores: Removes the outer underscores from all
+            column names. Default `None` keeps outer underscores. Values can be
+            either `'left'`, `'right'` or `'both'` or the respective shorthand
+            `'l'`, `'r'` and `True`.
 
-    :param df: The pandas DataFrame object.
-    :param strip_underscores: (optional) Removes the outer underscores from all
-        column names. Default `None` keeps outer underscores. Values can be
-        either `'left'`, `'right'` or `'both'` or the respective shorthand
-        `'l'`, `'r'` and `True`.
-    :returns: A pandas DataFrame with underscores removed.
+    Returns:
+        A pandas DataFrame with underscores removed.
     """
     df = df.rename(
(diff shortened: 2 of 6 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1237--pyjanitor.netlify.app

samukweku (approved):

ericmjl (approved):
Approving! @samukweku or @Zeroto521 would either of you like to do the honors of merging the pull request?

PR #1242, opened 2023-02-16
[DOC] Add more examples to `pivot_longer`

samukweku:
- Add more examples to highlight flexible column selection in `pivot_longer`

### DIFF
--- a/janitor/functions/pivot.py
+++ b/janitor/functions/pivot.py
@@ -54,6 +54,7 @@ def pivot_longer(
     [`select_columns`][janitor.functions.select.select_columns] syntax.
 
     Examples:
+
         >>> import pandas as pd
         >>> import janitor
         >>> df = pd.DataFrame(
@@ -70,7 +71,7 @@ def pivot_longer(
         0           5.1          3.5           1.4          0.2     setosa
         1           5.9          3.0           5.1          1.8  virginica
 
-        Replicate pandas' melt:
+    Replicate pandas' melt:
 
         >>> df.pivot_longer(index = 'Species')
              Species      variable  value
@@ -83,8 +84,8 @@ def pivot_longer(
         6     setosa   Petal.Width    0.2
         7  virginica   Petal.Width    1.8
 
-        Convenient, flexible column selection in the `index` via the
-        [`select_columns`][janitor.functions.select.select_columns] syntax:
+    Convenient, flexible column selection in the `index` via the
+    [`select_columns`][janitor.functions.select.select_columns] syntax:
 
         >>> from pandas.api.types import is_string_dtype
         >>> df.pivot_longer(index = is_string_dtype)
@@ -98,7 +99,7 @@ def pivot_longer(
         6     setosa   Petal.Width    0.2
         7  virginica   Petal.Width    1.8
 
-        Split the column labels into parts:
+    Split the column labels into parts:
 
         >>> df.pivot_longer(
         ...     index = 'Species',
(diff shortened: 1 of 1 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1242--pyjanitor.netlify.app

samukweku:
@thatlittleboy pls lemme know if i broke anything with the new google style format :grin:

janitor/functions/pivot.py
```diff
@@ -54,6 +54,7 @@ def pivot_longer(
     [`select_columns`][janitor.functions.select.select_columns] syntax.
 
     Examples:
+
```
thatlittleboy: In general, we don't need this newline anymore after a section heading with the new parser. The entire block just needs to be indented once, whether it's a code-block or a text block within Examples. So it can be like

```
Examples:
    This is some text

    >>> x = 1
```

or 

```
Examples:
    >>> x = 1
```

---

This is exactly the same and consistent with the Args/Params section, we write

```
Args:
    x: An input param
```

And not

```
Args:

    x: An input param
```

janitor/functions/pivot.py
```diff
@@ -70,7 +71,7 @@ def pivot_longer(
         0           5.1          3.5           1.4          0.2     setosa
         1           5.9          3.0           5.1          1.8  virginica
 
-        Replicate pandas' melt:
+    Replicate pandas' melt:
```
thatlittleboy: Same here, the **entire** Examples block should be indented on the same level. So the indentation of your `Replicate pandas' melt` line is off.

```python
Examples:
    This is a text block.

     >>> x = 1

     This is another text block.
```

The way to understand this is, we denote new sub-blocks (paragraphs) within the Examples section with new lines, not with indentations.

I think there's no visual difference in the output (whether we indent "Replicate pandas' melt" text or not), just a difference in the parsed AST. so this is just a nitpick. :)

janitor/functions/pivot.py
```diff
@@ -83,7 +84,22 @@ def pivot_longer(
         6     setosa   Petal.Width    0.2
         7  virginica   Petal.Width    1.8
 
-        Split the column labels into parts:
+    Convenient, flexible column selection in the `index` via the
+    [`select_columns`][janitor.functions.select.select_columns] syntax:
```
thatlittleboy: Indentation here as well -- indent once to the right.

Same goes for all the diffs that u introduced below.

thatlittleboy (changes requested):
Yea, it seems broken 😅 Have a look at my comments, see if it makes sense to you.

janitor/functions/pivot.py
samukweku: @thatlittleboy I feel I'm still missing something from the styling ... local preview still failing

thatlittleboy: @samukweku Local preview? Can you take a screenshot to show what you mean? The netlify documentation seems to look okay to me

samukweku: netlify does not show the latest ... it shows what is already merged into the repo. not sure why that is 


image


Notice how the selection takes the `....` as well. I liked the initial setup where the `...` is not copied. Weirdly enuf this happens only in this PR, which makes me wonder where else I had touched to cause the break :man_shrugging:

thatlittleboy: Huh, weird. This netlify link? https://deploy-preview-1242--pyjanitor.netlify.app/

It looks fine for me (the `...` is not selected, and the example u added in this PR shows up just fine):

image

---

maybe it's just your browser cache?

thatlittleboy (approved):
unwilling-cat-low-five-shipit

ericmjl (approved):

ericmjl:
Merged! Thank you, @samukweku!

Issue #1236, opened 2023-01-30
[INF] darglint is archived -> move to ruff?


thatlittleboy:
While working on #1235, I've experienced darglint to be really slow (takes minutes to run for the `functions/pivot.py` file), and also the project has been archived as of Dec 16 2022.

https://github.com/terrencepreilly/darglint/

We should transition to using `ruff`, which also provides darglint rules.

In which case, perhaps we can also drop `flake8` in favour of ruff.

EDIT:

oops, I misremembered. darglint is not covered by ruff yet, but it's in the works. https://github.com/charliermarsh/ruff/issues/458

ericmjl:
@thatlittleboy I am going to switch our infra over to ruff. Thanks for raising this issue, and oh my, I just realized I've really neglected our shared project for a long time. (Work has been craaaazy.)

PR #1246, opened 2023-02-18
[INF] Avoid timeouts in pre-commit ci/cd due to darglint

thatlittleboy:
Just a tiny PR to turn off darglint in the pre-commit CI workflow for now --- as it's consistently taking >3min to run and timing out ---, and replacing it with GH actions.  
It is currently hindering the other pre-commit checks (black) from being applied/auto-fixed in PRs as a result as well.

Locally, one would still use pre-commit to check darglint as per normal.

The plan is to use #1236 to replace darglint.

Also, `black` has adopted a new style with the new 2023 release, so this PR also fixes that to conform our code to the new style.

### DIFF
--- a/.github/workflows/darglint-checks.yml
+++ b/.github/workflows/darglint-checks.yml
@@ -1,0 +1,42 @@
+# Temporarily run darglint using github actions rather than pre-commit CI because it is consistently
+# more than 5min to run. To be removed once ruff supports darglint rules.
+name: Darglint checks
+
+on:
+  # only run darglint checks when src-code changes
+  push:
+    branches:
+      - dev
+    paths:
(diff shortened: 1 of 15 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1246--pyjanitor.netlify.app

samukweku (approved):

asottile:
hello -- please merge this ASAP as it's significantly impacting queue at the moment

thatlittleboy:
@ericmjl I'll make the executive decision to merge this PR in first since it's affecting others; let's address further concerns in a separate PR, hope you don't mind!

thatlittleboy:
@asottile done

Issue #1250, opened 2023-02-26
avoid explicit copy ?


samukweku:
See if it is possible to avoid copy and get more performance  - 

- `complete`
- `pivot_longer`

samukweku:
For `complete`, there is a `pd.merge` call - pandas makes a copy within `pd.merge`; as such, there is no need for an explicit copy again.

For `pivot_longer`, the transformations are done on the columns, and Pandas creates a new Index altogether - it is safe to create a view on the original dataframe since it won't be tampered with. 

Avoid copying if it isn't required, or if some other pandas process already takes care of that 😁

ericmjl:
@samukweku I think we can close this one, perhaps? Definitely want to thank you for handling the two PRs!

samukweku:
Yes @ericmjl it is ok to close

PR #1252, opened 2023-02-28
[ENH] Avoid explicit copy in `pivot_longer`

samukweku:
# PR Description

- Avoid explicit copy `df.copy`, since copy is already invoked when selecting the columns (line 866)
- Minor change, perf not affected

**This PR relates to #1250 .**

- @ericmjl

### DIFF
--- a/janitor/functions/pivot.py
+++ b/janitor/functions/pivot.py
@@ -856,13 +856,9 @@ def _computations_pivot_longer(
     if len(column_names) != len(set(column_names)):
         column_names = pd.unique(column_names)
 
-    # creates a new object - essentially a copy
-    # for large dataframes, it would be more efficient
-    # to build a new dataframe from a dictionary comprehension
-    # with copy=False
-    # but that could have adverse effects downstream
-    # if the user executes a computation that mutates the
-    # original dataframe - how to act as a copy without being a copy?
+    # this is the reason why there is no explicit copy (df.copy())
+    # at the very beginning for `pivot_longer`,
+    # because `.loc` makes a copy of the dataframe
     out = df.loc[:, column_names]
 
     if all((names_pattern is None, names_sep is None)):
--- a/janitor/functions/pivot.py
+++ b/janitor/functions/pivot.py
@@ -858,7 +858,7 @@ def _computations_pivot_longer(
 
     # creates a new object - essentially a copy
     # for large dataframes, it would be more efficient
-    # to build a new dataframes from a dictionary comprehension
+    # to build a new dataframe from a dictionary comprehension
     # with copy=False
     # but that could have adverse effects downstream
     # if the user executes a computation that mutates the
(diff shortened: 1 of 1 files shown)


ericmjl:
🚀 Deployed on https://deploy-preview-1252--pyjanitor.netlify.app

janitor/functions/pivot.py
```diff
@@ -852,11 +855,19 @@ def _computations_pivot_longer(
...
-    df = df.loc[:, column_names]
+
+    # creates a new object - essentially a copy
+    # for large dataframes, it would be more efficient
+    # to build a new dataframe from a dictionary comprehension
+    # with copy=False
+    # but that could have adverse effects downstream
+    # if the user executes a computation that mutates the
+    # original dataframe - how to act as a copy without being a copy?
+    out = df.loc[:, column_names]
```
thatlittleboy: Sorry, bear with me: this "creates a new object" comment, is that in relation to the fact that `.loc[...]` does a copy or in relation to that fact that you renamed it `out`?

And as to your question at the end, would it eventually be resolved by the new CoW that pandas implemented since v1.5 (opt-in, but still.)

samukweku: my comment is not good enuf - I meant there is no need for an explicit copy at the start : `df = df.copy()` since on line 866, `loc` when selecting columns creates a new object - in essence a copy is made of the original dataframe. 
you are right - eventually CoW should take care of this

samukweku: any suggestions on how to rephrase the comments? @thatlittleboy

ericmjl: Let me see, @samukweku!

```
# No need to copy at the beginning because `.loc[...]` will create a new object for us.
```

Making sure we don't mutate the original dataframe object is something I think we've strived for. There should be an issue discussing this point. Therefore, I think this line is sufficient and we don't have to elaborate further.

Regarding doing a `.copy()` at the beginning of a function - we've used it as a pattern in the past, so putting a comment here is a good idea since that's a departure from a (previously) established pattern. Small side question: should we (1) make the pattern more explicit and document departures, or (2) abandon the pattern and rely on knowledge of `pandas` API to know when copies are made?

samukweku: Knowledge of pandas here helped. If we can delegate the copy to pandas early great. If there is a chance that any of our early operations will mutate the dataframe, then we should make an explicit copy. If we are doing an operation on only the columns (like collapse_levels), then we only need a slice, since only the columns/index is mutated (and pandas creates a new index)

samukweku: There are also some scenarios where making an explicit copy isn't needed, like in the `select` functions ( since we are calling loc under d hood) or d `move` function (iloc under d hood).

ericmjl:
I think we can merge this PR. Do you feel it's ready, @samukweku? If so, I'll hit the big green button!

samukweku:
@ericmjl yes U can

janitor/functions/pivot.py
thatlittleboy: Sorry for the delayed reply, yeah I think making it clear that the copy is made by the `.loc` is good enough for me.

PR #1351, opened 2024-04-20
Samukweku/polars clean names

### LINKED ISSUE #1343: Make clean_names() compatible with polars and geopandas dataframes
# Brief Description

I would like to propose for the clean_names() method to work with polars dataframe and geopandas dataframe. I work in the climate domain and also came from R. Polars has become a default choice to work with super large tabular climate datasets thanks to its implicit parallelism and efficient memory management. geopandas is a default go-to for working with GIS data. Would be great to see clean_names() method be compatible with both.

# Example API

in polars, would look like....
`polars.clean_names()`

in geo pandas would look like...
`gdf.clean_names()`

samukweku:
# PR Description

- polars submodule created.
- `clean_names` function extend to work on Polars objects - works on a polars expression (cleans the column values) or a polars dataframe (cleans the column names).

**This PR resolves #1343**

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1351--pyjanitor.netlify.app

ericmjl:
@samukweku thanks for kickstarting this! I think I get where you're going with the PR, though I have some minor reservations about the current implementation. Would you be open to discussing here async before we commit to a decision on the PR?

My biggest reservation is that the implementation of `clean_names` in this PR deviates from the pattern that we use for `pandas` dataframes, namely, by registering a function as part of the dataframe global namespace. My secondary reservation is that the current implementation in this PR has code paths for polars and pandas within a single function. On thinking about the overall design and studying the [API extension documentation](https://docs.pola.rs/py-polars/html/reference/api.html) and I think there may be a different way we could consider here.

Firstly, it may make our lives easier in the long run if we have the polars-enabled functions within their own submodules, following the pattern for spark dataframes. The main reason for this is the incompatibilities in the dataframe APIs -- a pet peeve of mine, but alas, a symptom of a vibrant ecosystem! While this arrangement means the polars submodule is likely to be less well-developed in the long run, it does reduce the energy barrier to having polars enthusiasts contribute to the library because of reduced mental overhead with adding a code path within existing functions. This arrangement should address the secondary reservation I wrote about above.

Secondly, by taking advantage of separate implementations and submodules, we can establish a suite of idioms to follow for creating polars-enabled functions. Early on, I would suggest that we simply follow the polars API extension documentation at first, but later on, we could create a `polars-flavor` library, akin to the `pandas-flavor` library, in which we use similar patterns to register a `dataframe.jn` namespace for polars-enabled functions. In this new world, each of the polars-enabled functions would be decorated as follows:

```python
import polars_flavor as pf 

@pf.register_dataframe_method(namespace="jn") # jn is the default namespace, for janitor, but just showing here for completeness
def clean_names(...):
    ...
```

What are your thoughts, @samukweku?

samukweku:
+1 on your suggestions @ericmjl ... Allows for easier/better maintenance. I will make the changes

samukweku:
@ericmjl is this better?

ericmjl (approved):
Looks awesome to me, @samukweku! Thank you for kickstarting this off. I am going to focus my other PR reviews on the `polars` API implementations.

Issue #1352, opened 2024-04-21
implement similar functions for polars


samukweku:
in relation to #1343 - this is a list of functions missing in the polars library that could be implemented : 
- [x] `clean_names`
- [x] `pivot_longer`
- [x] ~`pivot_wider`~ covered by `pl.pivot`
- [x] `xlsx_tables`
- [x] `xlsx_cells`
- [x] `read_commandline`
- [x] ~`conditional_join`~ polars has a pl.join_where to cover this
- [x] `complete`
- [x] ~`expand_grid`~ `pl.join` with `how='cross'` covers this
- [x] `convert_excel_date`
- [x] `convert_matlab_date`
- [x] ~`convert_unix_date`~ `pl.from_epoch` covers this
- [x] ~`bin_numeric`~ `pl.Expr.cut` covers this
- [x] ~`concatenate_columns`~ can be replicated with `pl.concat_str`
- [x] ~`deconcatenate_columns`~ `pl.Expr.str.split` covers this
- [x] ~`factorize_columns`~ `pl.rank(dense)` or `pl.Expr.to_physical` covers this
- [x] ~`get_dupes`~ `Expr.is_duplicated()` covers this
- [ ] `jitter`
- [ ] `limit_column_characters`
- [ ] `min_max_scale`
- [x] ~`move`~ can be replicated with polars' selectors
- [x] `row_to_names`
- [x] ~`shuffle`~ `pl.Expr.shuffle` covers this
- [ ] `sort_naturally`
- [x] ~`take_first`~ `group_by.first()` covers this
- [ ] `also`

Care should be taken to not create the function, if an existing solution already exists for any of these functions (probably named differently, or a combination of existing polars functions that covers all use cases of any of the listed functions above)

samukweku:
the current `pivot_longer` implementation is not good enough. I'll submit a PR with improvements

samukweku:
I assumed (wrongly) that polars' join maintains order (it only does so for left join). need to rethink the computation logic for `complete`. I'll submit a PR with improvements

3SMMZRjWgS:
eagerly awaits for 0.28.0 release!

samukweku:
@3SMMZRjWgS version 0.28.0 is released. would love feedback on the functions - would also love PRs if you are interested.

samukweku:
[...]

PR #1363, opened 2024-05-20
[ENH]`row_to_names` for polars

samukweku:
# PR Description

- Add `row_to_names` method for polars

**This PR relates to #1352 **

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1363--pyjanitor.netlify.app

ericmjl:
@samukweku it looks like this one also needs to have merge conflicts resolved before we can continue. Would you be kind enough to take a look at it please?

ericmjl:
@samukweku it looks like merging other polars dataframe functions results in large merge conflicts, which may make further development on the `polars` implementation of functions a bit more tricky. We may need to think carefully about how the code is organized here.

Do you think there is a way to leverage the polars API to create a `polars-flavor` package, not unlike the `pandas-flavor` package that we currently maintain? I think that would be the easiest way for us to handle the matter.

I tried thinking of a few other strategies, but none of them seem satisfactory. This included dynamically monkey-patching the docstrings of the PolarsFrame class methods from the underlying implementation function. But that has problems, just like other ideas I thought of, and didn't seem satisfactory.

samukweku:
@ericmjl i thought of going with the pandas_flavor format but decided against it (polars has three namespaces which we support - lazyframe, expr and eager frames - it would be an issue in the docs rendering for these namespaces) - so i chose another route, creating 3 files for each of the namespaces. Still not sure if this is good enough - hopefully there is a better way. or maybe there is a better way to modify pandas_flavor to better support the namespaces?

ericmjl:
@samukweku that's a good point, I didn't think of that.

PR #1396, opened 2024-09-06
[ENH] More control on creating the output dataframe for `conditional_join` - PR no 1

samukweku:
This is the first part of a series of PRs to improve conditional_join, with the final aim being to support aggregations within conditional_join. Supporting aggregations within conditional_join means we can avoid materializing the dataframe before aggregating, and should generally improve performance for such operations.

# PR Description

Please describe the changes proposed in the pull request:

- more control on how the final dataframe is created
- special focus on non inner join operations ('left','right','outer')

**This PR relates to #1269 .**

Please tag maintainers to review.

- @ericmjl

Issue #1269, opened 2023-06-14
add row count for conditional_join


samukweku:
# Brief Description

add a row count for matches from the left dataframe in the right dataframe

# Example API

```python
(d2
.conditional_join(
    d1, 
    ('pos', 'segment.start', '>='), 
    ('pos', 'segment.end', '<='),
    ('chr', 'chr', '=='), 
    return_counts=True)
)
```

samukweku:
We can possibly extend this idea to aggregate (sum, min,max,...) ... Might be limited to numba at the moment, unless we figure out an efficient way to take advantage of pandas groupby without building a dataframe first 🤔

PR #1397, opened 2024-09-07
[ENH] `conditional_join` can return ragged_arrays where possible - PR no. 2

samukweku:
This is the second part of a series of PRs that ultimately adds support for aggregations within `conditional_join`.  where possible, `ragged_arrays` can be returned to the user, either as slices or arrays of indices, which can be used in `akimbo`, or `awkward` or `pyarrow` to aggregate the data. this should be faster than materializing the entire dataframe within pandas before aggregating.

# PR Description

- return ragged_arrays where possible with the `get_join_indices` function
- limited to single join conditions or a range join, where the columns on the right dataframe are both monotonically increasing

**This PR relates to #1269 and #1396 .**

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1397--pyjanitor.netlify.app

janitor/functions/_numba.py
```diff
@@ -166,7 +166,7 @@ def _numba_equi_join(
     # this also lets us know if there are equi matches
     keep_rows = slice_starts < slice_ends
     if not keep_rows.any():
-        return None
+        return None, None
```
ericmjl: @samukweku do we have documentation in-line about what the tuple elements `(None, None)` semantically mean?

ericmjl (approved):
Pre-approving, @samukweku! Just one question re: documentation.

ericmjl:
I also noticed that the project test coverage has significantly gone down to <90%. I might need to dig further, but this is definitely a sign that we may need to invest more in covering edge cases throughout the repo.

janitor/functions/_numba.py
samukweku: thanks for the prompt. I'll refactor the code to a clearer form

PR #1400, opened 2024-09-12
Samukweku/pivot longer simple

samukweku:
# PR Description

- all forms of pivot_longer handled by two major functions `reshape_by_spec` and `reshape_by_spec_others`
- at its core, any pivot_longer function has to create a spec and pass to one of the two functions above.
- add lots of comments on the implementation logic (for future me or others)
- deprecate `dropna` - no perf difference if user does it, and is not needed within `pivot_longer`, plus we avoid the complexity of determing which NA rows to drop

**This PR improves pivot_longer.**

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1400--pyjanitor.netlify.app

PR #1440, opened 2025-01-25
ENH: Add read archive function 

### LINKED ISSUE #1171: Read files from archive - zip, tar, etc
# Brief Description

I would like to propose `read_archive` to read from zip/tar/...files in a simple form. Also read multiple files or nested files.

# Example API

Please modify the example API below to illustrate your proposed API, and then delete this sentence.

```python
df.read_archive(path, filename) 
```

Sabrina-Hassaim:
# PR Description

_**1. Implementation of the read_archive Function:**_

- Added a new method to read archive files (.zip, .tar, .tar.gz) and extract their contents as a DataFrame or a list of compatible files.

- Supports CSV and Excel file formats within the archives.

_**2. Unit Tests**_

- Added tests to validate the behavior of the read_archive method:
- Ensures correct reading of files from .zip and .tar.gz formats.
- Handles cases where the file is not a valid archive or does not contain compatible files.
- Tests include interactive behavior for file selection.

**This PR resolves #1171 **

# PR Checklist

# Automatic checks

# Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/io.py
+++ b/janitor/io.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import inspect
 import os
 import subprocess
+import tarfile
 import warnings
+import zipfile
 from collections import defaultdict
 from glob import glob
 from io import StringIO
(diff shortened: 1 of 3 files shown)


samukweku:
@Sabrina-Hassaim kudos on getting docs and tests working ... kindly work on adding tests for the missing parts, as indicated by the CI. from there we can work on the code itself ... i've got a couple of suggestions but we could do it in steps, after coverage is ok for the existing setup

samukweku:
also @Sabrina-Hassaim kindly close the other PR

janitor/io.py
```diff
@@ -689,3 +691,187 @@ def _object_to_dict(obj):
...
+        member = archive.getmember(selected_file)
+        with archive.extractfile(member) as file:
+            if selected_file.endswith(".csv"):
+                dfs.append(pd.read_csv(file))
+            elif selected_file.endswith(".xlsx"):
+                dfs.append(pd.read_excel(file))
+    return dfs if len(dfs) > 1 else dfs[0]
+
+
+def _select_files_interactively(compatible_files: list[str]) -> list[str]:
```
samukweku, line 803: i'm not sure we should support this - is there any benefit to this? @ericmjl @pyjanitor-devs/core-devs thoughts?

janitor/io.py
```diff
@@ -689,3 +691,187 @@ def _object_to_dict(obj):
...
+
+#################################################################
+
+
+def read_archive(
+    file_path: str,
+    extract_to_df: bool = True,
+    file_type: str | None = None,
+    selected_files: list[str] | None = None,
+) -> pd.DataFrame | list[str]:
```
samukweku, line 704: i think we should allow for more flexibility, via kwargs, where you can pass extra info to `read_csv`, `read_excel`, `read_parquet`, etc

samukweku, line 704: we should add an `engine` argument to support other dataframe libraries, e.g polars. Have a look at some of the IO functions that support polars

samukweku:
kindly add a line to changelog.md

janitor/io.py
ericmjl: I think it's worth keeping around just to see what it might do for the library. If it turns out not to be used very widely we can just deprecate it at a later date. On the other hand, if it's very popular, then we have the benefit of having it around.

PR #1460, opened 2025-03-30
[ENH] Fix `summarise` for MultiIndex

samukweku:
# PR Description

Please describe the changes proposed in the pull request:

- Fix error when MultiIndex is used with `summarise`

```
def dfmi():
    """Create a MultiIndex DataFrame"""

    # https://pandas.pydata.org/docs/user_guide/advanced.html#using-slicers
    def mklbl(prefix, n):
        return ["%s%s" % (prefix, i) for i in range(n)]

    miindex = pd.MultiIndex.from_product(
        [mklbl("A", 4), mklbl("B", 2), mklbl("C", 4), mklbl("D", 2)]
    )

    micolumns = pd.MultiIndex.from_tuples(
        [("a", "foo"), ("a", "bar"), ("b", "foo"), ("b", "bah")],
        names=["lvl0", "lvl1"],
    )
    dfmi = (
        pd.DataFrame(
            np.arange(len(miindex) * len(micolumns)).reshape(
                (len(miindex), len(micolumns))
            ),
            index=miindex,
            columns=micolumns,
        )
        .sort_index()
        .sort_index(axis=1)
    )
    dfmi.index.names = list("ABCD")
    return dfmi


df = dfmi()
df.summarise(('a',['min','max']),by=('a','bar'))

lvl0        a               
lvl1      bar       foo     
          min  max  min  max
(a, bar)                    
1           1    1    0    0
5           5    5    4    4
9           9    9    8    8
13         13   13   12   12
17         17   17   16   16
...       ...  ...  ...  ...
237       237  237  236  236
241       241  241  240  240
245       245  245  244  244
249       249  249  248  248
253       253  253  252  252

[64 rows x 4 columns]

```

**This PR resolves #1459 .**

# PR Checklist

Please ensure that you have done the following:

1. [x] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [x] If you're not on the contributors list, add yourself to `AUTHORS.md`.

3. [x] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.
    - Do use some discretion here; if there are multiple PRs that are related, keep them in a single line.

# Automatic checks

There will be automatic checks run on the PR. These include:

- Building a preview of the docs on Netlify
- Automatically linting the code
- Making sure the code is documented
- Making sure that all tests are passed
- Making sure that code coverage doesn't go down.

# Relevant Reviewers

[...]

PR #1462, opened 2025-04-08
[ENH] Method for adding functionality to pandas groupby

### LINKED ISSUE #587: [ENH] Method for adding functionality to GroupBy
It would be nice to be able to add functionality to the Pandas `GroupBy` objects: `GroupBy`, `DataFrameGroupBy`, `SeriesGroupBy`. There's no convenient accessor interface to do this, but maybe there's a way to reliably monkeypatch them. This would allow us to create nifty aggregation / apply functions and avoid the `.groupby(...).apply()` route for tasks we may encounter routinely. It could also potentially open up opportunities to speed up such operations... `.groupby().apply()` can often be slow for large numbers of groups.

samukweku:
# PR Description

- Create function that allows creation of methods on pandas GroupBy objects, similar to the accessors/methods we have for pd.Series/pd.DataFrame
- Extend `jn.select` to pandas GroupBy object - ```df.groupby('a').select('col_*')```

**This PR resolves #587 .**

- @ericmjl


ericmjl:
🚀 Deployed on https://deploy-preview-1462--pyjanitor.netlify.app

janitor/functions/groupby_flavor.py
ericmjl, line 1: @samukweku I like this! Any chance we could move this into `pandas-flavor` instead? I think you have admin privileges on the repo, and can also do releases there.

janitor/functions/select.py
```diff
@@ -327,17 +329,35 @@ def select_rows(
     return _select(df, rows=list(args), invert=invert)
 
 
+@register_groupby_method
+def pull(df: DataFrameGroupBy, label: Any) -> pd.Series:
```
ericmjl: Is there a reason for this to be named `pull` instead of `select`? (I'd imagine with the latter, you'd probably have to move this out to a different `.py` file?)

ericmjl (approved):
Pre-approving! Just wanted to think twice about the name `pull` -- it defines the API, so it's worth thinking more deeply about this.

janitor/functions/select.py
samukweku: `groupby.select` returns a groupby object, while `groupby.pull` reaches further and returns a series. I am open to a better name (naming is hard 😁)

samukweku:
@ericmjl can I get your review again for #1460

janitor/functions/groupby_flavor.py
samukweku: PR here: https://github.com/pyjanitor-devs/pandas_flavor/pull/36

samukweku:
any suggestions on how to update conda with pandas_flavor @ericmjl ?

janitor/functions/select.py
samukweku: I think we can park this for now (dont push this API) while we consider a better approach.  I've got some ideas that I'll probably push in a separate issue. I'll update the PR accordingly.

samukweku: @ericmjl I have excluded the `pull` function.

ericmjl (approved):
Oh my, I just saw the re-review request on this one, @samukweku. My many apologies! I'm in agreement with the code changes, let's merge it!

Issue #1476, opened 2025-05-27
Make `filter_column_isin` able to handle multiple columns


raffaelemancuso:
I propose to make `filter_column_isin` able to handle multiple columns.

Here is an example (currently doesn't work):

```python
df = pd.DataFrame(
     {
         "Year": [1999, 2000, 2004, 1999, 2004],
         "Taxon": [
             "Saccharina",
             "Saccharina",
             "Saccharina",
             "Agarum",
             "Agarum",
         ],
         "Abundance": [4, 5, 2, 1, 8],
     }
 )

myfilt = [[1999, "Saccharina"], [2000, "Saccharina"]]

# Works
mask = df.set_index(["Year", "Taxon"]).index.isin(myfilt)
df[mask]

# Doesn't work
df.filter_column_isin(["Year", "Taxon"], myfilt)
```

samukweku:
a dictionary approach would be cleaner. At the same time maybe we deprecate and just use pandas isin directly?

chentalia317-alt:
Hi!! I’ve opened a PR that brings filter_column_isin back with multi-column and dict-based support. It keeps backward compatibility for single-column usage. Happy to see the feedback!

samukweku:
This function should be deprecated; pandas' isin already solves this. Lets point users to pandas isin and possibly add usage examples as well. Cc: @ericmjl @pyjanitor-devs/core-devs

samukweku:
```py
year,taxon=zip(*myfilt)
dictionary = {'Year':year,'Taxon':taxon}
In [23]: df.isin(dictionary).any(axis=1)
Out[23]:
0     True
1     True
2     True
3     True
4    False
dtype: bool

In [24]: df.loc[df.isin(dictionary).any(axis=1)]
Out[24]:
   Year       Taxon  Abundance
0  1999  Saccharina          4
1  2000  Saccharina          5
2  2004  Saccharina          2
3  1999      Agarum          1
```

samukweku:
Closing this as `filter_column_isin` is deprecated

PR #1519, opened 2025-10-18
Extend filter_column_isin to support multi-column and dict-based API

chentalia317-alt:
Addresses #1476
Summary: reintroduces and extends filter_column_isin in filter.py.
Supports Single-column, multi-column, and dict-based filtering
Adds complement flag for inverse filtering.


samukweku:
Thanks @chentalia317-alt we are still having a conversation on whether this function should be deprecated and point users to the pandas' `isin` function

samukweku:
closing this, as the function will be deprecated

PR #1537, opened 2025-12-06
DeclanBuchanan add_column automatic fill remaining

DeclanBuchanan:
## PR Description

- the function add_column willl now automatically fill remaining, without the user needing to use the fill-remaining parameter
-
-

**This PR resolves #(put issue number here, and remove parentheses).**

## PR Checklist

1. [✓ ] PR in from a fork off your branch. Do not PR from `<your_username>`:`dev`, but rather from `<your_username>`:`<feature-branch_name>`.

2. [✓ ] If you're not on the contributors list, add yourself to `AUTHORS.md`.

3. [✓ ] Add a line to `CHANGELOG.md` under the latest version header (i.e. the one that is "on deck") describing the contribution.

## Automatic checks

## Relevant Reviewers

- @ericmjl

### DIFF
--- a/janitor/functions/add_columns.py
+++ b/janitor/functions/add_columns.py
@@ -96,6 +96,9 @@ def add_column(
 
     nrows = len(df)
 
+    if fill_remaining:
+        warnings.warn("The fill_remaining parameter is no longer being used")
+
     if hasattr(value, "__len__") and not isinstance(value, (str, bytes, bytearray)):
         len_value = len(value)
 
--- a/janitor/functions/add_columns.py
+++ b/janitor/functions/add_columns.py
@@ -114,18 +114,18 @@ def add_column(
         if not len_value:
             raise ValueError("`value` has to be an iterable of minimum length 1")
 
-    elif fill_remaining:
-        # relevant if a scalar val was passed, yet fill_remaining == True
-        len_value = 1
-        value = [value]
-
-    df = df.copy()
-    if fill_remaining:
+        #automatic fill remaining
         times_to_loop = int(np.ceil(nrows / len_value))
         fill_values = list(value) * times_to_loop
         df[column_name] = fill_values[:nrows]
-    else:
-        df[column_name] = value
+
+    else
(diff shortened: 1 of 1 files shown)


ericmjl (commented):
## Code Review Feedback

Thanks for the PR! I understand the goal is to make `fill_remaining` automatic.

The essence of "automatic fill remaining" is a **behavior change**: currently, when `fill_remaining=False` (the default) and the value length doesn't match the dataframe length, it raises an error. Making it "automatic" means this error would no longer occur - values would be filled/repeated instead.

**This is a breaking change** that would require a minor version bump (e.g., 0.x → 0.y) -- that's the versioning convention we stick to pre-1.0.

### Suggested Approach

If we're going to make this behavior change, the **simplest and cleanest implementation** is to:

1. Change the default parameter from `False` to `True`:

   ```python
   fill_remaining: bool = True,  # Changed from False
   ```

2. Revert all the other logic changes - the existing code already handles `fill_remaining=True` correctly.

3. Remove the unnecessary additions:
   - The warning on lines 99-100
   - The broken `else` branch (lines 125-126)
   - The logic modifications that break scalar handling

This approach:

- Achieves the goal with minimal code changes
- Keeps all existing logic intact (no risk of new bugs)
- Maintains backward compatibility (users can still pass `fill_remaining=False`)
- Requires a version bump (which is necessary for any behavior change anyway)

The current implementation tries to achieve the same goal but introduces bugs and unnecessary complexity. If we're making a breaking change, let's do it the simple way.

samukweku:
@ericmjl should we be updating this function? It is slated for depreciation, and we advise users to use another function instrad

ericmjl:
@samukweku I think we have enough syntactic sugar here that this can be kept around, actually. Now that I'm thinking of it, I'm can't exactly remember why we decided to deprecate it to begin with. Do you remember, @samukweku?

samukweku:
@ericmjl I believe it is covered mostly by the `assign` method

samukweku:
Of course nothing is set in stone

PR #1594, opened 2026-02-12
[DOC] Incorrect attribution (previous account name)

Nalmeder:
## PR Description

Contributions attributed to deprecated account name.

Also, I believe I contributed more than a bug fix, but I assume the username change messed up the history there. One thing at a time, though.

Would you like me to add this under changelog.md or is this small enough it shouldn't matter?

- @ericmjl

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -70,10 +70,10 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


ericmjl:
This change will suffice!

Issue #1659, opened 2026-08-20
[PERF] Selectivity-aware anchor selection for range joins (le_lt/ge_gt)


samukweku:
## Background

#1641 (implemented in #1658) fixed anchor selection for the *same-direction* multi-predicate path (`mapping["le_or_ge"]` in `_le_ge_1_or_more.py`): the anchor used to always be whichever `<`/`<=`/`>`/`>=` predicate the caller listed first, regardless of selectivity.

The same "first encountered" pattern exists for **range joins** - queries combining one `<`/`<=` predicate with one `>`/`>=` predicate (e.g. `a < right.hi` and `a > right.lo`). `_helpers._separate_conditions_based_on_op` picks `le_lt` and `ge_gt` this way:

```python
# check for possibility of a range join
# keep the first match for le_lt or ge_gt
le_lt = None
ge_gt = None
for condition in conditions:
    left_on, right_on, op = condition
    if le_lt and ge_gt:
        break
    if (op in less_than_join_types) and not le_lt:
        le_lt = (left_on, right_on, op)
    elif (op in greater_than_join_types) and not ge_gt:
        ge_gt = (left_on, right_on, op)
```

If a caller supplies multiple eligible `<`-type or multiple eligible `>`-type predicates, whichever appears first in argument order wins, with no selectivity awareness - the same class of bug #1641 fixed, in a different dispatch branch (`_range_join_default.py` / `_range_join_regions.py`).

## Why this is a separate issue, not part of #1657

#1657 extends the *same* fix (`_select_anchor` in `_le_ge_1_or_more.py`) to `keep='all'` - same file, same mechanics, just relaxing a guard.

This is structurally different: range joins narrow candidates using a lower **and** upper bound simultaneously (two boundaries at once), not a single anchor window against which the rest are post-filtered. A fix here likely can't be a straight copy of `_select_anchor` - it needs its own design for how selectivity is estimated and combined when two bounds are chosen independently (e.g. do you evaluate `le_lt` and `ge_gt` selection independently, or jointly since together they define the actual candidate range?). Folding it into #1657 would blur that issue's scope and make both harder to close independently.

## Ask

[...]

Issue #1657, opened 2026-08-20
[PERF] Extend selectivity-aware anchor predicate selection to keep='all'


samukweku:
## Background

#1641 adds selectivity-aware anchor predicate selection for the multi-predicate `<`/`<=`/`>`/`>=` path in `conditional_join` (`_le_ge_1_or_more.py`), but scopes it to `keep='first'` and `keep='last'` only.

That scoping was deliberate, not just caution: for those two modes, output is **provably invariant** to which predicate is chosen as anchor — `keep='first'`/`'last'` resolve to a true min/max reduction over the right index across the full matching window (see `janitor-rs/src/index_builder.rs`, `index_starts_only_keep_first`/`_last`), so reordering the anchor cannot change the result at all, only its speed.

`keep='all'` doesn't have that property. The anchor predicate's column determines how `right` gets sorted before candidate pairs are scanned, and `build_indices_matches`'s `keep=="all"` branches (`_helpers.py`) emit matching right-index values in that scan order. So picking a different anchor for `keep='all'` changes **row order** in the output, even though the **set** of matched (left, right) pairs is identical either way (same values, dtypes, null semantics - see discussion on #1641).

## Ask

Extend the same selectivity-aware `_select_anchor` logic (or an equivalent) to the `keep='all'` path, accepting that the current (undocumented, implementation-defined) output row order may change as a result. This trades away an incidental ordering guarantee for the same class of performance win #1641 delivers for `keep='first'`/`'last'` - potentially large on multi-predicate joins with a badly-ordered anchor (#1641 reports up to ~77x at 10k rows).

## Constraints

- Match-set correctness must remain exact: same values, dtypes, null handling, extension-array behavior - only row *order* is allowed to change, never content.
- This is a user-visible behavior change (even though row order was never a documented API guarantee for `keep='all'`) and should be called out explicitly in the changelog, since some downstream code may incidentally depend on today's ordering.
- Open design question for whoever picks this up: leave the new order fully unspecified/implementation-defined (simplest, matches today's status quo of "order isn't guaranteed"), or pay for a final normalization pass (e.g. sort by left index) if that's cheap enough not to erode the performance win - needs a quick benchmark either way.

## Related

- Follow-up to #1641.

Issue #1641, opened 2026-08-19
[PERF] Optimize conditional_join anchor predicate selection


samukweku:
## Brief

The maintained non-Numba conditional-join path currently chooses the first eligible non-equality predicate as its anchor. Predicate order can therefore make equivalent joins differ substantially in runtime and peak memory when one predicate is much more selective than another.

A local benchmark with two equivalent less-than conditions and keep=first showed the broad-first ordering becoming increasingly expensive compared with selective-first ordering:

- 1,000 rows: about 2x slower
- 3,000 rows: about 8.8x slower
- 10,000 rows: about 77x slower (about 100 ms versus 1.3 ms)

## Goal

Research and implement a readable predicate-selection strategy that chooses a useful anchor before materializing candidate pairs. Possible approaches include:

- cheap selectivity estimates from column min/max values and distributions;
- exact or sampled candidate-count estimates;
- a conservative optimization limited to keep=first and keep=last;
- early rejection of predicates that would produce very broad candidate sets.

## Correctness constraint

Anchor choice can affect the ordering of matches returned by keep=all. Any optimizer must preserve current row ordering as well as values, indexes, dtypes, extension-array behavior, and null semantics. This ordering constraint is why the optimization should not be folded into the result-materialization work in #1632 / #1638.

## Benchmark and acceptance matrix

Cover at least:

- two or more eligible non-equality predicates with different selectivities;
- operators <, <=, >, and >=;
- keep=all, keep=first, and keep=last;
- sorted and unsorted inputs;
- dense, sparse, zero-match, and full-match cases;
- nullable and extension dtypes where supported.

Acceptance requires exact behavioral parity, including output ordering, with measurable improvement on badly ordered predicates and no material regression on already selective-first inputs. The implementation should remain understandable without relying on the deprecated Numba path.

## Related work

- #1415
- #1125
- #1632
- #1382
- #1638

PR #1667, opened 2026-08-21
[PERF] Extend selective bound selection to keep='all' for range joins

samukweku:
## Summary

Extends #1659's selective `le_lt`/`ge_gt` bound selection for range joins
(a `<`/`<=` predicate combined with a `>`/`>=` predicate) to `keep='all'`,
not just `keep='first'`/`'last'`. Removes the `keep in ('first', 'last')`
guard around `_maybe_select_better_range_bounds` in
`_get_indices_non_equi.py`.

This mirrors #1657's identical extension of #1658/#1641's same-direction
anchor selection to `keep='all'`, on the same basis: the matched row
*content* is invariant to bound choice for every `keep` mode (same proof
#1641/#1658/#1657/#1659 already established and reuse). Only `keep='all'`
row *order* can change - and that was never documented or guaranteed to
begin with.

Depends on #1663 (base branch), which is itself stacked on #1658/#1662.

## ELI5

`conditional_join` can take several `<`/`<=`/`>`/`>=` conditions at once.
For range joins (one `<`-type plus one `>`-type condition), when you give
it *more than one* candidate for either side, it used to just always use
whichever one you happened to list first as the "anchor" - the one it
uses to narrow down candidate rows before checking the rest. If that
first one wasn't very selective (didn't rule out many rows), the join
could get much slower - unboundedly so for `keep='all'`, since it's not
just picking one match per row like `keep='first'`/`'last'` does.

#1659 already fixed this for `keep='first'`/`'last'` by picking the more
selective candidate instead of just the first-supplied one. This PR
applies the exact same fix to `keep='all'`. The only thing that can
change is *which order* the output rows come back in (which was never
promised to be any particular order anyway) - never *which* rows, or
their values.

## Benchmark

Same skewed-data shape as #1659/#1666's issue write-up (one narrow `<`
candidate, one broad `<`-type "distractor" candidate, `keep='all'`),
before vs. after this fix:

| rows | before | after |
|-----:|-------:|------:|
| 1,000 | ~0.005s (2.18x baseline) | 0.0027s |
| 3,000 | ~0.03s (5.84x) | 0.0025s |
| 10,000 | ~0.3s (32.05x) | 0.0087s |
| 30,000 | ~3.5s (154.30x, still growing) | 0.0613s |

The pre-fix pathology for `keep='all'` was unbounded (kept growing past
30k rows), unlike the mild, plateauing ~2-9x seen for `keep='first'`/
`'last'` before #1659. After this fix, `keep='all'` timings track the
same selective-bound-choice cost as `keep='first'`/`'last'` already do.

## Test plan

- New tests in `tests/functions/test_conditional_join.py`:
  - `test_range_join_bound_selection_used_for_keep_all` - asserts
    `_maybe_select_better_range_bounds` is now invoked (via `wraps=`) for
    `keep='all'`, inverting the old assert-not-called test.
  - `test_range_join_le_lt_content_invariant_keep_all` (both
    `join_algorithm` values) - matched row content is order-of-argument
    invariant for `keep='all'` (compared sorted).
  - `test_range_join_row_order_can_differ_for_keep_all_on_tie` - documents
    the accepted trade-off with an empirically-verified repro: row order
    can differ between argument orders only when there's an actual tie in
    the `ge_gt` right column that also isn't already monotonic (a
    constant `ge_gt` column never triggers a real sort, so never
    reproduces this) - content still matches once sorted.
- `pixi run pytest tests/functions/test_conditional_join.py -n auto --color=no -q` - 291 passed, 1 skipped.
- `pixi run pytest --doctest-modules janitor/functions/conditional_join.py` - 2 passed.
- `pixi run lint` on changed files - clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -400,6 +400,24 @@ that looked like it might fade out below 30k rows instead held flat at
 ~1.8-1.9x all the way to 50M, which is the number that actually matters
 for deciding whether it's worth fixing.
 
+### Conditional-Join Performance Research
+
+When investigating current or future `conditional_join` performance work,
+include the region-number algorithms from Dathan and Trausan-Matu,
+*Algorithms for Computing Inequality Joins* (DATA 2018), among the design
+options considered: <https://doi.org/10.5220/0006826803570364>.
+
+The paper's two-comparison algorithm underlies the existing `regions` path.
+Its Section 3.2 also describes a distinct multi-comparison strategy that
+computes regions for every predicate and dynamically chooses the driving
+field. Consider both the existing implementation and the paper's fuller
+algorithm when researching predicate selection, field-role assignment,
+relation orientation, candidate materialization, or other inequality-join
+optimizations. Treat the paper as an option to evaluate, not an automatic
+implementation mandate: preserve pyjanitor's ordering, dtype, null, index,
+and extension-array behavior, and validate performance claims against
+representative real inputs and the large-scale benchmark guidance above.
+
 ### Code Style Rules
 
 - **Line length**: 88 characters (ruff default)
@@ -580,6 +598,8 @@ pass, and removed from here to avoid restating the same rule twice.
 - **2026-08-21**: Added mandatory adversarial review before every PR
 - **2026-08-21**: Added PR/issue writing convention (ELI5 for non-obvious changes)
 - **2026-08-21**: Added convention to extend benchmarks to scale (10M-50M rows)
+- **2026-08-21**: Added the inequality-joins paper as a design reference for
+  current and future `conditional_join` performance work
 - **2026-08-21**: Cleanup pass - removed 5x-duplicated markdownlint rule, 3x
   -duplicated pixi/notebook rules, and Learned Patterns entries already
   integrated into main sections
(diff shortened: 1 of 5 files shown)


samukweku:
Adversarial review (independent worktree, fresh context): **SAFE TO MERGE AS-IS**.

Verified independently (own repro, not reusing this PR's fixtures): content-invariance to bound choice for `keep='all'` holds across duplicate right-index rows, NaN in non-anchor columns, nullable `Int64` dtype, and empty-match cases, both `join_algorithm` values. Full suite (291 passed, 1 skipped) and doctests (2 passed) reproduced independently; lint clean on changed files. Benchmark trend (unbounded growth before, flat after) reproduced on an independent run, though absolute numbers differ by machine/seed as expected.

One non-blocking note: this diff also folds in an in-place `.index()`-based swap for `le_or_ge` (replacing the old list-rebuild), not just the `keep in ('first','last')` guard removal. Traced it - correct (le_lt/ge_gt candidate pools are structurally disjoint, so no stale-state issue across the two loop iterations) - but it's scope creep relative to "just remove the guard," worth calling out explicitly for anyone skimming the diff. No action needed unless you want it split out.

Also confirmed: `_get_indices_equi.py` (equi+range-join combo path) still doesn't call this selection logic - unfixed, but this PR never claims to cover it, so no overclaim.

PR #1681, opened 2026-08-24
[PERF] Avoid materializing not-equal pairs for keep first or last

### LINKED ISSUE #1651: [PERF] Avoid materializing all != matches for keep first/last
## Brief description

Optimize `conditional_join` with a `!=` predicate and `keep="first"` or
`keep="last"` so it does not materialize nearly every matching pair before
discarding all but one match per left row.

The current `_not_equal_indices` implementation constructs the `<` and `>`
pair arrays, adds null-related pairs, concatenates them, and only then calls
`_keep_output`.

## ELI5

If we only need the first different item, we should stop as soon as we know
which item that is. The current code writes down every different item and then
throws away all but the first or last one.

## Proposed direction

Investigate an O(n + m) or O((n + m) log m) selector that finds the earliest or
latest right position whose value is not equal to each left value. One possible
approach is to track the first/last positions of the first two distinct right
values, with dedicated handling for nulls.

The algorithm must preserve output ordering and the project's existing null
semantics. In particular, `pd.NA` and `np.nan` do not necessarily behave the
same way for inequality.

`keep="all"` is not the target: its output can genuinely contain nearly every
pair and therefore cannot avoid output-proportional work.

## Acceptance criteria

- `keep="first"` and `keep="last"` avoid constructing all matching pairs.
- Results match the current implementation for sorted and unsorted right
  inputs, duplicate values, one or many distinct values, empty inputs, all-null
  inputs, nullable extension dtypes, and floating `NaN`.
- Match ordering remains unchanged.
- Benchmarks cover low- and high-cardinality data at several sizes and show an
  end-to-end improvement.
- `keep="all"` behavior remains unchanged unless separately justified.
- The core algorithm has an ELI5 comment.
- The changelog entry includes this issue number, the implementing PR number
This should not be folded into #1641; it is a specialized output-selection
algorithm with a sensitive null contract.

tunglambk:
## PR Description

- select the first or last unequal right row from at most two candidates instead of materializing every unequal pair
- preserve the existing materialized-path ordering and null semantics
- cover duplicates, sorted and unsorted inputs, single-value inputs, nullable integers, datetimes, floating NaN, all-null data, and empty inputs

## ELI5

For any left value, the first different right value is either the first right value or, if those are equal, the first value different from it. The same rule works backwards for `keep="last"`, so the join only needs two candidates rather than a nearly Cartesian list of pairs.

## Performance

End-to-end timings for 2,000 left and 2,000 right rows (best of three runs):

| Cardinality | Keep | Before | After | Speedup |
| ---: | --- | ---: | ---: | ---: |
| 2 | first | 47.252 ms | 3.541 ms | 13.3x |
| 2 | last | 50.935 ms | 3.529 ms | 14.4x |
| 2,000 | first | 87.863 ms | 3.451 ms | 25.5x |
| 2,000 | last | 83.081 ms | 3.441 ms | 24.1x |

## Correctness

- a direct-scan oracle covers the edge cases listed above for both keep modes
- an explicit regression test preserves the legacy output order
- 2,000 randomized int64, nullable-integer, and floating-NaN trials matched the materializing implementation
- `keep="all"` remains on the existing code path

## Validation

- focused conditional-join suite on current `dev`: 296 passed, 1 skipped
- documentation build passes locally
- GitHub's Python 3.11, 3.12, and 3.13 test jobs pass after the rebase
- the documentation deployment still fails at the upstream `gh-pages` permission step, and Codecov's project status remains stale; patch coverage passes

**This PR resolves #1651.**

## PR Checklist

1. [x] PR from a feature branch on a fork.
2. [x] Add a line to `CHANGELOG.md` under the latest version header.

## Relevant Reviewers

- @ericmjl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 - [DOC] Add documentation explaining when cumulative-event aggregation is preferable to range join aggregations. - Issue #1702 @sumangouda
 -   [ENH] Avoid materializing all unequal pairs in `conditional_join` with
-    `keep="first"` or `keep="last"`. - Issue #1651, PR #1681 @samukweku
+    `keep="first"` or `keep="last"`. - Issue #1651, PR #1681 @tunglambk
 -   [PERF] Reduce peak memory in `_build_indexer_reorder_contents` for wide frames (reps >= 8) using single NumPy allocation; tall frames with few repetitions retain the original reshape path. - Issue #1655 @Anupam2400
 -   [ENH] `conditional_join` now picks the most selective `<`/`<=`/`>`/`>=` predicate as its binary-search anchor (instead of the first one supplied) when `keep` is `'first'` or `'last'`, fixing pathological slowdowns from unfavorable predicate ordering; the choice is estimated from a fixed-size sample so the selection cost no longer scales with input size; `keep='all'` output is unaffected. - Issue #1641 @samukweku
 -   [BUG] Fix `conditional_join` crash for `keep='last'` joins with a single `<`/`<=` window and multiple non-equi conditions - a missing `counts` argument to the Rust `index_starts_only_keep_last` call. - Issue #1641 @samukweku
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 - [DOC] Add documentation explaining when cumulative-event aggregation is preferable to range join aggregations. - Issue #1702 @sumangouda
+-   [ENH] Avoid materializing all unequal pairs in `conditional_join` with
+    `keep="first"` or `keep="last"`. - Issue #1651, PR #1681 @samukweku
 -   [PERF] Reduce peak memory in `_build_indexer_reorder_contents` for wide frames (reps >= 8) using single NumPy allocation; tall frames with few repetitions retain the original reshape path. - Issue #1655 @Anupam2400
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased]
+-   [ENH] Avoid materializing all unequal pairs in `conditional_join` with
+    `keep="first"` or `keep="last"`. - Issue #1651, PR #1681 @samukweku
```
samukweku: @tunglambk kindly update this with your github username

samukweku (changes requested):
thanks @tunglambk 

kindly update the docs as well, informing users about this approach for `!=` join

tunglambk:
Addressed the review in d64e55ed: the changelog now credits @tunglambk, and the `conditional_join` documentation describes the optimized single-`!=` path for `keep="first"` and `keep="last"`. I also rebased onto the current `dev`. Locally, the full `conditional_join` test file passes (296 passed, 1 skipped), and the documentation build completes successfully. The CI documentation job reaches the preview deployment step, then receives a 403 when `github-actions[bot]` pushes to the upstream `gh-pages` branch.

samukweku (approved):

samukweku:
Thanks @tunglambk

samukweku:
@tunglambk kindly merge

tunglambk:
Rebased onto the current `dev`; the only conflict was the top of `CHANGELOG.md`, and both entries were retained. The focused suite passes with 296 tests and 1 skip, the documentation builds locally, and the Python 3.11–3.13 CI jobs are green. The docs deployment still reaches the upstream `gh-pages` permission failure noted earlier.

Issue #93, opened 2018-12-14
new function proposal: find and replace


ericmjl:
Colleague proposed this function: within a column that houses strings, within each cell, provide a `find_replace()` function that finds a substring in the cell, and replaces it with another substring.

Note to self:

- Figure out if this is already possible with pandas.
- Decide if implementing this provides a more fluent function API.

zbarry:
Should it support regular expressions? 
🤔

szuckerman:
Regex is ok.... but globbing is nicer:

https://docs.python.org/3.6/library/fnmatch.html

szuckerman:
The above looks good, except “exact” is unnecessary.  Regex or glob without wildcards will be equivalent.

szuckerman:
So they’re kind-of different things. I think this answer can give a better description of the differences than I could write here: 

https://askubuntu.com/questions/714503/regular-expressions-vs-filename-globbing

In general, when searching for things I find globbing more useful (cause you just need an asterisk) but for validation regex is better because one can be more precise with what is expected.

zbarry:
Yeah 90% of the time I just use globbing. I only use regex when I absolutely have to. No idea how often you would need them in pandas-style applications.

ericmjl:
Ok! Synthesizing the discussion thus far:

```python
def find_replace(df, find, replace, kind='glob'):
    """
    # docstring opener above
    Whether or not you use kind="glob" or "regex", if you do not use any wildcards,
    it will default to an exact match. We expect an 80/20 split between exact/wildcard
    usage, and with globbing being more easily understood than regexes, it is hence
    the default choice.
    # continue docstring below
    """
    kinds=['regex', 'glob']
    if kind not in kinds:
        raise ValueError(f'kind kwarg has to be one of {kinds}')
    if kind == 'regex':
        # implementation goes here
    else:
        # implementation goes here
    return df
```

Comments?

zbarry:
Needs a `column_name` / `col_name` (?), right?

ericmjl:
> Needs a column_name / col_name (?), right?

Thank you. Updating...

zbarry:
Also think that maybe `col_name` should come before `find` and `replace` in the args list... From the at-a-glance understanding perspective, maybe `df.do_this(to_this, with_this_particularly)` is more in line with what we have done with other functions.

Issue #91, opened 2018-12-14
why the scipy dependency ?


louis-red:
It does not seem to be needed anywhere in the source code

ericmjl:
Very good point - I think it's a legacy requirement. Thanks for the catch!

PR #87, opened 2018-12-12
Query

ericmjl:
This PR resolves #86.

# PR Checklist

Please ensure that you have done the following:

1. [x] PR in from a fork off your branch. Do not PR from <your_username>:master, but rather from <your_username>:<branch_name>.
2. [x] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

## Quick Check

To do a very quick check that everything is correct, follow these steps below:

- [x] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

If you are adding code changes, please ensure the following:

- [x] Ensure that you have added tests.
- [x] Run all tests (`$ pytest .`) locally on your machine.
    - [x] Check to ensure that test coverage covers the lines of code that you have added.
    - [x] Ensure that all tests pass.

## Documentation Changes

If you are adding documentation changes, please ensure the following:

- [x] Build the docs locally.
- [x] View the docs to check that it renders correctly.

# PR Description

Please describe the changes proposed in the pull request:

- Changes the implementation of filter_on to use the 'query' API in pandas. This enables nice "string" queries that pandas can parse underneath the hood. 

# Relevant Reviewers

Please tag maintainers to review.

- @zbarry 
- @szuckerman

PR #1684, opened 2026-08-25
[PERF] Update reverse starts-match aggregation boundary

samukweku:
## Summary

- update the Python boundary for reverse `starts + matches` aggregation across `sum`, `min`, `max`, `prod`, and `size`
- omit the redundant `length` argument for the new janitor-rs API
- retain a narrow fallback for older installed janitor_rs releases that still require `length`
- preserve the existing length argument for unrelated kernels and ordinary reverse paths

## Contract

The Rust kernels validate the starts/matches tape contract. `start == index.len()` is a valid zero-width suffix and consumes no matches entries. Negative starts remain invalid. Empty-input behavior is disclosed in janitor-rs issue #79.

## Verification

The focused reverse aggregation suite passes: `pixi run pytest tests/functions/test_conditional_join.py -k agg_rev -q` (20 passed).

Rust test and boundary changes are tracked in janitor-rs PRs #87, #88, #91, #92, and #93. Full old-vs-compact runtime and memory tables, including tiny through super-large duplicate/unique cases, are documented in janitor-rs issue #85.

Closes #85
Part of #23

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -548,6 +548,17 @@ CLI.
 include MkDocs. The documentation task is available in the `docs` environment.
 **Recommendation**: Run `pixi run -e docs build-docs` to build documentation.
 
+### [2026-08-26] Include performance evidence in optimization PRs and issues
+
+**Context**: Performance changes are coordinated with implementation changes
+in janitor-rs.
+**Learning**: Benchmark results are part of the performance change's review
+record, not merely local investigation notes.
+**Recommendation**: Every performance PR and its tracking issue must include a
+properly formatted comparison with the old implementation, covering runtime
+and memory for the agreed tiny, large, very-large, and super-large cases and
+duplicate/unique label distributions. State benchmark limitations explicitly.
+
 ---
 
 ## Version History
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -556,8 +556,8 @@ in janitor-rs.
 record, not merely local investigation notes.
 **Recommendation**: Every performance PR and its tracking issue must include a
 properly formatted comparison with the old implementation, covering runtime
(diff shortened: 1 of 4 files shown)


samukweku:
Resolved the CI failures from the released `janitor_rs` compatibility gap.

The pyjanitor test environment currently installs a released Rust extension whose starts+matches `sum`/`min` kernels still require `length`, while the paired PRs remove that redundant argument. The boundary now tries the new API first and retries with `index.size` only when the old extension reports the missing `length` argument.

Focused verification: `tests/functions/test_conditional_join.py -k 'agg_rev'` — 20 passed.

samukweku:
### Strict empty-tape contract

The Rust reverse match kernels intentionally reject an empty `matches` tape. Pyjanitor therefore must not call them when every candidate range is zero-width and the flattened tape is empty; it short-circuits that case and returns the empty aggregation result instead.

For non-empty tapes, Rust validates the tape shape. Pyjanitor remains responsible for supplying the binary 0/1 match values. This test covers the all-zero-width path so the strict Rust contract is enforced at the integration boundary rather than only documented.

samukweku:
### Empty-tape dispatch fix

The strict Rust contract remains intentional: direct reverse match kernels reject an empty `matches` tape, and reverse positions kernels reject an empty `positions` tape. Pyjanitor now enforces the integration side of that contract in `_get_join_aggs.py`: if the candidate tape is empty, it returns the correctly typed empty aggregation frame before dispatching to Rust, even when `left_index` still contains rows.

ELI5: Python checks whether there are any tickets to hand to Rust. If there are no tickets, it returns an empty answer itself; Rust only receives a real tape that it can validate.

Covered by the existing all-zero-width regression test. Targeted result: `2 passed`.

PR #94, opened 2018-12-15
Update AUTHORS.rst

jcvall:
This PR resolves #(put issue number here).

# PR Checklist

Please ensure that you have done the following:

1. [x] PR in from a fork off your branch. Do not PR from <your_username>:master, but rather from <your_username>:<branch_name>.
2. [x] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

## Quick Check

To do a very quick check that everything is correct, follow these steps below:

- [ ] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

If you are adding code changes, please ensure the following:

- [ ] Ensure that you have added tests.
- [ ] Run all tests (`$ pytest .`) locally on your machine.
    - [ ] Check to ensure that test coverage covers the lines of code that you have added.
    - [ ] Ensure that all tests pass.

## Documentation Changes

If you are adding documentation changes, please ensure the following:

- [ ] Build the docs locally.
- [ ] View the docs to check that it renders correctly.

# PR Description

Please describe the changes proposed in the pull request:

- 
- 
- 

# Relevant Reviewers

Please tag maintainers to review.

- @ericmjl

PR #1686, opened 2026-08-26
perf: update reverse ends aggregation boundary

samukweku:
Closes the pyjanitor side of janitor-rs issue #94.

Updates reverse ends+matches aggregation dispatch for the length-free Rust API and includes a narrow fallback for older installed janitor_rs releases that still require length. The ordinary reverse ends path keeps its existing length argument.

Tests: pixi run pytest tests/functions/test_conditional_join.py -k agg_rev -q (20 passed).

Benchmark: the corresponding Rust PRs compare old maps vs compact slots across tiny, large, very large, and super large duplicate/unique workloads; results are documented on janitor-rs issue #94.


samukweku:

Issue #1688, opened 2026-08-26
docs: explain conditional_join and reverse-aggregation index representations


samukweku:
## Goal

Create a dedicated documentation pass for `conditional_join` and `join_agg` that explains the internal join representations in plain language and documents the contracts relied on by the Python and Rust layers.

This should remain separate from performance implementation PRs. The purpose is to make the code understandable to users and contributors before further optimization work lands.

## ELI5 overview

Add a short, user-oriented explanation of the pipeline:

```text
find possible right-hand matches
  -> store them efficiently
  -> optionally filter those candidates
  -> aggregate values for each left-hand row
```

Explain that the implementation may represent matches in different forms depending on the join shape, rather than always materializing a full pair of left/right arrays immediately.

## Representations to document

Explain, with small concrete examples, the meaning and invariants of:

- `starts`: where each left row’s candidate range begins;
- `ends`: where each left row’s candidate range ends (exclusive);
- `counts_array`: number of candidates surviving for each left row;
- `matches`: a flat boolean/flag tape used to filter candidates;
- `positions`: a flat integer tape whose entries indirectly index `right_index`;
- `left_index` and `right_index`: whether they contain dataframe labels, row positions, or intermediate index values at each stage;
- `total`: number of materialized/surviving matches.

For `positions`, include an example such as:

```text
right_index = ["a", "b", "c", "d"]
positions   = [2, 0, 2, 3, 1]
starts      = [0, 2, 4]
ends        = [2, 4, 5]

left row 0 -> positions[0:2] -> ["c", "a"]
left row 1 -> positions[2:4] -> ["c", "d"]
left row 2 -> positions[4:5] -> ["b"]
```

## Trigger conditions

Document when each representation is produced, including the broad distinction between:

- simple/equi joins;
- starts-only and ends-only paths;
- range joins;
- multi-condition joins that retain a positions tape;
- additional conditions that filter an existing candidate tape;
- `keep="all"`, `keep="first"`, and `keep="last"`.

The documentation should explicitly explain why `positions` is needed: it is an indirection layer that allows candidate matches to remain compact and filterable before output labels are materialized.

## Docstring requirements

Improve docstrings for the relevant private helpers in `janitor/functions/_conditional_join/` so they state:

- accepted input shapes and dtypes;
- whether indexes are labels or positional indexes;
- range boundaries and whether `end` is exclusive;
- required parallel-array lengths;
- empty-range behavior;
[...]

PR #100, opened 2018-12-18
Convert malab date

ocefpaf:
This PR resolves #(put issue number here).

# PR Checklist

Please ensure that you have done the following:

1. [X] PR in from a fork off your branch. Do not PR from <your_username>:master, but rather from <your_username>:<branch_name>.
2. [ ] If you're not on the contributors list, add yourself to `AUTHORS.rst`.

## Quick Check

To do a very quick check that everything is correct, follow these steps below:

- [X] Run the command `make check` from pyjanitor's top-level directory. This will automatically run:
    - black formatting
    - pycodestyle checkin
    - running the test suite
    - docs build

## Code Changes

If you are adding code changes, please ensure the following:

- [X] Ensure that you have added tests.
- [X] Run all tests (`$ pytest .`) locally on your machine.
    - [ ] Check to ensure that test coverage covers the lines of code that you have added.
    - [X] Ensure that all tests pass.

## Documentation Changes

If you are adding documentation changes, please ensure the following:

- [ ] Build the docs locally.
- [ ] View the docs to check that it renders correctly.

# PR Description

Add a Matlab date converter akin to the the Excel date converter.

# Relevant Reviewers

- @ericmjl :wink:

PR #1690, opened 2026-08-26
perf: coordinate reverse positions aggregation APIs

samukweku:
## Summary

Coordinate the pyjanitor calls for all five reverse positions aggregations. Remove the redundant `length` argument from the positions wrappers and dispatch only the remaining positions aggregations with that argument.

## Related

- janitor-rs #100
- janitor-rs #23
- Documentation issue #1688

## Verification

- Pre-commit checks passed for formatting, lint, and docstring validation
- Python syntax checks passed
- Rust-side tests and old-vs-compact benchmarks are included in the five corresponding janitor-rs PRs

Runtime tests and benchmark execution remain pending because the local macOS host lacks the configured Python 3.9 framework.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -548,6 +548,23 @@ CLI.
 include MkDocs. The documentation task is available in the `docs` environment.
 **Recommendation**: Run `pixi run -e docs build-docs` to build documentation.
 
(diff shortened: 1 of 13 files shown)


samukweku:

PR #1692, opened 2026-08-26
perf: remove redundant explicit reverse range length

samukweku:
## Summary

Closes the pyjanitor side of janitor-rs#107 and contributes to #23.

The Rust explicit reverse starts/ends kernels now derive their own bounded capacity hint and no longer accept a redundant `length` argument. This change removes the corresponding Python-side `ends.max() - starts.min()` computation and stops passing `length` to the size kernel.

ELI5: the Python layer was calculating a size estimate that Rust could calculate more safely from the actual ranges. Removing it keeps one source of truth and avoids reserving based on a potentially misleading span, especially with duplicate labels.

## Performance and memory

The paired janitor-rs PR includes Criterion comparisons against the old HashMap-per-label implementation for unique and duplicate labels at n=32, 10,000, 100,000, and 1,000,000. The compact kernels are generally ~1.5–3x faster for min/max/prod/size and ~15–40% faster for sum, with broadly comparable memory and substantially lower peak live memory in representative duplicate-label cases. The benchmark records allocation count, allocated bytes, and peak live memory.

## Verification

- Python syntax checks passed.
- Pre-commit checks passed except `pixi-install`, which was skipped because dependency resolution was unavailable due to a transient DNS failure; ruff, formatting, pydoclint, interrogate, whitespace, and file checks passed.

Coordinated janitor-rs PR: branch `issue-107-explicit-ranges`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -548,6 +548,23 @@ CLI.
 include MkDocs. The documentation task is available in the `docs` environment.
 **Recommendation**: Run `pixi run -e docs build-docs` to build documentation.
 
+### [2026-08-27] Document the Reverse Aggregation Boundary Contract
+
+**Context**: Coordinating pyjanitor with janitor-rs reverse aggregation kernels.
+**Learning**: Reverse match kernels consume a flattened candidate tape. Direct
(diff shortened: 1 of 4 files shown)


samukweku:

